PR #6, opened 2019-09-12
Debian support

ldelossa:
Adds Debian Oval XML support along with a new method of identifying Version Code Name more aggressively. 

Debian and Ubuntu CVE definitions use this os-release field to match vulnerabilities so an emphasis is made to identify it correctly. 

Draft PR to get comments.

### DIFF
--- a/debian/matcher.go
+++ b/debian/matcher.go
@@ -8,7 +8,9 @@ import (
 
 type Matcher struct{}
 
-func (*Matcher) Interested(pkg *claircore.Package) bool {
+var _ driver.Matcher = (*Matcher)(nil)
+
+func (*Matcher) Filter(pkg *claircore.Package) bool {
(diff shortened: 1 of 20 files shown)


ldelossa:
@hdonnay take one more peak thru - handled checking http responses.

debian/matcher.go
```diff
@@ -0,0 +1,56 @@
+package debian
+
+import (
+	version "github.com/knqyf263/go-deb-version"
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+type Matcher struct{}
```
hdonnay: You should pop a static verification of the interface in here, like `var _ driver.Matcher = (*Matcher)(nil)`, because I think this has the old names.

ldelossa: :+1:

PR #8, opened 2019-09-16
driver: create libvuln/driver package

hdonnay:
This allows external packages to implement driving logic for libvuln.

It also updates the in-tree `ubuntu` package to use the changed interfaces.


ubuntu/matcher.go
```diff
@@ -23,13 +26,13 @@ func (*Matcher) Interested(pkg *claircore.Package) bool {
 	}
 }
 
-func (*Matcher) How() []vulnstore.MatchExp {
-	return []vulnstore.MatchExp{
-		vulnstore.PackageDistributionVersionCodeName,
+func (*Matcher) Matching() []driver.MatchExp {
```
ldelossa: @hdonnay  maybe Query() here ? it would match the Controller method.

ldelossa:
@hdonnay https://github.com/quay/claircore/blob/master/docs/matching_vulns.md update docs plzzzz <3

PR #15, opened 2019-09-18
integration: add test/integration package 

hdonnay:
This is based on my PR #14, so I'll merge it after that.

### DIFF
--- a/internal/vulnstore/postgres/gethash_integration_test.go
+++ b/internal/vulnstore/postgres/gethash_integration_test.go
@@ -1,16 +1,17 @@
-//+build integration
-
 package postgres
 
 import (
 	"context"
 	"testing"
 
+	"github.com/quay/claircore/test/integration"
+
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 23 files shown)


internal/vulnstore/postgres/gethash_integration_test.go
```diff
@@ -1,15 +1,18 @@
...
 
+	"github.com/quay/claircore/test/integration"
+
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 )
 
 func Test_GetHash_KeyNotExists(t *testing.T) {
+	integration.Skip(t)
+	ctx := context.Background()
```
ldelossa: Now that context exists, may not be a bad options to time these out so the test do not run indefinitely if there's a lock.

test/integration/integration.go
```diff
@@ -0,0 +1,9 @@
+package integration
+
+import "testing"
+
+func Skip(t testing.TB) {
```
ldelossa: comments here just to explain what it's doing, and include how it works with run.go and skip.go

ldelossa:
noice

PR #19, opened 2019-09-20
Makefile: add some podman

hdonnay:


ldelossa:
@hdonnay would you be able to bring the http servers up and down as well ? Similar to the local-dev-up target?

hdonnay:
Yeah, I can do that. I'll add that when I get the chance.

Makefile
```diff
@@ -70,3 +70,59 @@ libscanhttp-restart:
...
+	podman create\
+		--pod claircore-dev\
+		--name claircore-database\
+		--env POSTGRES_USER=claircore\
+		--env POSTGRES_DB=claircore\
+		--env PGPORT=5434\
+		--expose 5434\
+		--volume $$(git rev-parse --show-toplevel)/internal/scanner/postgres/bootstrap.sql:/docker-entrypoint-initdb.d/libscan-bootstrap.sql:z\
+		--volume $$(git rev-parse --show-toplevel)/internal/vulnstore/postgres/bootstrap.sql:/docker-entrypoint-initdb.d/libvuln-bootstrap.sql:z\
+		--health-cmd "pg_isready -U claircore -d claircore"\
```
ldelossa: My hunch is this is not going to block the other containers like 

```
	$(docker) exec -it claircore_claircore-db_1 bash -c 'while ! pg_isready; do echo "waiting for postgres"; sleep 2; done'
```

does - is that okay?

hdonnay: You're right, I'll add a thing

PR #21, opened 2019-09-25
rhel updater

hdonnay:
This depends on #20, the first commit is that one.


libvuln/driver/driver.go
```diff
@@ -73,3 +76,20 @@ type Fetcher interface {
 	// a sha265 sum of a OVAL xml file.
 	Fetch() (io.ReadCloser, string, error)
 }
+
+// FetcherNG is an experimental fetcher interface.
+//
+// This may go away or be renamed without warning.
+type FetcherNG interface {
```
ldelossa: Do we want to make this the interface here-on-out ?

libvuln/driver/driver.go
```diff
@@ -73,3 +76,20 @@ type Fetcher interface {
...
+	FetchContext(context.Context, Fingerprint) (io.ReadCloser, Fingerprint, error)
+}
+
+// Unchanged is returned by Fetchers when the database has not changed.
+var Unchanged = errors.New("database contents unchanged")
+
+// Fingerprint is some identifiying information about a vulnerability database.
+type Fingerprint struct {
```
ldelossa: In this fingerprint what is SHA256 a byte array of ? 

This is also in conflict with the current update_cursor record: https://github.com/quay/claircore/blob/master/internal/vulnstore/postgres/bootstrap.sql#L50

We should reconcile what gets persisted and subsequently looked up to understand if a vulnerability database has changed.

libvuln/driver/driver.go
hdonnay: Maybe. Keeping it separate for a bit to feel it out means we don't have to port all the code over at once while we find something that works.

libvuln/driver/driver.go
hdonnay: It's the SHA256 of the fetched content.

And yeah, this should get persisted but the first step is figuring out what the Fetcher will use to prevent duplicating work.

libvuln/driver/driver.go
ldelossa: :+1:

libvuln/driver/driver.go
ldelossa: @hdonnay we cannot assume etags, what if sources are git repos?

ldelossa: @hdonnay what I had in my head originally was for the Fetch method implementer to decide what opaque value they want to use understand if the database has changed, and we simply store this without knowing the details. This gives the implementer a large degree of freedom. 

If the implementer decides to use the etag, or a sha256 of the actual contents, do we need to be aware of this from a "caller of the interface" perspective ?

hdonnay: Yeah it might not exist, that's why there's multiple members and in the experimental interface an argument for passing something into the fetcher.

ldelossa: I guess what I'm struggling with is can't 

SHA256 []byte and ETag string be reduced to just string

string([]byte) == sha256

ldelossa: I guess what I'm struggling with is can't

SHA256 []byte and ETag string be reduced to just string

string([]byte) == sha256

hdonnay: Because one is an opportunistic way to avoid work and one is a strong checksum.

hdonnay: thinking about it more, I don't feel strongly about what's in the fingerprint, just that we should have something like a fingerprint

rhel/rhel.go
```diff
@@ -0,0 +1,329 @@
...
+
+		for _, crit := range crit {
+			select {
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			default:
+			}
+			v := mkVuln()
+			var rt *oval.RPMInfoTest
+			for _, c := range crit {
```
ldelossa: would you be opposed to breaking up the contents of this for loop over `crit` to smaller functions. It may read better if it's a bit more declarative.

rhel/rhel.go
```diff
@@ -0,0 +1,329 @@
...
+				k, i, err := root.Tests.Lookup(c.TestRef)
```
ldelossa: Can root.Tests.Lookup() take a 'kind' string also? This removes the need for the switch cases in the next few lines.

rhel/rhel.go
```diff
@@ -0,0 +1,329 @@
...
+				case "rpminfo_test": // This one
+					rt = &root.Tests.RPMInfoTests[i]
+				default:
+					continue
+				}
+				if len(rt.ObjectRefs) != 1 {
+					return nil, fmt.Errorf("unsure how to handle multple object references in a test")
+				}
+
+				objRef := rt.ObjectRefs[0].ObjectRef
```
ldelossa: Would be nice to make a set of functions which are responsible for pulling out, RMPInfoTests, Obj, and StateRefs, so that this parse method really just calls methods which do the OVAL stuff, spawns a Vulnerability, and attaches data to it.

rhel/rhel.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+	}
+	if u.client == nil {
+		u.client = http.DefaultClient
+	}
+	return u, nil
+}
+
+// Option is a type to configure an Updater.
+type Option func(*Updater) error
```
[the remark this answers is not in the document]
hdonnay: Depends on how much code we share/expose between updaters. We'd probably have to move to a structure like [google.golang.org/api](https://godoc.org/google.golang.org/api)'s option package to be able to share these.

rhel/rhel.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+// Name satisifies the driver.Updater interface.
+func (u *Updater) Name() string {
+	return fmt.Sprintf("rhel-%d-updater", u.version)
+}
+
+// Fetch satisifies the driver.Updater interface.
+func (u *Updater) Fetch() (io.ReadCloser, string, error) {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
hdonnay: No, it causes the returned reader to error when the timeout is canceled.

I could rework the internal fetcher to buffer to disk if you'd rather this have a timeout.

rhel/rhel.go
hdonnay: No, because in the general case, that package can't be sure what kind of Test/Object/State a reference will resolve to.

ldelossa: Isn't that okay tho ? You check this here: 

```
				switch k {
				case "rpminfo_test": // This one
					rt = &root.Tests.RPMInfoTests[i]
				default:
					continue
				}
```

but would it not be okay to do root.Tests.Lookup(c.TestRef, "rpminfo_test") and return error if not present? It just cleans up the branching in calling code.

rhel/rhel.go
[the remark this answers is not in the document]
ldelossa: @hdonnay - up to you - I think the move is to move all of them to downoading the DBs to disk first. We can do this all in one PR if you'd like.

rhel/rhel.go
hdonnay: I don't think moving an `if` into the function is worth it.

ldelossa: fair enough

hdonnay:
Sooooo, this all works, but adding this updater and importing things adds enough rows to make the DB fall apart.

PR #29, opened 2019-09-30
Louis/amazon updater

ldelossa:
This PR implements the Amazon Linux ALAS updater. A parser was written in a separate repository https://github.com/quay/alas


aws/client.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// RepoMD returns a alas.RepoMD containing sha256 information of a repositories contents
+func (c *Client) RepoMD() (alas.RepoMD, error) {
+	for _, mirror := range c.mirrors {
+		m := *mirror
+		m.Path = path.Join(m.Path, repoDataPath)
+
+		c.logger.Debug().Msgf("attempting repomd download from mirror %v", m)
+		resp, err := c.c.Get(m.String())
```
hdonnay: this should have a context.Context wired in through here

aws/client.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	err := client.getMirrors(release)
+	return client, err
+}
+
+// RepoMD returns a alas.RepoMD containing sha256 information of a repositories contents
+func (c *Client) RepoMD() (alas.RepoMD, error) {
+	for _, mirror := range c.mirrors {
+		m := *mirror
+		m.Path = path.Join(m.Path, repoDataPath)
```
hdonnay: If this is a url.URL, you can use `.Parse()` to resolve the reference.

aws/client.go
```diff
@@ -0,0 +1,148 @@
...
+	return alas.RepoMD{}, fmt.Errorf("all mirrors failed to retrieve repo metadata")
+}
+
+// Updates returns the *http.Response of the first mirror to establish a connection
+func (c *Client) Updates() (*http.Response, error) {
+	for _, mirror := range c.mirrors {
+		m := *mirror
+		m.Path = path.Join(m.Path, updatesPath)
+
+		req, err := http.NewRequest(http.MethodGet, m.String(), nil)
```
hdonnay: should take a context.Context

aws/client.go
```diff
@@ -0,0 +1,148 @@
...
+		m := *mirror
+		m.Path = path.Join(m.Path, repoDataPath)
+
+		c.logger.Debug().Msgf("attempting repomd download from mirror %v", m)
+		resp, err := c.c.Get(m.String())
+		if err != nil {
+			c.logger.Error().Msgf("failed to retrieve repomd from mirror %v: %v", m, err)
+			continue
+		}
+		if (resp.StatusCode <= 199) || (resp.StatusCode >= 300) {
```
hdonnay: Why doesn't this just check for `http.StatusOK`? Any other 2xx response would still be weird.

aws/client.go
```diff
@@ -0,0 +1,148 @@
...
+	case Linux1:
+		resp, err = c.c.Get(amazonLinux1Mirrors)
+	case Linux2:
+		resp, err = c.c.Get(amazonLinux2Mirrors)
+	}
+
+	if err != nil {
+		c.logger.Error().Msgf("failed to make request for %v mirrors: %v", release, err)
+		return fmt.Errorf("failed to make request for %v mirrors: %v", release, err)
+	}
```
hdonnay: need to close the response Body

aws/client.go
```diff
@@ -0,0 +1,148 @@
...
+
+		c.logger.Info().Msgf("successfully retrieved updates from mirror %v", m)
+		return resp, nil
+	}
+
+	c.logger.Error().Msg("exhausted all mirrors requesting updates")
+	return nil, fmt.Errorf("all update_info mirrors failed to return a response")
+}
+
+func (c *Client) getMirrors(release Release) error {
```
hdonnay: Should take a context.Context

aws/client_test.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/quay/claircore/test/integration"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Client_GetMirrors(t *testing.T) {
```
hdonnay: The tests that handle the body should have an in-process http server spun up and at least be tested against a known-good body.

aws/updater.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return nil, "", fmt.Errorf("updates repo metadata could not be retrieved: %v", err)
+	}
+
+	resp, err := client.Updates()
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to retrieve update info: %v", err)
+	}
+
+	gzip, err := gzip.NewReader(resp.Body)
```
hdonnay: This should be buffered into a temporary file.

aws/client.go
ldelossa: Going to skip on this one, amazon returns urls which do not end in a slash so the u.Parse() does not to the right thing. https://play.golang.org/p/Uiy1zWJh4sY

I don't want to add a "/" to the end of each string, id prefer just the path.Join method

PR #40, opened 2019-10-09
Alpine Updater

ldelossa:
This PR introduces the Alpine updater. 
It also fixes a small issue where pgx closes a tx on context cancelation

### DIFF
--- a/alpine/release.go
+++ b/alpine/release.go
@@ -1,6 +1,6 @@
 package alpine
 
-// Release is a particular
+// Release is a particular release of the Alpine linux distribution
 type Release string
 
 const (
--- a/alpine/release.go
+++ b/alpine/release.go
@@ -1,0 +1,15 @@
+package alpine
+
+// Release is a particular
+type Release string
+
+const (
+	V3_10 Release = "v3.10"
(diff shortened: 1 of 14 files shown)


alpine/release.go
```diff
@@ -0,0 +1,15 @@
+package alpine
+
+// Release is a particular
```
hdonnay: Comment needs fixing

alpine/secdb.go
```diff
@@ -0,0 +1,41 @@
+package alpine
+
+import (
+	"io"
+
+	"gopkg.in/yaml.v2"
```
hdonnay: why `v2` instead of `v3`?

internal/vulnstore/postgres/putvulnerabilities.go
```diff
@@ -150,8 +150,8 @@ func putVulnerabilities(ctx context.Context, pool *pgxpool.Pool, updater string,
 				return fmt.Errorf("failed processing batch response: %v", err)
 			}
 		}
-		cancel()
 		res.Close()
+		cancel()
```
hdonnay: I think this change needs to be made above, as well.

PR #42, opened 2019-10-14
DB test

hdonnay:
This series adds a database test harness, then fixes some issues I ran into while building it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ genmocks:
 # runs integration tests. database must be available. use the db commands below to ensure this
 .PHONY: integration
 integration:
-	go test -count=1 -p 1 -race -tags integration ./...
+	go test -count=1 -race -tags integration ./...
 
 # runs integration test which may fail on darwin but must succeed on linux/unix
 # using the docker-shell makefile first will drop you into a unix container where
--- a/libscan/http/scanreporthandler.go
+++ b/libscan/http/scanreporthandler.go
@@ -3,46 +3,56 @@ package http
 import (
 	"encoding/json"
 	"fmt"
-	"log"
 	h "net/http"
 	"strings"
 
 	"github.com/quay/claircore/libscan"
+
+	"github.com/rs/zerolog/log"
 )
 
 func ScanReport(lib libscan.Libscan) h.HandlerFunc {
 	return func(w h.ResponseWriter, r *h.Request) {
 		ctx := r.Context()
+		log := log.Logger
 		if r.Method != h.MethodGet {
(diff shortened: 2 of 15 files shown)


Makefile
```diff
@@ -10,7 +10,7 @@ genmocks:
 # runs integration tests. database must be available. use the db commands below to ensure this
 .PHONY: integration
 integration:
-	go test -count=1 -p 1 -race -tags integration ./...
+	go test -count=1 -tags integration ./...
```
ldelossa: @hdonnay should we still run race?

libscan/http/scanreporthandler.go
```diff
@@ -29,7 +28,6 @@ func ScanReport(lib libscan.Libscan) h.HandlerFunc {
 		sr, ok, err := lib.ScanReport(ctx, hash)
 		if err != nil {
 			h.Error(w, fmt.Sprintf("error receiving scanreport"), h.StatusInternalServerError)
-			log.Printf("failed to retrieve scanreport for %v: %v", hash, err)
```
ldelossa: I feel these logs are still convenient - the error above it is returned to the client, we would not see those errors on our log stream watching stdout server side

Makefile
ldelossa: @hdonnay should we still run race

libscan/http/scanreporthandler.go
ldelossa: this message echos for all log statements removed in http handler

test/integration/db.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	defer conn.Close(ctx)
+
+	if _, err := conn.Exec(ctx,
+		`SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname = $1`,
+		db.cfg.ConnConfig.Database); err != nil {
+		t.Error(err)
+	}
+
+	if _, err := conn.Exec(ctx, fmt.Sprintf(`DROP DATABASE %s;`, db.cfg.ConnConfig.Database)); err != nil {
```
ldelossa: nit:

move these to const at the top of the file

test/integration/db.go
```diff
@@ -0,0 +1,127 @@
...
+	seed := rand.NewSource(int64(binary.BigEndian.Uint64(b)))
+	rng = rand.New(seed)
+}
+
+// DefaultDSN is a dsn for database server usually set up by the project's
+// Makefile.
+const DefaultDSN = `host=localhost port=5434 user=claircore dbname=claircore sslmode=disable`
+
+const (
+	createLock = 0xdeadbeef
```
ldelossa: I see no usage of this const - is it necessary?

PR #43, opened 2019-10-16
osrelease: add os-release DistributionScanner

hdonnay:
This attempts to extract the information in os-release (as documented in
https://www.freedesktop.org/software/systemd/man/os-release.html) from
layers as a DistributionScanner.


osrelease/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	if d == nil {
+		return nil, nil
+	}
+	return []*claircore.Distribution{d}, nil
+}
+
+// Readfile looks for the specifed file in the tar archive provided on r.
+func readfile(ctx context.Context, r io.ReadSeeker, n string) (*claircore.Distribution, error) {
+	defer trace.StartRegion(ctx, "readfile").End()
```
ldelossa: Is this going to work the way we think since this is a recursive method ?

osrelease/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+			return nil, fmt.Errorf("osrelease: error reading tar: %w", err)
+		}
+
+		if h.Name != n {
+			continue
+		}
+
+		switch {
+		case h.Typeflag == tar.TypeSymlink:
+			return readfile(ctx, r, h.Linkname)
```
ldelossa: h.Linkname can be a relative path when TypeSymlink is the header type giving you something like `../../path`
that will not exist in the tar bundle and must be handled. We do this here: https://github.com/quay/claircore/blob/master/filer.go#L57-L66

review that logic and see if its necessary.

PR #44, opened 2019-10-21
Libscan plumbing

ldelossa:
Rebased ontop of data-model-changes. This PR implements the plumbing necessary for discrete Repository and Distribution scanners. Prior to this both Distribution and Package scanning was a facility of just a PackageScanner. 

A bunch of code cleanup occurred too which reduces the amount of copying between Libscan Opts and Scanner Opts.


ldelossa:
@hdonnay addressed the allocation concerns in latest push.

PR #46, opened 2019-10-22
dpkg: refit into PackageScanner

hdonnay:
This commit turns the 'dpkg' package into a PackageScanner, instead of
an old-style scanner.


dpkg/scanner.go
```diff
@@ -0,0 +1,208 @@
...
+		case nil:
+		case io.EOF:
+			break Find
+		default:
+			log.Warn().Err(err).Msg("reading next header failed")
+			return nil, fmt.Errorf("reading next header failed: %w", err)
+		}
+		switch filepath.Base(h.Name) {
+		case "status", "available":
+			if h.Typeflag == tar.TypeReg {
```
ldelossa: edge case: status or available files are symlinks to a different name on the FS. Most likely not worth worrying about but worth mentioning for conversation

dpkg/scanner.go
```diff
@@ -0,0 +1,208 @@
...
+
+		// Reset the tar reader.
+		if n, err := r.Seek(0, io.SeekStart); n != 0 || err != nil {
+			log.Warn().Err(err).Msg("unable to seek reader")
+			return nil, fmt.Errorf("unable to seek reader: %w", err)
+		}
+		tr = tar.NewReader(r)
+
+		// We want the "status" file, so search the archive for it.
+		fn := filepath.Join(p, "status")
```
ldelossa: echo symlink issue: this will break if status file is relative symlink.

dpkg/scanner.go
```diff
@@ -0,0 +1,208 @@
...
+			log.Warn().Err(err).Msg("reading status file from layer failed")
+			return nil, fmt.Errorf("reading status file from layer failed: %w", err)
+		case db == nil:
+			log.Error().Str("filename", fn).Msg("unable to get reader for file")
+			panic("file existed, but now doesn't")
+		}
+
+		// Take all the packages found in the database and attach to the slice
+		// defined outside the loop.
+		found := make(map[string]*claircore.Package)
```
ldelossa: nit: 
for semantic reasons I would move this and and the md5'ing below it to their own methods.

dpkg/scanner.go
```diff
@@ -0,0 +1,208 @@
...
+			log.Warn().Err(err).Msg("unable to seek reader")
+			return nil, fmt.Errorf("unable to seek reader: %w", err)
+		}
+		tr = tar.NewReader(r)
+
+		// We want the "status" file, so search the archive for it.
+		fn := filepath.Join(p, "status")
+		var db io.Reader
+		var h *tar.Header
+		for h, err = tr.Next(); err == nil; h, err = tr.Next() {
```
ldelossa: almost certainly not worth it as far as optimization but for a conversation, you already seeked the file to the header to find the status file. I believe you could use that offset and record it in your map, then seek the reader directly to that file's header.

dpkg/scanner.go
```diff
@@ -0,0 +1,208 @@
...
+func (ps *Scanner) Scan(layer *claircore.Layer) ([]*claircore.Package, error) {
+	ctx := context.TODO()
+	ctx = log.Logger.WithContext(ctx)
+	return ps.ScanContext(ctx, layer)
+}
+
+// ScanContext attempts to find a dpkg database within the layer and read all of the
+// installed packages it can find in the "status" file.
+//
+// It's expected to return (nil, nil) if there's no dpkg database in the layer.
```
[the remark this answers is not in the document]
hdonnay: It's not an error case for the scanner to return nothing, and a nil slice still gets ranged over correctly.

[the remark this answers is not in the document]
ldelossa: whoops you're right :+1:

PR #48, opened 2019-10-23
etc: make new base image based on RHEL UBI

hdonnay:


ldelossa:
Thank you for your docker.

PR #52, opened 2019-11-01
Added tracing via OpenTelemetry with Jaeger exporter

### LINKED ISSUE #11: Request Tracing
If this is going to become a part of the Quay stack, we'll want to make sure it's compatible with the distributed tracing solution.

cc @jpkrohling

jpkrohling:
This PR is a first draft adding distributed tracing to the main claircore code base.

The current state is just enough code to get something to discuss, to validate if the approach is acceptable to the project maintainers.

With this PR, this is a sample trace for the initial vulnerability scanner:

image

Other simpler traces are also present.

Closes #11.

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>


ldelossa:
@jpkrohling very cool.

We just merged a very large PR into master. Can you re-base off this? 

I do not have a ton of experience with tracing. My first concerns are shoving any code right into DB handling methods - this may not be warranted however. 

Does this implementation provide a way to toggle tracing on and off based on configuration?

jpkrohling:
I thought I had added a flag `tracing.enabled`, but looks like I removed it. I can add it to make it explicit. The current PR disables tracing if the flag `JaegerAgentHostPort` is set to empty, but I think the default value might be set (localhost:6831).

About the database operations: recording a span is a low-risk operation, as the general philosophy is that tracing should *never* break the application being traced. Creating a span is a fast operation, and the real blocking work happens in a background thread. That all said, it might indeed be the case that the value provided by tracing the DB operations is low. I guess only you can judge that :-) I can certainly remove the instrumentation from that part.

I'm out this week and next, but I can certainly rebase the PR based on the current master in the week of Nov 18. Should I take it that the points above are the only concerns about this PR and that the general direction is valid?

jpkrohling:
PR rebased. Here's a new trace based on the current state of this PR:

image

jpkrohling:
I just updated this PR to include a flag to enable/disable tracing. By default, tracing is disabled:

```
$ podman logs libscanhttp
9:26AM INF tracing is disabled
9:26AM DBG starting http server on 0.0.0.0:8080
```

The flag is explicitly set when running `make podman-dev-up`:

```
$ podman logs libscanhttp
9:28AM INF tracing is enabled with the Jaeger Exporter
9:28AM DBG starting http server on 0.0.0.0:8080
```

jpkrohling:
Is this something that is of interest to the project?

ldelossa:
@jpkrohling 

I'm pretty interested in incorporating tracing. However ClairCore is currently under heavy development and I'd like to see it reach v1.0.0 and integrate with Clair under a release branch before we implement this. Sorry for the delay - we are still fleshing out the stable APIs and architecture of the project.

jpkrohling:
Sure, just ping me when this is ready to be reviewed/merged and I'll rebase. For now, I'm removing this from my queue.

PR #69, opened 2019-11-22
RHEL matcher

hdonnay:
This series of commits creates a matcher that works from an OVAL RPM database and adds it to the default set.

Also contains bugfixes and utilities created along the way.


rpm/matcher.go
```diff
@@ -0,0 +1,34 @@
...
+	return true
+}
+
+func (*Matcher) Query() []driver.MatchExp {
+	return []driver.MatchExp{
+		driver.PackageDistributionCPE,
+	}
+}
+
+func (*Matcher) Vulnerable(record *claircore.IndexRecord, vuln *claircore.Vulnerability) bool {
```
ldelossa: I'd put a little comment here just explaining when/why we'd want out the cmp function.

internal/vulnstore/postgres/querybuilder.go
```diff
@@ -24,58 +24,39 @@ func getBuilder(matchers []driver.MatchExp) (string, []driver.MatchExp, error) {
 	seen := make(map[driver.MatchExp]struct{})
 	deduped := []driver.MatchExp{}
 	for _, m := range matchers {
+		if _, ok := seen[m]; ok {
+			continue
+		}
+		var ex goqu.Ex
 		switch m {
+		case driver.PackageName: // ???
```
ldelossa: Can we just nix these from this switch and from MatchExps ?

hdonnay: Yeah, I think removing them and renaming those fields is in order. Is that a separate PR?

ldelossa: @hdonnay you can make a judgement call on that. If it's a far-reaching rename and touches a lot maybe we punt to a dedicated PR. If it seems quick then by all means do it in this one.

rpm/matcher.go
hdonnay: OK

internal/vulnstore/postgres/get.go
```diff
@@ -11,20 +14,34 @@ import (
...
 	query, dedupedMatchers, err := getBuilder(opts.Matchers)
+	if err != nil {
+		return nil, err
+	}
+	h := fnv.New64a()
+	if _, err := io.WriteString(h, query); err != nil {
+		return nil, err
+	}
+	name := hex.EncodeToString(h.Sum(nil))
+	log.Debug().Str("name", name).Msg("built query")
```
ldelossa: Is there value in displaying a hash of the query?

package.go
```diff
@@ -4,17 +4,17 @@ type Package struct {
...
-	// the version of the distribution
+	// the version of the package
 	Version string `json:"version"`
 	// type of package. currently expectations are binary or source
-	Kind string `json:"kind"`
+	Kind string `json:"kind,omitempty"`
 	// if type is a binary package a source package maybe present which built this binary package.
 	// must be a pointer to support recursive type:
-	Source *Package `json:"source"`
+	Source *Package `json:"source,omitempty"`
```
ldelossa: omit empty can break downstream clients no? I usually avoid the need to explain "this field might be there" to ppl utilizing structs via json.

rpm/matcher.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// Matcher implements driver.Matcher.
+type Matcher struct{}
+
+var _ driver.Matcher = (*Matcher)(nil)
+
+// Filter implements driver.Matcher.
+func (*Matcher) Filter(record *claircore.IndexRecord) bool {
+	// record.Repository.Kind != "rpm"
```
ldelossa: This is wrong in the case when the record stream this Matcher is dealing with has more then just RPM records. This is an issue of having a single rpm `Matcher`. You must still filter for RPM records or you'll attempt to query non-rpm packages.

rpm/packagescanner.go
```diff
@@ -129,7 +132,7 @@ func (ps *Scanner) ScanContext(ctx context.Context, layer *claircore.Layer) ([]*
 	found := make([]string, 0)
 	for k, score := range possible {
 		if score == len(dbnames) {
-			found = append(found, k)
+			found = append(found, filepath.Join("/", k))
```
ldelossa: Isn't this the case for **just** RHEL containers? Not other RPM based distros?

osrelease/scanner.go
```diff
@@ -123,21 +139,48 @@ func parse(ctx context.Context, r io.Reader) (*claircore.Distribution, error) {
 
 		switch key {
 		case "ID":
+			log.Debug().Msg("found ID")
```
ldelossa: I think these debug messages should go in favor or a single one that summarizes the fully realized dist struct.

osrelease/scanner.go
```diff
@@ -123,21 +139,48 @@ func parse(ctx context.Context, r io.Reader) (*claircore.Distribution, error) {
...
+			log.Debug().Msg("found VERSION")
 			d.Version = value
 		case "ID_LIKE":
 		case "VERSION_CODENAME":
+			log.Debug().Msg("found VERISON_CODENAME")
 			d.VersionCodeName = value
 		case "PRETTY_NAME":
+			log.Debug().Msg("found PRETTY_NAME")
+			d.PrettyName = value
+		case "REDHAT_BUGZILLA_PRODUCT":
```
ldelossa: I would feel better if this was done correctly and we plumbed it into the datamodel. If it's a field that's worth having on the os-release file we don't need to munge it.

ldelossa: Lets void this for now. I think this becomes a moot point once we introduce inverted index.

internal/vulnstore/postgres/get.go
hdonnay: I put it in there as assurance that a query is changing with a different set of matchexps provided

rpm/matcher.go
hdonnay: There's no way to filter for rpm records

osrelease/scanner.go
hdonnay: you mean by loading the os-release file into something like an hstore? That's currently my thinking on how we should approach this.

rpm/packagescanner.go
hdonnay: No, this makes sure the paths are absolute.

rpm/matcher.go
ldelossa: @hdonnay I see.
I would create per-distro matchers in this case like the others. So a RHEL matcher which filters for something in record.Dist.

osrelease/scanner.go
ldelossa: @hdonnay right now the inverted_index branch has us just injesting the vulnerabilities as jsonb documents to obtain the same thing as Hstore in principal. You can make a GIN index on hstore also. But I implemented as jsonb so far.

package.go
hdonnay: this can already be there or not, because it's a pointer. this just makes the resulting json smaller.

ldelossa: @hdonnay the pointer is used to support a recursive datastructure - not so much to support a nil value. 

I can see the point, but won't this make clients more brittle ? And our openapi docs less reliable? I guess right now we return an empty value for it.

PR #71, opened 2019-11-25
Misc fixes

hdonnay:
Found these while doing some testing.

### DIFF
--- a/rpm/packagescanner.go
+++ b/rpm/packagescanner.go
@@ -169,9 +169,19 @@ func (ps *Scanner) ScanContext(ctx context.Context, layer *claircore.Layer) ([]*
 	// permissions later. Otherwise, some layers will have directories without
 	// the write bit created and the permissions set before their contents are
 	// created.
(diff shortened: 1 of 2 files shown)


hdonnay:
Blocks #69

rpm/packagescanner.go
```diff
@@ -169,17 +169,24 @@ func (ps *Scanner) ScanContext(ctx context.Context, layer *claircore.Layer) ([]*
 	// permissions later. Otherwise, some layers will have directories without
 	// the write bit created and the permissions set before their contents are
 	// created.
-	tarcmd := exec.CommandContext(ctx, "tar", "-xC", root)
+	errbuf := bytes.Buffer{}
+	tarcmd := exec.CommandContext(ctx, "tar", "-xC", root, "--exclude", "dev")
```
ldelossa: @hdonnay a little bit of nit, but a comment here about how unprivileged containers will fail without this would help readers understand wh we now exclude.

hdonnay: Agree

PR #79, opened 2019-12-06
Louis/interface swap

ldelossa:
This PR swaps all interfaces to ones which take ctx for cancellation. 
Along the way several un-ran tests (due to bad usage of build tags) were not ran. This PR also makes all integration tests pass. 
Additional work:
* Each TestStore constructor (vulnstore/postgres, indexer/postgres) now returns a DSN of the generated database connection string.
* indexer/controller integration tests were fixed


libvuln/driver/driver.go
```diff
@@ -68,25 +68,25 @@ type Parser interface {
...
+// 	// Fetch should retrieve the target vulnerability data and return an io.ReadCloser
+// 	// with the contents. Fetch should also return a string which can used to determine
+// 	// if these contents should be applied to the vulnerability database. for example
+// 	// a sha265 sum of a OVAL xml file.
+// 	Fetch() (io.ReadCloser, string, error)
+// }
 
 // FetcherNG is an experimental fetcher interface.
 //
 // This may go away or be renamed without warning.
```
hdonnay: Fix this documentation.

PR #84, opened 2019-12-10
Cctool manifest

hdonnay:
This enables `cctool` to understand manifests and generate its own URLs.

This also contains commits for fixing bugs that were shook out while testing this change.


internal/indexer/fetcher/fetcher.go
```diff
@@ -191,10 +191,9 @@ func (f *fetcher) onDisk(ctx context.Context, contents io.ReadCloser, layer *cla
 		return nil, fmt.Errorf("defaultFetcher: unable to create temp file for archive contents")
 	}
 	defer fd.Close()
-	bufferedFD := bufio.NewWriter(fd)
 
 	// write tar to temp file
-	_, err = io.Copy(bufferedFD, contents)
+	_, err = io.Copy(fd, contents)
```
ldelossa: We can still use buff io and call the Flush method, no ?

PR #85, opened 2019-12-11
Add and plumb indexer state

hdonnay:
These commits make a stable identifier for the code of the indexer and its
scanners, and exposes it in libindexhttp.


libindex/libindex.go
```diff
@@ -52,6 +57,26 @@ func New(ctx context.Context, opts *Opts) (*Libindex, error) {
...
+
+	h := md5.New()
+	var ns []string
+	m := make(map[string][]byte)
+	for _, s := range vscnrs {
+		n := s.Name()
+		m[n] = []byte(n + s.Version() + s.Kind() + "\n")
+		ns = append(ns, n)
+	}
+	if _, err := io.WriteString(h, "libindex number: 1\n"); err != nil {
```
ldelossa: why? @hdonnay

hdonnay: I mentioned this in person; this is a string we can change when code inside libindex itself changes.

ldelossa: that was my hunch, feels like it should be somewhere/something more then an line string. My head goes to a Version constant of some sort.

PR #86, opened 2019-12-11
Louis/sql optimize

ldelossa:
This PR optimizes the multi column index to favor the most used MatchExp in each Matcher. 
The query builder is now changed to create queries on a per ScanRecord basis. 
Draft until I rewrite test for the query builder.


rhel/matcher.go
```diff
@@ -19,6 +21,9 @@ func (*Matcher) Name() string {
 
 // Filter implements driver.Matcher.
 func (*Matcher) Filter(record *claircore.IndexRecord) bool {
+	if record.Distribution == nil {
+		log.Printf("Dist is nil")
```
hdonnay: This looks like a leftover debugging print

internal/vulnstore/postgres/get.go
```diff
@@ -14,71 +11,28 @@ import (
...
-				args = append(args, record.Distribution.Arch)
-			case driver.DistributionCPE:
-				args = append(args, record.Distribution.CPE)
-			case driver.DistributionPrettyName:
-				args = append(args, record.Distribution.PrettyName)
-			}
+		query, err := getQueryBuilder(record, opts.Matchers)
+		if err != nil {
+			// if we cannot build a query for an individual record continue to the next
+			log.Debug().Msgf("could not build query for record: %+v", record)
```
hdonnay: Add record as a field and make the message static

PR #95, opened 2020-01-08
Louis/fastest url

ldelossa:
This PR implements a method of obtaining the fastest http.Response given a list of urls. The intended usage is for lists of mirrors where the same content lives at different hosted servers and you want to a client to receive the first mirror that responds.

### DIFF
--- a/pkg/fastesturl/fastesturl.go
+++ b/pkg/fastesturl/fastesturl.go
@@ -62,7 +62,7 @@ func (f *FastestURL) Do(ctx context.Context) *http.Response {
 	var response *http.Response
 	tctx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
-	// immediately lock so workers do not write to f.Response
+	// immediately lock so workers do not write to response
 	// before Do method blocks on cond.Wait()
 	cond.L.Lock()
 	for _, url := range f.URLs {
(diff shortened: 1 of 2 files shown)


pkg/fastesturl/fastesturl.go
```diff
@@ -0,0 +1,107 @@
...
+	for _, url := range f.URLs {
+		u := url
+		go func() {
+			defer cond.Signal()
+			req := f.Request.Clone(tctx)
+			req.URL = u
+			req.Host = u.Host
+			resp, err := f.Client.Do(req)
+			if err != nil {
+				log.Error().Err(err).Msgf("failed to make request for url: %v", url)
```
hdonnay: Use a constant message, and add the URL as a field.

pkg/fastesturl/fastesturl.go
```diff
@@ -0,0 +1,107 @@
...
+				return
+			}
+			if f.RespCheck(resp) {
+				cond.L.Lock()
+				if response == nil {
+					response = resp
+				} else {
+					// another routine has set f.Response, close this body and discard
+					resp.Body.Close()
```
hdonnay: The body also needs to be closed if `RespCheck() == false`

pkg/fastesturl/fastesturl.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Do will return the first *http.Response which passes
+// f.RespCheck.
+//
+// If no successful *http.Response is obtained a nil is returned
+func (f *FastestURL) Do(ctx context.Context) *http.Response {
+	log := zerolog.Ctx(ctx).With().Str("routine", "fasesturl_do").Logger()
+	cond := sync.NewCond(&sync.Mutex{})
+	var response *http.Response
+	tctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
ldelossa: @hdonnay 

What do you think is the right way to handle this timeout? Should it remain constant

hdonnay: I think the caller should handle it with the passed-in context

ldelossa: cool, I won't derive a ctx at all then and assume the one being passed in will be used as the caller defines.

PR #98, opened 2020-01-10
Louis/opaque ids

ldelossa:
Transforms all IDs to opaque strings. Right now these strings are DB ids but this puts us in the position to obfuscate if desired moving forward.


ldelossa:
@hdonnay 

I need to regen the test data in the rhel package. What image did you use to gen that IndexReport?

PR #104, opened 2020-01-14
Don't persist empty packages

hdonnay:
Does what it says on the tin, and also adds some logging about what was discarded.

### DIFF
--- a/internal/indexer/postgres/indexpackage.go
+++ b/internal/indexer/postgres/indexpackage.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/jmoiron/sqlx"
+	"github.com/rs/zerolog"
 
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/internal/indexer"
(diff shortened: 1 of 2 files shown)


internal/indexer/postgres/indexpackage.go
```diff
@@ -139,6 +156,10 @@ func indexPackages(ctx context.Context, db *sqlx.DB, pool *pgxpool.Pool, pkgs []
 	if err != nil {
 		return fmt.Errorf("final batch insert failed for package_scanartifact: %v", err)
 	}
+	log.Debug().
+		Int("skipped", skipCt).
+		Int("inserted", len(pkgs)-skipCt).
+		Msg("scanartifactss inserted")
```
ldelossa: double s :snake:

internal/vulnstore/postgres/putvulnerabilities.go
```diff
@@ -113,11 +117,13 @@ func putVulnerabilities(ctx context.Context, pool *pgxpool.Pool, updater string,
...
 	// safe sized batch inserts to postgres
 	mBatcher := microbatch.NewInsert(tx, 2000, time.Minute)
 	for _, vuln := range vulns {
-		if vuln.Package == nil {
-			vuln.Package = &claircore.Package{}
+		if vuln.Package == nil || vuln.Package.Name == "" {
+			skipCt++
+			continue
 		}
 		if vuln.Dist == nil {
```
ldelossa: @hdonnay 
I'm thinking... if we do not have a distribution associated with the vulnerability - we don't match it either...
With your knowledge of the RPM stuff as well does it make sense to ignore vulns where we can't get a distribution either? Most of the time we munge this field ourselves anyway.

hdonnay: I'm fine with that, I think. I'll think about it a bit and complain if I can come up with a counterexample.

ldelossa: @hdonnay 

Thinking about it... all matchers return a MatchContraint which means they require their pending vulns to have dists as well.

PR #106, opened 2020-01-15
Louis/remove db coalescer

ldelossa:
This PR removes the indexer's Store implementation from the Coalescer implementations. This is step one to moving internal indexer interfaces to a driver package.


internal/indexer/linux/coalescer.go
```diff
@@ -23,18 +22,12 @@ type layerArtifacts struct {
 // for dpkg, os-release, and apt scanners
 type Coalescer struct {
 	// a store to access scanartifacts
```
hdonnay: old comment

internal/indexer/controller/coalesce.go
```diff
@@ -13,22 +13,55 @@ import (
 
 // coalesce calls each ecosystem's coalescer and merges the returned IndexReports
 func coalesce(ctx context.Context, s *Controller) (State, error) {
-	coalescers := []indexer.Coalescer{}
+	cctx, cancel := context.WithCancel(ctx)
```
hdonnay: Why this inner context? I don't think anything gets spawned outside the errgroup. Using `errgroup.WithContext` would be simpler

ldelossa: If the main go routine encounters an error dispatching Coalesce go routines, I want the entire tree of routines to end, so I derive a ctx with cancel immediately and then defer the cancel. If the main go routines ends with an exception, all children are killed as well. 

it seems like errgroup.WithContext() just adds an additional ctx to maintain in the method and doesn't provide much other value. 

for example:

```
// coalesce calls each ecosystem's coalescer and merges the returned IndexReports
func coalesce(ctx context.Context, s *Controller) (State, error) {
	cctx, cancel := context.WithCancel(ctx)
	defer cancel() // still need to derive this so main routine can cancel
	mu := sync.Mutex{}
	reports := []*claircore.IndexReport{}
	g, ctx := errgroup.WithContext(cctx) // don't have much use for this derived ctx
	// dispatch a coalescer go routine for each ecosystem
	for _, ecosystem := range s.Ecosystems {
```

instead I just pass the launched go routines the cctx which, if there is an error in the dispatching main go routine, any inflight will cancel.

hdonnay: Oh, I see, because the layer artifacts are being constructed in serial and then only concurrently coalescing at the end.

ldelossa: yup, kinda like a "prep then dispatch" thing.

ldelossa:
@hdonnay 

Just dealt with merge conflicts. The new Digest type makes it tough to test arbitrary layer fixtures. 
I made Repr public with a comment in order to create Layer objects in test tables without trouble.

PR #110, opened 2020-01-21
cctool: add "manifest" subcommand

hdonnay:
This adds a subcommand to cctool that just emits claircore manifests on stdout.

### DIFF
--- a/cmd/cctool/manifest.go
+++ b/cmd/cctool/manifest.go
@@ -1,0 +1,76 @@
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+	"time"
+)
+
+type manifestConfig struct {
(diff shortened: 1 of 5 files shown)


docs/contributor.md
```diff
@@ -1 +1,3 @@
 # Contributors
+This is some documentation helpful for the actual nuts-n-bolts of adding code to
```
ldelossa: 👍

cmd/cctool/manifest.go
```diff
@@ -0,0 +1,76 @@
...
+	"time"
+)
+
+type manifestConfig struct {
+	timeout time.Duration
+	pretty  bool
+}
+
+// Manifest is the subcommand for generating container manifests.
+func Manifest(cmd context.Context, cfg *commonConfig, args []string) error {
```
ldelossa: var name `cmd` intentional?

hdonnay: Yeah, it's keeping with the bad names I established in the report command.

PR #116, opened 2020-01-29
Louis/advisory parsing

ldelossa:
This PR re-works rpm vulnerability parsing to use a consistent data model for matching integrity.
This is follow-up work for per-distribution matchers

### DIFF
--- a/pkg/ovalutil/rpm.go
+++ b/pkg/ovalutil/rpm.go
@@ -1,0 +1,107 @@
+package ovalutil
+
+import (
+	"context"
+
+	"github.com/quay/claircore"
+	"github.com/quay/goval-parser/oval"
+	"github.com/rs/zerolog"
+)
+
+// ProtoVulnFunc allows a caller to create a prototype vulnerability that will be used
+// copied and further defined for every applicable oval.Criterion discovered.
+//
+// This allows the caller to use oval.Definition fields and closure syntax when
+// defining how a vulnerability should be parsed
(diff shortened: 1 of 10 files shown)


pkg/ovalutil/rpm.go
```diff
@@ -0,0 +1,107 @@
...
+// copied and further defined for every applicable oval.Criterion discovered.
+//
+// This allows the caller to use oval.Definition fields and closure syntax when
+// defining how a vulnerability should be parsed
+type ProtoVulnFunc func(def oval.Definition) (*claircore.Vulnerability, error)
+
+// RPMDefsToVulns iterates over the definitions in an oval root and assumes RPMInfo objects and states.
+//
+// Each Criterion encountered with an EVR string will be tranlated into a claircore.Vulnerability
+func RPMDefsToVulns(ctx context.Context, root oval.Root, protoVuln ProtoVulnFunc) ([]*claircore.Vulnerability, error) {
```
hdonnay: I don't understand why the existing `RPMInfo.Extract` method wasn't extended to do this, as it already uses this sort of closure passing for creating vulnerability structs internally.

ldelossa: This is simplified and easier to grok.

ldelossa: Current method attempts to use OR and AND criteria prematurely, we need to work that into the entire application logic first. This is a refactor that grabs just what is necessary and keeps a slim DFS of the criterion. 

When we take a holistic look at Oval we can look back at the previous code which extracts AND and ORs.

PR #120, opened 2020-02-06
Static Typing for Vulnerability Severity Level

### LINKED ISSUE #119: Vulnerability Severity Should be an Enum
### Description

In previous Clair implementation, the `Severity` of a given `Vulnerability` would be one of a defined enum. Currently, [this struct field is typed as `string`](https://github.com/quay/claircore/blob/master/vulnerability.go#L17). We can make this a static type.

alecmerdler:
### Description

Add Go-style string type alias enum for vulnerability severity field.

Fixes #119


ldelossa:
closing this - merged into louis/severity-mapping

PR #122, opened 2020-02-17
python language support

hdonnay:
This PR adds support for scanning and matching python packages.


Makefile
```diff
@@ -103,7 +103,7 @@ podman-dev-up:
 		--env LOG_LEVEL="debug"\
 		--expose 8080\
 		--volume $$(git rev-parse --show-toplevel)/:/src/claircore/:z\
-		quay.io/claircore/golang:1.13.3\
+		quay.io/claircore/golang:$(GO_VERSION)\
```
ldelossa: should a default at top of file be declared?

alpine/packagescanner.go
```diff
@@ -119,7 +119,12 @@ func (*Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircore.
 		pkgs = append(pkgs, &p)
 	}
 	if err := s.Err(); err != nil {
-		return nil, err
+		if len(pkgs) == 0 {
```
ldelossa: what's the idea here? if you've already scanned some packages, ignore the scan error?
we've been following a "if any error, fail hard" policy - should we not here?

indexreport.go
```diff
@@ -7,6 +7,7 @@ package claircore
 //
 // A list of these can be thought of as an "unpacked" IndexReport
 type IndexRecord struct {
+	ID           string
```
ldelossa: Why does the root IndexRecord grow an ID?
Each object within the IndexRecord has an ID.

indexreport.go
```diff
@@ -35,15 +36,30 @@ type IndexReport struct {
...
-	for _, pkg := range report.Packages {
-		for _, env := range report.Environments[pkg.ID] {
-			record := &IndexRecord{}
-			record.Package = pkg
-			record.Distribution = report.Distributions[env.DistributionID]
-			record.Repository = report.Repositories[env.RepositoryID]
+	for pkgID, pkg := range report.Packages {
+		for _, env := range report.Environments[pkgID] {
+			record := &IndexRecord{
+				ID:      pkgID,
```
ldelossa: This seems odd... The IndexRecord is just a container, why would you float the pkgID to a root field ?

internal/indexer/postgres/indexpackage.go
```diff
@@ -44,13 +45,15 @@ INSERT INTO package_scanartifact (layer_hash, package_db, repository_hint, packa
           ON CONFLICT DO NOTHING;`
 )
 
+var zeroPackage = claircore.Package{}
```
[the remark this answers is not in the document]
ldelossa: @hdonnay actually, I think this is an error... nested pointers for source pkg will probably cause de reference errors.
Fully initialize it.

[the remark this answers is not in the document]
hdonnay: Nothing's nested, this just avoids allocating every time a package doesn't have a source package associated.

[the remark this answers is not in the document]
ldelossa: Package nests a pointer to source package.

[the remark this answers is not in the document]
ldelossa: https://github.com/quay/claircore/blob/master/package.go#L15

Makefile
hdonnay: It's defined just below this, above the `baseimage` target

internal/matcher/controller.go
```diff
@@ -61,6 +71,16 @@ func (mc *Controller) Match(ctx context.Context, records []*claircore.IndexRecor
 	return filteredVulns, nil
 }
 
+// DbFilter reports whether the db-side version filtering can be used, and
+// whether it's authoritative.
+func (mc *Controller) dbFilter() (bool, bool) {
+	f, ok := mc.m.(driver.VersionFilter)
```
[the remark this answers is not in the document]
hdonnay: No, it just allows a matcher to opt-in very easily by implementing the interface after setting up the versions.

internal/indexer/postgres/indexpackage_test.go
```diff
@@ -201,7 +201,7 @@ func Test_IndexPackages_Success(t *testing.T) {
 			ctx, done := context.WithCancel(ctx)
 			defer done()
 			ctx = log.TestLogger(ctx, t)
-			db, store, _, teardown := TestStore(ctx, t)
+			db, store, teardown := TestStore(ctx, t)
```
ldelossa: still confused about why dsn was removed?

version.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	return buf.String()
+}
+
+// Compare returns an integer describing the relationship of two Versions.
+//
+// The result will be 0 if a==b, -1 if a < b, and +1 if a > b. If the Versions
+// are of different kinds, the Kinds will be compared lexographically.
+func (v *Version) Compare(x *Version) int {
```
ldelossa: If we hardcode a compare method, what is the point of having "kind" on our Version? We are essentially hard coding one "kind" a "kind" that sorts linearly from 0-9

python/ecosystem.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/internal/indexer"
+)
+
+var scanners = []indexer.PackageScanner{&Scanner{}}
```
ldelossa: by doing this you are sharing a single scanner between all go routines, this is okay?

python/ecosystem.go
```diff
@@ -0,0 +1,73 @@
...
+func NewEcosystem(ctx context.Context) *indexer.Ecosystem {
+	return &indexer.Ecosystem{
+		PackageScanners:      func(_ context.Context) ([]indexer.PackageScanner, error) { return scanners, nil },
+		DistributionScanners: func(_ context.Context) ([]indexer.DistributionScanner, error) { return nil, nil },
+		RepositoryScanners:   func(_ context.Context) ([]indexer.RepositoryScanner, error) { return nil, nil },
+		Coalescer:            NewCoalescer,
+	}
+}
+
+func NewCoalescer(_ context.Context) (indexer.Coalescer, error) {
```
ldelossa: nit: separate file.

python/ecosystem.go
```diff
@@ -0,0 +1,73 @@
...
+	return &coalescer{}, nil
+}
+
+type coalescer struct {
+}
+
+func (c *coalescer) Coalesce(ctx context.Context, ls []*indexer.LayerArtifacts) (*claircore.IndexReport, error) {
+	repos := make(map[string]string)
```
ldelossa: Can you comment here explaining your Coalescing strategy ?

We must evaluate the "newest" pip database (the ones higher in the layer array) to get the final set of pip packages in the resulting image.

rhel/matcher.go
```diff
@@ -19,7 +19,8 @@ func (*Matcher) Name() string {
 
 // Filter implements driver.Matcher.
 func (*Matcher) Filter(record *claircore.IndexRecord) bool {
-	return record.Distribution.DID == "rhel"
+	return record.Distribution != nil &&
```
ldelossa: why this random change and only for rhel?

python/matcher.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/quay/claircore/libvuln/driver"
+	"github.com/quay/claircore/pkg/pep440"
+)
+
+var (
+	_ driver.Matcher       = (*Matcher)(nil)
+	_ driver.VersionFilter = (*Matcher)(nil)
+)
+
+// Matcher attempts to
```
alecmerdler: Incomplete comment

internal/vulnstore/postgres/putvulnerabilities.go
```diff
@@ -126,12 +137,50 @@ func putVulnerabilities(ctx context.Context, pool *pgxpool.Pool, updater string,
 			skipCt++
 			continue
 		}
-		if vuln.Dist == nil || vuln.Dist.Name == "" {
-			skipCt++
-			continue
+		if vuln.Dist == nil {
```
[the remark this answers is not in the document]
hdonnay: We can't, be cause we're now talking about packages that aren't tied to dists.

python/ecosystem.go
hdonnay: I'm not sure what you mean by "pip database".

The data model currently doesn't have any concept of operations on the set of packages, it just presents the complete set on each layer. Because we can't do that, this does the conservative thing and assumes nothing is uninstalled.

rhel/matcher.go
hdonnay: It was the only instance I ran into while testing.

version.go
hdonnay: The point of kind is so that application code can ensure it's making not making nonsensical comparisons, like trying to compare a pep440 version with an rpm version, as they may encode differently.

internal/indexer/postgres/teststore.go
```diff
@@ -16,22 +16,16 @@ import (
...
 
-func TestStore(ctx context.Context, t testing.TB) (*sqlx.DB, *store, string, func()) {
-	cmd := exec.Command("go", "list", "-f", "{{.Dir}}", "github.com/quay/claircore/internal/indexer/postgres")
-	o, err := cmd.Output()
-	if err != nil {
-		t.Fatal(err)
-	}
-	o = bytes.TrimSpace(o)
-
+func TestStore(ctx context.Context, t testing.TB) (*sqlx.DB, *store, func()) {
```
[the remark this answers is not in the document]
hdonnay: The only place using the returned DSN was in the libindex integration test, which doesn't use a store. So, I changed that test to create its own database and simplified every call site for this function.

python/matcher.go
hdonnay: will fix

version.go
ldelossa: I do not understand how something can encode differently, but still use this compare method which infers encoding.

python/ecosystem.go
ldelossa: discussed out of band, we'll determine the best way to go  about this in subsequent PR.

rhel/matcher.go
ldelossa: makes me a bit skeptical that it maybe an issue somewhere else, maybe just confirm its not.

PR #143, opened 2020-03-12
fastesturl: deflake the test

hdonnay:
This commit changes the tests to sequence responses without using
sleeps.

Over the course of testing, using high counts caused the tests to hang.
I ended up restructuring the internals to fix that.

### DIFF
--- a/pkg/fastesturl/fastesturl.go
+++ b/pkg/fastesturl/fastesturl.go
@@ -30,17 +30,14 @@ type FastestURL struct {
 // New is a constructor for a FastestURL
 func New(client *http.Client, req *http.Request, check RespCheck, urls []*url.URL) *FastestURL {
 	if client == nil {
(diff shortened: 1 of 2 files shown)


pkg/fastesturl/fastesturl.go
```diff
@@ -55,55 +52,43 @@ func New(client *http.Client, req *http.Request, check RespCheck, urls []*url.UR
...
+	result := make(chan *http.Response)
+	wg.Add(len(f.URLs))
+	go func() {
+		wg.Wait()
+		close(result)
+	}()
 	for _, url := range f.URLs {
 		u := url
 		go func() {
-			defer cond.Signal()
```
hdonnay: I think under some conditions, this could be hit before the main goroutine was `Wait`ing below, making the tests hang.

PR #144, opened 2020-03-13
update diffs

hdonnay:
This PR is a bit of a mess, but does what it says on the tin.


internal/vulnstore/postgres/updatevulnerabilities.go
```diff
@@ -0,0 +1,192 @@
...
+// giving us a stable, context-free identifier for this revision of the
+// Vulnerability.
+func md5Vuln(v *claircore.Vulnerability) (string, []byte) {
+	h := md5.New()
+	h.Write([]byte(v.Name))
+	h.Write([]byte(v.Description))
+	h.Write([]byte(v.Links))
+	h.Write([]byte(v.Severity))
+	if v.Package != nil {
+		h.Write([]byte(v.Package.Name))
```
ldelossa: this performs a copy for every field value, including descriptions. 
it's also in the hot path of a write. I think we should use go's optimization: http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#string_byte_slice_conv

internal/vulnstore/postgres/getupdateoperationdiff.go
```diff
@@ -0,0 +1,162 @@
...
+	dist_cpe,
+	dist_pretty_name,
+	repo_name,
+	repo_key,
+	repo_uri
+FROM vuln
+WHERE
+	vuln.id IN (
+		SELECT vuln AS id FROM uo_vuln WHERE uo = lhs.id
+		EXCEPT ALL
```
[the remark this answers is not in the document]
hdonnay: Oh, like an outer join?

internal/vulnstore/postgres/getupdateoperationdiff.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+const (
+	// SelectUpdateOperation retrieves an UpdateOperation.
+	selectUpdateOperation = `SELECT updater, fingerprint, date FROM update_operation WHERE ref = $1;`
+
+	// DiffQuery takes two update IDs and returns rows that only exist in first
+	// argument's set of vulnerabilities.
+	diffQuery = `WITH
```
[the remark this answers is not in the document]
hdonnay: Yeah, sorry.

internal/vulnstore/postgres/getupdateoperationdiff.go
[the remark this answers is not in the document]
ldelossa: EXCEPT seems to be equivalent to LEFT OUTTER JOIN with a null check:

```
SELECT * FROM tableA a 
  LEFT JOIN tableB b
ON a.Key = b.Key
WHERE B.Key IS NULL
```

internal/vulnstore/postgres/querybuilder.go
```diff
@@ -19,13 +19,14 @@ func buildGetQuery(record *claircore.IndexRecord, opts *vulnstore.GetOpts) (stri
 	matchers := opts.Matchers
 	psql := goqu.Dialect("postgres")
 	exps := []goqu.Expression{}
-	// add Package.Name as first condition in query
+
+	// add package name as second condition in query
```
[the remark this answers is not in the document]
hdonnay: This snuck back in during a rebase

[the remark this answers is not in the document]
ldelossa: up to you, I tend to like using the code constructs in the comment, but one could argue the code will change and the comments will not.

internal/vulnstore/postgres/getupdateoperationdiff.go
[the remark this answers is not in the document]
hdonnay: I take your point, but I'm going to hold off on measuring this and just opt for the (IMO) more readable version with `EXCEPT`.

internal/vulnstore/postgres/updatevulnerabilities.go
hdonnay: I think I can solve this by just using a `bytes.Buffer`. Looking at the code, it's just a `copy` in the happy path.

ldelossa: Cool, a single copy is better then copying each field. If that strikes a balance between readability and perf, would be a good move.

internal/vulnstore/postgres/getupdateoperationdiff.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/jackc/pgx/v4/pgxpool"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+func getUpdateDiff(ctx context.Context, pool *pgxpool.Pool, prev, cur uuid.UUID) (*driver.UpdateDiff, error) {
+	// Query takes two update IDs and returns rows that only exist in first
+	// argument's set of vulnerabilities.
+	const query = `WITH
```
ldelossa: don't we want this defined at the package level so it is not constructed on every function call?

internal/vulnstore/postgres/getupdateoperationdiff.go
```diff
@@ -0,0 +1,147 @@
...
+		repo_uri
+	FROM vuln
+	WHERE
+		vuln.id IN (
+			SELECT vuln AS id FROM uo_vuln JOIN lhs ON (uo_vuln.uo = lhs.id)
+			EXCEPT ALL
+			SELECT vuln AS id FROM uo_vuln JOIN rhs ON (uo_vuln.uo = rhs.id));`
+
+	if cur == uuid.Nil {
+		return nil, errors.New("nil uuid is an invalid endpoint")
```
ldelossa: invalid endpoint? "invalid update operation"?

internal/vulnstore/postgres/getupdateoperationdiff.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, err
+	}
+	rows.Close() // OK according to the docs.
+
+	// If we're starting at the beginning of time, nothing is going to
+	// be removed.
+	if prev == uuid.Nil {
+		return &diff, nil
+	}
+	rows, err = pool.Query(ctx, query, prev, cur)
```
ldelossa: is it safe to re-use this var after deferring it's Close() above?

internal/vulnstore/postgres/getupdateoperations.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/rs/zerolog"
+
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+// GetLatestUpdateRef implements driver.Updater.
+func (s *Store) GetLatestUpdateRef(ctx context.Context) (uuid.UUID, error) {
```
ldelossa: Is this used? Libvuln calls store.GetLatestUpateRef, and that calls the private method below this

internal/vulnstore/postgres/getupdateoperations.go
```diff
@@ -0,0 +1,150 @@
...
+		case err == nil:
+		case errors.Is(err, pgx.ErrNoRows):
+			rows.Close()
+		default:
+			rows.Close()
+			return nil, fmt.Errorf("failed to get distinct updates: %w", err)
+		}
+		// use a closure to defer rows.Close() and ensure connection can be re-used
+		// see: https://pkg.go.dev/github.com/jackc/pgx?tab=doc#Rows
+		err = func() error {
```
ldelossa: I saw you use another pattern above, does defering rows.Close and then explicitly calling rows.Close() if an early return is not necessary apply here?

internal/vulnstore/postgres/querybuilder_test.go
```diff
@@ -149,6 +165,13 @@ func Test_GetQueryBuilder_Deterministic_Args(t *testing.T) {
 			},
 		},
 	}
+
+	// This is safe to do because SQL doesn't care about what whitespace is
+	// where.
+	//
+	// Also, it produces more intelligible diffs when things break.
+	normalizeWhitespace := cmpopts.AcyclicTransformer("normalizeWhitespace", strings.Fields)
```
ldelossa: clever!!

internal/vulnstore/postgres/updatevulnerabilities.go
```diff
@@ -0,0 +1,195 @@
...
+		Msg("update_operation committed")
+	return ref, nil
+}
+
+// Md5Vuln creates an md5 hash from the members of the passed-in Vulnerability,
+// giving us a stable, context-free identifier for this revision of the
+// Vulnerability.
+func md5Vuln(v *claircore.Vulnerability) (string, []byte) {
+	var b bytes.Buffer
+	b.WriteString(v.Name)
```
ldelossa: was looking at WriteString and stubled upon the copy comments 
```
// The copy built-in function copies elements from a source slice into a
// destination slice. (As a special case, it also will copy bytes from a
// string to a slice of bytes.) The source and destination may overlap. Copy
// returns the number of elements copied, which will be the minimum of
// len(src) and len(dst).
func copy(dst, src []Type) int
```

sounds like copy() alone could handle this function as it copies strings to byte slices. but we may be responsible for growing the array. this is fine as is, just food for thought.

internal/vulnstore/postgres/getupdateoperations.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	log.Debug().
+		Int("count", len(ret)).
+		Msg("found updaters")
+	return ret, nil
+}
+
+func getUpdateOperations(ctx context.Context, pool *pgxpool.Pool, updater ...string) (map[string][]driver.UpdateOperation, error) {
+	const (
+		query       = `SELECT ref, updater, fingerprint, date FROM update_operation WHERE updater = $1 ORDER BY id DESC;`
```
ldelossa: I am not sure I like the var "query", it gives no value to the intent of the query being used. also, this will be created eac and everytime the method is called. Maybe compiler will optimize away tho, not sure.

internal/vulnstore/postgres/getupdateoperationdiff.go
hdonnay: It's not constructed on every call. The placement of constants only affects their scope.

internal/vulnstore/postgres/getupdateoperations.go
hdonnay: This is code just carried forward. I can update.

internal/vulnstore/postgres/getupdateoperationdiff.go
hdonnay: Yes; defer binds immediately

internal/vulnstore/postgres/getupdateoperations.go
hdonnay: This will be used by clair for conditional requests. For example, a client could issue a GET with `If-None-Match: <validator>`, and if we use the absolute latest update operation as that validator, we can know there's no possible way the result has changed, and short-circuit the response.

internal/vulnstore/postgres/querybuilder_test.go
hdonnay: Yeah, pretty happy with this one. Makes problems the tests _much_ more readable.

Issue #128, opened 2020-02-26
Add support for modular rpm


Allda:
Modular rpms are a special type of rpms which require different vulnerability detection.

Modular rpms are stored in the rpm database and can be distinguished from normal rpms by `RPMTAG_MODULARITYLABEL` label.

Clair v3 already has a modular rpm detection: https://github.com/quay/clair/pull/891

In oval vulnerability data, modular rpms can be also distinguished by looking at criterion comments.

`<criterion test_ref="oval:com.redhat.rhsa:tst:20190980027" comment="Module httpd:2.4 is enabled"/>`

ldelossa:
@Allda would you have the time to work this into ClairCore?

Allda:
This feature has been already implemented in the following PRs.
https://github.com/quay/claircore/pull/156 https://github.com/quay/claircore/pull/148

PR #148, opened 2020-03-27
Add module property to Package

Allda:
Module property stores information about package module and stream which
the package is part of.

This is mostly used for modular rpm content. Rpm parser gets module
info from RPMTAG_MODULARITYLABEL in rpm database.

### DIFF
--- a/internal/indexer/postgres/bootstrap.sql
+++ b/internal/indexer/postgres/bootstrap.sql
@@ -60,9 +60,10 @@ CREATE TABLE package (
     id SERIAL PRIMARY KEY,
     name text NOT NULL,
     kind text NOT NULL,
-    version text NOT NULL
+    version text NOT NULL,
+    module text NOT NULL
 );
-CREATE UNIQUE INDEX package_unique_idx ON package (name, version, kind);
+CREATE UNIQUE INDEX package_unique_idx ON package (name, version, kind, module);
 
 --- PackageScanArtifact
 --- A relation linking discovered packages with the 
--- a/libindex/migrations/migration1.go
+++ b/libindex/migrations/migration1.go
@@ -67,9 +67,10 @@ const (
 		kind text NOT NULL,
 		version text NOT NULL,
 		norm_kind text,
-		norm_version integer[10]
+		norm_version integer[10],
+		module text NOT NULL
 	);
-	CREATE UNIQUE INDEX IF NOT EXISTS package_unique_idx ON package (name, version, kind);
+	CREATE UNIQUE INDEX IF NOT EXISTS package_unique_idx ON package (name, version, kind, module);
 
 	--- PackageScanArtifact
 	--- A relation linking discovered packages with the 
--- a/package.go
+++ b/package.go
@@ -21,4 +21,6 @@ type Package struct {
 	// correctly ordered when compared with other representations from the same
 	// producer.
 	NormalizedVersion Version `json:"normalized_version,omitempty"`
+	// Module and stream which this package is part of
+	Module string `json:"module,omitempty"`
 }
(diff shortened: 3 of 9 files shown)


Allda:
@ldelossa This is a modular rpm support in indexer which we talked about few weeks ago and it partially relates to https://github.com/quay/claircore/issues/128

package.go
```diff
@@ -21,4 +21,6 @@ type Package struct {
 	// correctly ordered when compared with other representations from the same
 	// producer.
 	NormalizedVersion Version `json:"normalized_version,omitempty"`
+	// Module and stream which this package is part of
+	Module string `json:"module,omitempty"`
```
ldelossa: is it okay to couple module and stream into a single string? Or should they be two separate fields?

rpm/packagescanner.go
```diff
@@ -240,6 +240,7 @@ const queryFmt = `%{name}\n` +
 	`%{payloaddigestalgo}:%{payloaddigest}\n` +
 	`%{sigpgp:pgpsig}\n` +
 	`%{sourcerpm}\n` +
+	`%{RPMTAG_MODULARITYLABEL}\n` +
```
ldelossa: Do we need to inform users to use a specific rpm version (or geater) to support this module query?

Allda: Yes the label was introduced in rpm-4.15.0-alpha.

package.go
Allda: I was thinking about it, but I didn't find a use-case of having it in separate fields. Basically it is very similar to a version property (it also contains epoch, version and release, but it is represented as a simple string)

rpm/packagescanner.go
ldelossa: noted. thanks.

internal/indexer/postgres/bootstrap.sql
```diff
@@ -60,9 +60,10 @@ CREATE TABLE package (
     id SERIAL PRIMARY KEY,
     name text NOT NULL,
     kind text NOT NULL,
-    version text NOT NULL
+    version text NOT NULL,
+    module text,
 );
-CREATE UNIQUE INDEX package_unique_idx ON package (name, version, kind);
+CREATE UNIQUE INDEX package_unique_idx ON package (name, version, kind, module);
```
ldelossa, line 66: @Allda this causes an little bit of an issue. 

We are trying to keep the package table de duplicated and small. In order to do this we use a unique constraint. However since you are not requiring module to "not be null" this will allow duplicates in the database for records such as 

{ "package-a", "binary", "v0.0.1", null } 

With the schema as is, I could enter that many times and take up considerable amount of space over time. 

What we've done to avoid this is set it to NOT NULL and our code inputs the zero value '""' empty string. In that case inserting 

{ "package-a", "binary", "v0.0.1", "" }   <- if module NOT NULL, this is only inserted ONCE.

libindex/migrations/migration1.go
```diff
@@ -67,9 +67,10 @@ const (
 		kind text NOT NULL,
 		version text NOT NULL,
 		norm_kind text,
-		norm_version integer[10]
+		norm_version integer[10],
+		module text
```
ldelossa, line 71: same comment for bootstrap.sql

internal/indexer/postgres/bootstrap.sql
```diff
@@ -60,9 +60,10 @@ CREATE TABLE package (
     id SERIAL PRIMARY KEY,
     name text NOT NULL,
     kind text NOT NULL,
-    version text NOT NULL
+    version text NOT NULL,
+    module text,
```
ldelossa, line 64: @Allda this causes an little bit of an issue.

We are trying to keep the package table de duplicated and small. In order to do this we use a unique constraint. However since you are not requiring module to "not be null" this will allow duplicates in the database for records such as

{ "package-a", "binary", "v0.0.1", null }

With the schema as is, I could enter that many times and take up considerable amount of space over time.

What we've done to avoid this is set it to NOT NULL and our code inputs the zero value "" empty string. In that case inserting

{ "package-a", "binary", "v0.0.1", "" } <- if module NOT NULL, this is only inserted ONCE.

Allda: I added NOT NULL and use "" as a default which prevents from inserting duplicated items.

PR #149, opened 2020-03-27
CCTool Unpack

ldelossa:
This PR implements a command to unpack a particular image reference to your temp directory.
I keep having to do this when detection fails for a particular image and this slipstreams that process

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ vendor
 *.swp
 book
 cctool
+!cctool/
--- a/cmd/cctool/main.go
+++ b/cmd/cctool/main.go
@@ -14,6 +14,7 @@ import (
 var cleanup sync.WaitGroup
 
 type commonConfig struct {
+	cleanup *sync.WaitGroup
 }
 
 type subcmd func(context.Context, *commonConfig, []string) error
@@ -33,7 +34,9 @@ func main() {
 		done()
 	}()
 
-	var cfg commonConfig
+	var cfg commonConfig = commonConfig{
+		cleanup: &cleanup,
+	}
 	fs := flag.NewFlagSet("main", flag.ExitOnError)
 	fs.Usage = func() {
 		out := fs.Output()
@@ -44,6 +47,8 @@ func main() {
 		fmt.Fprintln(out, "\tgenerate reports for containers provided as arguments or on stdin")
 		fmt.Fprintln(out, "manifest")
 		fmt.Fprintln(out, "\tgenerate manifests for containers provided as arguments or on stdin")
+		fmt.Fprintln(out, "unpack")
+		fmt.Fprintln(out, "\textracts each container's layers content for inspection")
 		fmt.Fprintln(out)
 	}
 
@@ -57,6 +62,8 @@ func main() {
 		cmd = Report
 	case "manifest":
 		cmd = Manifest
+	case "unpack":
+		cmd = Unpack
 	case "":
 		fs.Usage()
 		os.Exit(99)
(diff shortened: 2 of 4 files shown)


cmd/cctool/unpack.go
```diff
@@ -0,0 +1,137 @@
...
+func Unpack(cmd context.Context, cfg *commonConfig, args []string) error {
+	cmdcfg := unpackConfig{}
+	fs := flag.NewFlagSet("cctool unpack", flag.ExitOnError)
+	fs.DurationVar(&cmdcfg.timeout, "timeout", 5*time.Minute, "timeout for successful http responses")
+	fs.Usage = func() {
+		out := fs.Output()
+		fmt.Fprintf(out, "Usage:\n")
+		fmt.Fprintf(out, "\tcctool unpack <image-ref>\n")
+		fmt.Fprintf(out, "Arguments:\n")
+		fmt.Fprintf(out, "\timage-ref: a reference to an image and it's repository\n\n")
```
hdonnay: s/'//

.gitignore
```diff
@@ -2,3 +2,4 @@ vendor
 *.swp
 book
 cctool
+!cctool/
```
hdonnay: Why is this needed?

cmd/cctool/main.go
```diff
@@ -14,6 +14,7 @@ import (
 var cleanup sync.WaitGroup
 
 type commonConfig struct {
+	cleanup *sync.WaitGroup
```
hdonnay: removing the pointer means the zero-value of commonConfig is OK to use

cmd/cctool/unpack.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	log.Printf("successfully fetched layers")
+
+	// create a tmp dir we will unpack layers to
+	td, err := ioutil.TempDir("", "cctool-unpack-")
+
+	log.Printf("exacting layers into tmp dir: %v.", td)
+	for i, layer := range m.Layers {
+		dirName := filepath.Base(layer.Local())
+		dir := filepath.Join(td, strconv.Itoa(i)+"-"+dirName)
```
hdonnay: I think it'd be easier to use the layer digest instead of grabbing the local file name.

.gitignore
ldelossa: i assume the initial intent was to not include cctool binary. But by doing this that regex also matches the dir name and ignores any modifications to files children of /cmd/cctool

cmd/cctool/main.go
ldelossa: Is it okay to copy the wait group between method calls? I defaulted to the sync.Mutex rules where you dont want to copy the mutex

.gitignore
hdonnay: ah yeah whoops

cmd/cctool/main.go
hdonnay: the commonConfig is passed by a pointer to the subcommand, so it should be fine.

ldelossa: hmm, imo the usage of the cleanup waitgroup shouldn't break by you deciding to pass a reference or copy of the containing struct to a method. that seems like a hard to catch bug?

hdonnay: you'd have to purposefully copy the struct, as it comes in as a pointer

ldelossa: Well I get that, but if you wrote a function that *didn't* to that, you'd break the usage of commonConfig, which is weird.

layer.go
```diff
@@ -28,6 +28,10 @@ func (l *Layer) SetLocal(f string) error {
 	return nil
 }
 
+func (l *Layer) Local() string {
+	return l.localPath
+}
+
```
hdonnay, line 34:
```suggestion
```

PR #153, opened 2020-03-31
Add architecture property to Package

Allda:
Arch property is added to Package structure. Architecture is also parsed
from package database and stored in index report.

Architecture should be also used in vulnerability scanner since some
vulnerabilities are arch specific.


ldelossa:
@Allda is this pr dependent on: https://github.com/quay/claircore/pull/148 ?

Allda:
@ldelossa This PR doesn't directly depend on #148, but since both updates the same files/line I build this PR on top of the previous one. Once #148 is merged I'll rebase on top of master.

Please review only the last commit.

ldelossa:
Sounds good @Allda lets get #148 merged in this tackle this one.

PR #154, opened 2020-04-03
Parse architecture from oval file

Allda:
Rpm package architecture is parsed from oval files and stored in a vuln database table.

Architecture value depends on architecture operation (equal, not-equal, pattern match, ...) so operation is also stored in DB as an enum. This can be later used during DB query to select only the right architecture.


ldelossa:
@Allda some merge conflicts.

Allda:
@ldelossa it's ready for review

package.go
```diff
@@ -26,3 +26,8 @@ type Package struct {
 	// Package architecture
 	Arch string `json:"arch,omitempty"`
 }
+
+const (
```
ldelossa, line 30: :+1:

vulnerability.go
```diff
@@ -31,5 +60,6 @@ type Vulnerability struct {
 	// a string specifying the package version the fix was relased in
 	FixedInVersion string `json:"fixed_in_version"`
 	// Range describes the range of versions that are vulnerable.
-	Range *Range `json:"range,omitempty"`
+	Range         *Range `json:"range,omitempty"`
+	ArchOperation ArchOp `json:"arch_op,omitempty"`
```
ldelossa, line 64: Comment this.

PR #156, opened 2020-04-09
Rpm module vulnerability matching

Allda:
This pull request contains support for modular rpm vulnerability matching.

Modular rpm is a special type of rpm package which needs to be handled in a different way from normal rpms. Oval contains module metadata that determines the affected module + stream. Module metadata is stored in a vuln table and queried using new MatchConstraint.


libvuln/migrations/migration1.go
```diff
@@ -54,6 +54,7 @@ const (
 		normalized_severity    TEXT,
 		package_name           TEXT,
 		package_version        TEXT,
+		package_module         TEXT,
```
ldelossa, line 57: think this needs to be added to the index as it will be queried.

pkg/ovalutil/rpm.go
```diff
@@ -24,7 +25,6 @@ func RPMDefsToVulns(ctx context.Context, root oval.Root, protoVuln ProtoVulnFunc
 		Logger()
 	ctx = log.WithContext(ctx)
 	vulns := make([]*claircore.Vulnerability, 0, 10000)
-	pkgcache := map[string]*claircore.Package{}
```
ldelossa, line 27: why the removal of this?

pkg/ovalutil/rpm.go
```diff
@@ -105,3 +108,16 @@ func walkCriterion(ctx context.Context, node *oval.Criteria, cris *[]*oval.Crite
 		*cris = append(*cris, &c)
 	}
 }
+
+func getEnabledModules(cris []*oval.Criterion) []string {
+	enabledModules := []string{}
+	for _, criterion := range cris {
+		var regexComment = regexp.MustCompile(`(Module )(.*)( is enabled)`)
```
ldelossa, line 115: I'd move this to an init function in the module, no need to build the regex var everytime this method is called. 
something like... 
```
package ovalutil

var regexComment

func init() {
  regexComment = regex.MustCompile(...)
}
```

pkg/ovalutil/rpm.go
Allda: The package cache used a package name as a cache key which doesn't work because the module name is also part of Package struct so I decided to remove it. What do you think is a better solution? And when exactly and how often is this cache used?

ldelossa: @Allda can the key be updated to "name" + "module" and the uniqueness of a package retained? 

The cache is just a automatic var, it comes and goes with the invocation of this function, however it keeps us from allocating a ton of pointers.

pkg/ovalutil/rpm.go
Allda: Fixed

pkg/ovalutil/rpm.go
Allda: Fixed

libvuln/migrations/migration1.go
Allda: Fixed

Issue #157, opened 2020-04-14
Ubuntu Precise and Artful urls 404


ldelossa:
```
7:47PM WRN updater error error="updater ubuntu-precise-updater failed to update: failed to parse the fetched vulnerability database: failed to decode OVAL xml contents: expected element type <oval_definitions> but have <html>" component=libvuln/New
7:47PM WRN updater error error="updater ubuntu-artful-updater failed to update: failed to parse the fetched vulnerability database: failed to decode OVAL xml contents: expected element type <oval_definitions> but have <html>" 
```

@alexmurray did ubuntu stop hosting these oval sec dbs? The root site no longer has directory listing and ClairCore is failing to retrieve these.

alexmurray:
Yes, precise and artful are both very much end of life and are not supported. We only maintain and publish OVAL for the supported releases. See https://wiki.ubuntu.com/Releases for more details.

ldelossa:
@alexmurray that was my hunch, but thanks for confirming. Ill remove them from our code base.

ldelossa:
@alexmurray is there an API endpoint which tracks supported ubuntu releases? I know there is a website for user facing, but anything we can parse ?

alexmurray:
@ldelossa You could use the launchpad API for this - see https://lists.ubuntu.com/archives/ubuntu-release/2019-April/004757.html

PR #158, opened 2020-04-15
remove unsupported ubuntu

ldelossa:
precise and artful releases are no longer supported see:
https://github.com/quay/claircore/issues/157


ldelossa:
rethinking this a bit - probably do not want to remove unsupported os detection.

PR #161, opened 2020-04-21
Add Issued property to vulnerability

Allda:
Issued is a datetime that determines when a given vulnerability was
introduced (or when a fixed version was released). This date is important
to tell how long a package is vulnerable. The date is going to be used
for translating vulnerabilities to a container security score (A-F).


Allda:
@hdonnay @ldelossa Can you please review the PR? There is one test that is failing but it is not related to my changes.

PR #164, opened 2020-04-23
implement updater sets

ldelossa:
This PR creates a container object for working with sets of Updaters.
UpdaterSet data structures do not allow duplicate names to be added. 

Part of this work is to clean up libvuln for further initialization logic to occur.


libvuln/driver/updaterset.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// ErrExists is an error returned if the updater
+// already exists in the set.
+type ErrExists struct {
+	Updater []string
+}
+
+func (e ErrExists) Error() string {
+	return fmt.Sprintf("%v", e.Updater)
```
hdonnay, line 15: This error string should have something more than a list literal.

libvuln/driver/updaterset.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// UpdaterSet holds a deduped
+// set of updaters
+type UpdaterSet struct {
+	Set map[string]Updater
```
hdonnay, line 21: If users are meant to always use Add and Merge to manipulate the set, then this shouldn't be exported.

libvuln/opts.go
```diff
@@ -75,7 +75,7 @@ func (o *Opts) Parse() error {
 	// filter out updaters if regex was passed
 	if o.Run != "" {
 		var err error
-		o.Updaters, err = regexFilter(o.Run, o.Updaters)
+		o.Updaters.RegexFilter(o.Run)
 		if err != nil {
 			return fmt.Errorf("regex filtering of updaters failed: %w", err)
 		}
```
hdonnay, line 81: Error should be checked.

libvuln/driver/updaterset.go
```diff
@@ -0,0 +1,87 @@
...
+
+	for n, u := range set.Set {
+		s.Set[n] = u
+	}
+	return nil
+}
+
+// Updaters returns a map of updaters keyed by
+// their name.
+func (s *UpdaterSet) Updaters() map[string]Updater {
```
hdonnay, line 68: Since Updaters already have a Name method, this should probably return a slice of Updaters.

libvuln/opts.go
```diff
@@ -64,9 +64,9 @@ func (o *Opts) Parse() error {
 			&python.Matcher{},
 		}
 	}
-	if len(o.Updaters) == 0 {
+	if len(o.Updaters.Set) == 0 {
```
hdonnay, line 67: Thinking about airgap support, this needs to be able to express "no updaters". I suppose one would do it presently by providing a regexp that matches none of the defaults' names, but this seems roundabout.

libvuln/driver/updaterset.go
ldelossa:
```
fmt.Sprintf("the following updaters already exist in the set: %+v", e.Updater)
```
?

libvuln/opts.go
ldelossa: This is is the only method that uses the .Set field, and I agree we should probably not export Set, I can give the ```UpdaterSet``` a Length() field, or an Empty() field. I am thinking maybe Length is not that useful?

ldelossa:
@hdonnay latest commits add the ability to provide a pointer to a slice of strings. Each string will specify an updater set. 

If the slice is nil, all updaters are constructed and launched.
If the slice is empty, no updaters are launched
If the slice is not empty, only the specified updater sets are launched.

libvuln/opts.go
```diff
@@ -16,29 +17,70 @@ import (
...
+	// "alpine"
+	// "aws"
+	// "debian"
+	// "oracle"
+	// "photon"
+	// "pyupio"
+	// "rhel"
+	// "suse"
+	// "ubuntu"
+	UpdaterSets *[]string
```
hdonnay, line 53: This doesn't need to be a pointer.
```suggestion
	UpdaterSets []string
```

libvuln/opts.go
```diff
@@ -16,29 +17,70 @@ import (
...
-	Matchers []driver.Matcher
-	// returns the updaters to run on an interval
+	// A pointer to a slice of strings representing which
+	// updaters libvuln will create.
+	//
+	// If the pointer is nil all default UpdaterSets
+	// will be used
+	//
+	// If the pointe points to an empty array no UpdaterSets
+	// will run.
```
hdonnay, line 41:
```suggestion
	// UpdaterSets is a slice of strings representing which updaters libvuln
	// will create.
	//
	// If nil, the default UpdaterSets will be used.
```

libvuln/updatesets.go
```diff
@@ -0,0 +1,80 @@
...
+	"debian": debian.UpdaterSet,
+	"oracle": oracle.UpdaterSet,
+	"photon": photon.UpdaterSet,
+	"suse":   suse.UpdaterSet,
+	"ubuntu": ubuntu.UpdaterSet,
+}
+
+// UpdaterSets returns all UpdaterSets currently
+// supported by libvuln
+func updaterSets(ctx context.Context, sets *[]string) (driver.UpdaterSet, error) {
```
hdonnay, line 30:
```suggestion
func updaterSets(ctx context.Context, sets []string) (driver.UpdaterSet, error) {
```

libvuln/updatesets.go
```diff
@@ -0,0 +1,80 @@
...
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "libvuln/updaterSets").
+		Logger()
+
+	us := driver.NewUpdaterSet()
```
hdonnay, line 35: This can be restructured as `if sets == nil { sets = defaults }` and then a single loop merging the sets.

libvuln/opts.go
ldelossa: @hdonnay the zero value of the struct will instantiate the array, which would be confusing if you want the zero value to be "default" which means run all updaters.

ldelossa: nvm that's wrong.

libvuln/updatesets.go
ldelossa: defaultSets is a map of constructors, errors still need to be checked for each.

PR #169, opened 2020-04-30
Louis/layer scanned fix

ldelossa:
This squashes two bugs which were causing libindex to misidentify a layer's distribution.

First the photon distribution scanner was returning an empty distribution even when no distribution could be identified. This caused an empty distribution to be returned and bound to the index report. 
Also we change all distribution scanners to do the right thing and return an empty array when the file they want to parse is present, but does not contain values that they identify with. 

Second the sql query used in LayerScanned were erroneously identifying layers as already scanned and this layer would never be processed. 

Boy scout tasks around layer scanning logging and adding makefile entry exist in this PR as well.

### DIFF
--- a/alpine/distributionscanner.go
+++ b/alpine/distributionscanner.go
@@ -90,7 +90,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated alpine release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty slice is returned.
 func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]*claircore.Distribution, error) {
 	defer trace.StartRegion(ctx, "Scanner.Scan").End()
 	log := zerolog.Ctx(ctx).With().
@@ -111,7 +111,7 @@ func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]
 			return []*claircore.Distribution{dist}, nil
 		}
 	}
-	return nil, nil
+	return []*claircore.Distribution{}, nil
 }
 
 // parse attempts to match all alpine release regexp and returns the associated
--- a/aws/distributionscanner.go
+++ b/aws/distributionscanner.go
@@ -60,7 +60,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated AWS release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty slice is returned.
 func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]*claircore.Distribution, error) {
 	defer trace.StartRegion(ctx, "Scanner.Scan").End()
 	log := zerolog.Ctx(ctx).With().
@@ -83,7 +83,7 @@ func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]
 			return []*claircore.Distribution{dist}, nil
 		}
 	}
-	return nil, nil
+	return []*claircore.Distribution{}, nil
 }
 
 // parse attempts to match all AWS release regexp and returns the associated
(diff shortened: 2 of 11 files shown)


alpine/distributionscanner.go
```diff
@@ -90,7 +90,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated alpine release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty array is returned.
```
hdonnay:
```suggestion
// If the files are found but all regexp fail to match an empty slice is returned.
```

debian/distributionscanner.go
```diff
@@ -65,7 +65,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated Debian release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty array is returned.
```
hdonnay:
```suggestion
// If the files are found but all regexp fail to match an empty slice is returned.
```

aws/distributionscanner.go
```diff
@@ -60,7 +60,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated AWS release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty array is returned.
```
hdonnay:
```suggestion
// If the files are found but all regexp fail to match an empty slice is returned.
```

oracle/distributionscanner.go
```diff
@@ -74,7 +74,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated Oracle release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty array is returned.
```
hdonnay:
```suggestion
// If the files are found but all regexp fail to match an empty slice is returned.
```

photon/distributionscanner.go
```diff
@@ -67,7 +67,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated photon release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty array is returned.
```
hdonnay:
```suggestion
// If the files are found but all regexp fail to match an empty slice is returned.
```

rhel/distributionscanner.go
```diff
@@ -78,7 +78,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated Oracle release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty array is returned.
```
hdonnay:
```suggestion
// If the files are found but all regexp fail to match an empty slice is returned.
```

suse/distributionscanner.go
```diff
@@ -80,7 +80,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated Suse release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty array is returned.
```
hdonnay:
```suggestion
// If the files are found but all regexp fail to match an empty slice is returned.
```

ubuntu/distributionscanner.go
```diff
@@ -85,7 +85,7 @@ func (*DistributionScanner) Kind() string { return scannerKind }
 // and perform a regex match for keywords indicating the associated Ubuntu release
 //
 // If neither file is found a (nil,nil) is returned.
-// If the files are found but all regexp fail to match an empty distribution is returned.
+// If the files are found but all regexp fail to match an empty array is returned.
```
hdonnay:
```suggestion
// If the files are found but all regexp fail to match an empty slice is returned.
```

PR #173, opened 2020-05-01
Louis/drone to gh actions

ldelossa:
This PR drops the drone CI in favor of github CI


ldelossa:
Will remove the drone gate right after this merge.

PR #163, opened 2020-04-23
Get Red Hat's repositories from container API

Allda:
This change fetches Red Hat CPE based repositories from Red Hat's
container API. This is a fallback solution in case repositories are not
stored directly in an image.
 
Once Red Hat releases embedded repositories this repo scanner will be
extended.

PR #175, opened 2020-05-04
Use repositories in linux coalescer

Allda:
This PR brings support of repositories in linux coalescer. Enabled repositories are associated with packages in the given introduced layer.

This also changes rhel matcher to use CPE based repositories instead of distribution.

This depends on https://github.com/quay/claircore/pull/163 and it needs to be reviewed/merged after that.


environment.go
```diff
@@ -14,5 +14,5 @@ type Environment struct {
 	// the ID of the distribution the package was discovered on
 	DistributionID string `json:"distribution_id"`
 	// the ID of the repository where this package was downloaded from (currently not used)
-	RepositoryID string `json:"repository_id"`
+	RepositoryIDs []string `json:"repository_ids"`
```
ldelossa, line 17: @Allda what's the reasoning behind making this an array?

An environment is a way to differentiate finding two of the same packages on the same host but in different file system locations. 

Assuming a package must have a single place of origin, I would assume a package could only have one Repository. Is this not the case?

internal/indexer/linux/coalescer.go
```diff
@@ -170,3 +188,44 @@ func (c *Coalescer) Coalesce(ctx context.Context, artifacts []*indexer.LayerArti
 	}
 	return c.ir, nil
 }
+
+// shareRepos takes repository definition and share it with other layers
+// where repositories are missing
+func (c *Coalescer) shareRepos(ctx context.Context, artifacts []*indexer.LayerArtifacts) {
```
ldelossa, line 194: @Allda This makes sense, but since this is RHEL specific I think we can take a slightly different approach.

We have a concept called "Ecosystem". an Ecosystem groups specific scanners with a particular Coalescer and runs a Coalescer on layers populated with artifacts *just* from those scanners. 

I'd like to see if we can do the following...

Take this coalescer and your modifications and create a "rhel/coalescer.go" file. It can be very similar to this one.
Create a new Ecosystem in "rhel/ecosystem.go". Define the components that are involved in this Ecosystem. Let's use this as an example: https://github.com/quay/claircore/blob/master/alpine/ecosystem.go#L11

In your case Ecosystem.PackageScanners would be the RPM package scanner, Ecosystem.DistributionScanner will be the RHEL distro scanner, Ecosystem.RepositoryScanners will be the RHEL repo scanner you created in this PR, and Ecosystem.Coalescer will be the RHEL coalescer you are introducing as well. 

By doing this we are making sure that RHEL coalescing method is only used with the RPM pkg scanner, RHEL distro scanner, and RHEL coalescer ensuing the rhel specific logic will not conflict with any other coalescing happening for the generic linux use case.

environment.go
Allda: If multiple repositories (CPEs) provide a the same package which is currently installed we can't assigned it to only one of them. That's the reason why repositories is an array.

ldelossa:
@Allda we just moved to gh actions, I have rebased your branch so the .github folder now exists. can you create a dummy commit to have the github action trigger?

Allda:
@ldelossa Based on our chat last week I merged all relevant PR into one and now it reports the vulnerabilities properly. There are still some changes in Linux coalescer, but it just de-duplicate code and doesn't change any logic.

One of the tests is still failing, but it seems it is not caused by my changes because locally it passes and I can't retrigger it here.

Allda:
@hdonnay I guess this PR can be closed in favor of https://github.com/quay/claircore/pull/192 , right?

PR #177, opened 2020-05-13
CPE plumbing

hdonnay:
This PR adds in a `cpe` package that implements the standard's Well-Formed Name rules, then uses that throughout the scanners' basic types.

The thinking is that anything that one of our scanners detects in a layer *could* be identified by a CPE (even if in practice, few are) so we should extract that data where ever we find it.

The WFN is used as a value throughout this PR. I don't have strong opinions on that, but it does end up with a lot of empty strings in the json representation.


pkg/cpe/wfn.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+			if i != 0 && i != last {
+				return fmt.Errorf("cpe: invalid position for special character: %q at %d", r, i)
+			}
+		case '?':
+			if esc { // if escaped, it's not special
+				break
+			}
+			qRun = true
+		case '\\':
```
ldelossa, line 91: why is this a double escape?

pkg/cpe/wfn.go
```diff
@@ -0,0 +1,209 @@
...
+			esc = true
+			// skip resetting the esc bool
+			continue
+		default:
+			// The only place valid for a run of '?' is at the beginning and end
+			// of a string. So if we read a series of '?' then a non-'?'
+			// *twice*, that's invalid.
```
ldelossa, line 98: so via the spec is '?somechars?` valid? and if it is, doesn't this not account for that?

pkg/cpe/wfn.go
```diff
@@ -0,0 +1,209 @@
...
+		return Value{}, err
+	}
+	return Value{
+		Kind: ValueSet,
+		V:    v,
+	}, nil
+}
+
+// Value represents all the states for an attribute's value.
+type Value struct {
```
ldelossa, line 129: nit: move struct def above constructor

pkg/cpe/wfn.go
```diff
@@ -0,0 +1,209 @@
...
+type ValueKind uint
+
+//go:generate stringer -type ValueKind
+
+// These are the valid states for a wfn attribute's value.
+const (
+	ValueUnset ValueKind = iota
+	ValueAny
+	ValueNA
+	ValueSet
```
ldelossa, line 144: `ValueSet` makes me think of set-theory 'set' and may get more confusing if we plan on using adding the matching specifications into things as well. if this name does not come directly from the spec consider another term.

pkg/cpe/wfn.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// Value represents all the states for an attribute's value.
+type Value struct {
+	Kind ValueKind
+	V    string
+}
+
+// ValueKind indicates what "kind" a value is.
+type ValueKind uint
```
ldelossa, line 135: maybe move the value bits to it's own file.

pkg/cpe/wfn.go
```diff
@@ -0,0 +1,209 @@
...
+		if v := &w.Attr[i]; v.Kind == ValueUnset {
+			unset++
+		}
+	}
+	if unset == NumAttr {
+		return ErrUnset
+	}
+	if p := w.Attr[int(Part)]; p.Kind == ValueSet {
+		if len(p.V) != 1 ||
+			(p.V != "a" && p.V != "o" && p.V != "h") {
```
ldelossa, line 180: nit: make these strings being compared consts, `Application` 'OS' 'HWDev`

pkg/cpe/wfn.go
```diff
@@ -0,0 +1,209 @@
...
+			return fmt.Errorf("cpe: wfn attr %v is invalid: %q is a disallowed value", Part, p.V)
+		}
+	}
+	return nil
+}
+
+// ErrUnset is returned from (WFN).Valid() if it is the zero value.
+var ErrUnset = errors.New("cpe: wfn is empty")
```
ldelossa, line 188: Why not just have an `IsZero` method that returns bool?

pkg/cpe/wfn.go
```diff
@@ -0,0 +1,209 @@
...
+
+func (w WFN) String() string {
+	return w.BindFS()
```
ldelossa, line 191: This is a little strange and also makes a computation happen on every call to String().

Would it make sense to create a WFN constructor, validate the WFN, Bind it, and set a Repr attribute all in one go? Then you no longer need to cll this w.BindFS() all the time, and you also don't need conflate whether a user should use String() or BindFS() method.

pkg/cpe/wfn.go
hdonnay: need to match a single `\`; can't use backticks because that's a string literal, and single quote allows for escaping single quotes, so the backslash needs to be doubled to get a literal backslash.

pkg/cpe/wfn.go
hdonnay: It is, and this does.

Hitting an unquoted `?` sets `qRun = true`. On the first normal character *after* the run of `?`s, it's checked to make sure the `?`s were in a valid position.

I do see a problem with a string ending in `?*`, though: `qRun` needs to be checked on any non-`?` character.

pkg/cpe/wfn.go
hdonnay: I had this called `Specific` at one point, but that's also wrong because ti may be a pattern. It's a clunky name but I think should be fine along with the type.

pkg/cpe/wfn.go
hdonnay: The string has to be created on every call because the structure isn't immutable. To make it so the string is only created once, the values would have to be made private and everything mediated by methods that keep the string up to date, instead of just creating the string on demand.

Having the WFN implement Stringer is just a nicety and we can remove this method if you think it'll be abused.

pkg/cpe/wfn.go
hdonnay: Personal preference for having a single method for checking validity.

If it had an `IsZero` method, any caller that wants to make a distinction between malformed and completely unset would need to make two calls, instead of one and just inspecting the returned error.

pkg/cpe/wfn.go
ldelossa: Should a constructed WFN be immutable?

PR #178, opened 2020-05-18
Louis/affected manifests

ldelossa:
This PR implements the ability for libindex to inform a client which manifests may be affected by a given list of vulnerabilities.


libindex/migrations/migration1.go
```diff
@@ -3,7 +3,7 @@ package migrations
 const (
 	// migration1 is the initial schema necessary for an indexer to operate
 	migration1 = `
-    --- Layer
+	--- Layer
```
hdonnay, line 6: whitespace change

libindex/migrations/migration1.go
```diff
@@ -154,5 +144,19 @@ const (
...
+    --- A searchable index of a coalesced manifest's content.
+    --- a package id is required.
+    --- either a dist_id or a repo_id maybe null, but not both
+    CREATE TABLE IF NOT EXISTS manifest_index
+    (
+        id            bigserial PRIMARY KEY,
+        package_id    bigint NOT NULL REFERENCES package (id),
+        dist_id       bigint REFERENCES dist (id),
+        repo_id       bigint REFERENCES repo (id),
+        manifest_hash text REFERENCES manifest (hash)
```
hdonnay, line 158: The not-null rule could be enforced by a table constraint:
```suggestion
        manifest_hash text REFERENCES manifest (hash),
        CHECK (dist_id IS NOT NULL AND repo_id IS NOT NULL)
```

ldelossa: Think we need OR logic here. 

dist_id not nil, repo_id not nil: OK (future packages with both)
dist_id nil, repo_id not nil: OK (python packages)
dist_id not nil, repo_id nil: OK (linux packages)
dist_id nil, repo_id nil: Not OK (shouldn't happen)

internal/indexer/postgres/affectedmanifest.go
```diff
@@ -0,0 +1,248 @@
...
+			v.Dist.VersionCodeName,
+			v.Dist.VersionID,
+		)
+		var id pgtype.Int8
+		err := row.Scan(&id)
+		if err != nil {
+			if err != pgx.ErrNoRows {
+				return protoRecord, fmt.Errorf("failed to scan dist: %v", err)
+			}
+		}
```
hdonnay, line 198: This should use `errors.Is` to make sure it's all unwrapped correctly:
```suggestion
		switch {
		case err == nil:
		case errors.Is(err, pgx.ErrNoRows): // OK
		default:
			return protoRecord, fmt.Errorf("failed to scan dist: %v", err)
		}
```

internal/indexer/postgres/affectedmanifest.go
```diff
@@ -0,0 +1,248 @@
...
+			v.Repo.Name,
+			v.Repo.URI,
+		)
+		var id pgtype.Int8
+		err := row.Scan(&id)
+		if err != nil {
+			if err != pgx.ErrNoRows {
+				return protoRecord, fmt.Errorf("failed to scan repo: %v", err)
+			}
+		}
```
hdonnay, line 229: Same thing here, use `errors.Is`

internal/indexer/postgres/indexmanifest.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	hash := ir.Hash.String()
+
+	records := ir.IndexRecords()
+	if len(records) == 0 {
+		log.Warn().Msg("manifest being indexed has 0 scan records")
+		return nil
+	}
+
+	// obtain a transaction scopped batch
```
hdonnay, line 39: scoped

internal/indexer/postgres/indexmanifest.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	defer tx.Rollback(ctx)
+
+	queryStmt, err := tx.Prepare(ctx, "queryStmt", query)
+	if err != nil {
+		return fmt.Errorf("failed to create statement: %v", err)
+	}
+
+	mBatcher := microbatch.NewInsert(tx, 500, time.Minute)
+	var ignored int
```
hdonnay, line 52: this is unused

internal/indexer/postgres/indexmanifest.go
```diff
@@ -0,0 +1,158 @@
...
+// toValues is a helper method which checks for
+// nil pointers inside an IndexRecord before
+// returning an associated pointer to the artifact
+// in question.
+//
+// v[0] source package id or nil
+// v[1] package id or nil
+// v[2] distribution id or nil
+// v[3] repository id or nil
+func toValues(r claircore.IndexRecord) ([4]*uint64, error) {
```
hdonnay, line 113: This should take the pointer to the record, instead of making a copy.

internal/indexer/postgres/indexmanifest.go
```diff
@@ -0,0 +1,158 @@
...
+	if r.Repository == nil {
+		res[3] = nil
+	} else {
+		id, err := strconv.ParseUint(r.Repository.ID, 10, 64)
+		if err != nil {
+			// return res, fmt.Errorf("repository id %v: %v", r.Package.ID, err)
+			return res, nil
+		}
+		res[3] = &id
+	}
```
hdonnay, line 155: all the pointers in the array are already nil, so all these ifs can have the conditionals inverted and the extra `nil` assignments removed.

internal/indexer/postgres/indexmanifest.go
ldelossa: @hdonnay since the method that calls this creates the IR (it is not passed in from an even higher method) I was trying to avoid passing pointers and completely use the stack as a general "don't gc as much".

hdonnay: but you're having to dereference a pointer to call this, which came out of a returned slice of pointers. the value is almost certainly on the heap

ldelossa: crap, got this confused with protoRecord method, nvm the record is already a pointer when it enters indexManifest.

libindex/migrations/migration1.go
ldelossa: I think this should stay tho, it aligns with the var.

internal/indexer/postgres/indexmanifest.go
ldelossa: I'm going to leave this copy, as its being used in the affectedManifest method with the attention described above.

internal/indexer/postgres/affectedmanifest.go
```diff
@@ -0,0 +1,269 @@
...
+	default:
+		return nil, err
+	}
+
+	// collect all packages which may be affected
+	// by the vulnerability in question.
+	pkgsToFilter := []claircore.Package{}
+	rows, err := pool.Query(ctx, selectPackages, v.Package.Name)
+	if err != nil {
+		rows.Close()
```
hdonnay, line 84:
```suggestion
	if rows != nil {
		defer rows.Close()
	}
	if err != nil {
```

This one isn't done in a loop or transaction, so I think the close can just be deferred.

internal/indexer/postgres/affectedmanifest.go
```diff
@@ -0,0 +1,269 @@
...
+			}
+			set[hash.String()] = hash
+		}
+	}
+
+	n := len(set)
+	out := make([]claircore.Digest, 0, n)
+	for _, v := range set {
+		out = append(out, v)
+	}
```
hdonnay, line 164: This could avoid an extra copy by making `set` a `map[string]struct{}`, then doing an existence check and an append if that fails here. Otherwise, this is copying every Digest into a map, then copying the last writes again into a slice.

internal/indexer/postgres/indexmanifest.go
```diff
@@ -0,0 +1,149 @@
...
+	records := ir.IndexRecords()
+	if len(records) == 0 {
+		log.Warn().Msg("manifest being indexed has 0 scan records")
+		return nil
+	}
+
+	// obtain a transaction scoped batch
+	tx, err := pool.Begin(ctx)
+	if err != nil {
+		return fmt.Errorf("store:indexDistributions failed to create transaction: %v", err)
```
hdonnay, line 42:
```suggestion
		return fmt.Errorf("postgres: indexManifest failed to create transaction: %v", err)
```

internal/indexer/postgres/indexmanifest.go
```diff
@@ -0,0 +1,149 @@
...
+		v, err := toValues(*record)
+		if err != nil {
+			return fmt.Errorf("received a record with an invalid id: %v", err)
+		}
+
+		// if source package exists create record
+		if v[0] != nil {
+			err = mBatcher.Queue(
+				ctx,
+				queryStmt.SQL,
```
hdonnay, line 67: does this being prepared actually help?

internal/indexer/postgres/store.go
```diff
@@ -92,6 +97,11 @@ func (s *store) IndexReport(ctx context.Context, hash claircore.Digest) (*clairc
 	return sr, b, err
 }
 
+func (s *store) AffectedManifests(ctx context.Context, v claircore.Vulnerability) ([]claircore.Digest, error) {
+	hashes, err := affectedManifests(ctx, s.pool, v)
+	return hashes, err
```
hdonnay, line 102:
```suggestion
	return affectedManifests(ctx, s.pool, v)
```

internal/indexer/postgres/store.go
```diff
@@ -67,6 +67,11 @@ func (s *store) IndexRepositories(ctx context.Context, repos []*claircore.Reposi
 	return err
 }
 
+func (s *store) IndexManifest(ctx context.Context, ir *claircore.IndexReport) error {
+	err := indexManifest(ctx, s.pool, ir)
+	return err
```
hdonnay, line 72:
```suggestion
	return indexManifest(ctx, s.pool, ir)
```

libindex/handler.go
```diff
@@ -25,10 +25,59 @@ func NewHandler(l *Libindex) *HTTP {
...
+	ctx := r.Context()
+	log := zerolog.Ctx(ctx).With().
+		Str("method", "index").
+		Logger()
+	ctx = log.WithContext(ctx)
+
+	if r.Method != http.MethodPost {
+		resp := &jsonerr.Response{
+			Code:    "method-not-allowed",
+			Message: "endpoint only allows GET",
```
hdonnay, line 43:
```suggestion
			Message: "endpoint only allows POST",
```

libindex/handler.go
```diff
@@ -25,10 +25,59 @@ func NewHandler(l *Libindex) *HTTP {
...
+	var vulnerabilities struct {
+		V []claircore.Vulnerability `json:"vulnerabilities"`
+	}
+	err := json.NewDecoder(r.Body).Decode(&vulnerabilities)
+	if err != nil {
+		resp := &jsonerr.Response{
+			Code:    "bad-request",
+			Message: err.Error(),
+		}
+		jsonerr.Error(w, resp, http.StatusBadGateway)
```
hdonnay, line 58: Did you mean `http.StatusBadRequest`?

libindex/libindex.go
```diff
@@ -158,3 +161,58 @@ func (l *Libindex) IndexReport(ctx context.Context, hash claircore.Digest) (*cla
...
+
+	var manifests struct {
+		sync.Mutex
+		m []claircore.Digest
+	}
+
+	groupSize := int(math.Sqrt(float64(len(vulns))))
+	if groupSize > maxGroupSize {
+		groupSize = maxGroupSize
+	}
```
hdonnay, line 182: Why do this instead of a fixed unit of work?

internal/indexer/controller/coalesce_test.go
```diff
@@ -0,0 +1,43 @@
...
+		t.Run(table.name, func(t *testing.T) {
+			ctx, done := context.WithCancel(ctx)
+			defer done()
+			indexer := New(&indexer.Opts{})
+
+			state, err := coalesce(ctx, indexer)
+			if err != nil {
+				t.Fatalf("did not expect error: %v", err)
+			}
+			assert.Equal(t, table.expectedState, state)
```
hdonnay, line 40: Use an `if` instead of assert

internal/indexer/postgres/affectedmanifest.go
```diff
@@ -0,0 +1,269 @@
...
+				END
+			)
+		  AND (
+			CASE
+				WHEN $3::bigint IS NULL THEN
+					repo_id IS NULL
+				ELSE
+					repo_id = $3
+				END
+			);
```
hdonnay, line 58: Clever

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/knqyf263/go-cpe v0.0.0-20180327054844-659663f6eca2
 	github.com/knqyf263/go-deb-version v0.0.0-20190517075300-09fca494f03d
 	github.com/knqyf263/go-rpm-version v0.0.0-20170716094938-74609b86c936
+	github.com/lib/pq v1.2.0
```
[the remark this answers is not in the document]
ldelossa: pa.Array is required to get the int data type in NomralizedVersion into the database.
we should and will perform a dedicated "convert to pgx" branch and this will quickly go away.

internal/indexer/controller/indexmanifest_test.go
```diff
@@ -0,0 +1,87 @@
...
+			s := table.mock(t)
+			indexer := New(&indexer.Opts{
+				Store: s,
+			})
+
+			state, err := indexManifest(ctx, indexer)
+			if err != nil {
+				t.Fatalf("did not expect error: %v", err)
+			}
+			assert.Equal(t, table.expectedState, state)
```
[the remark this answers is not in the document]
ldelossa: copy pasta'd. I'll update.

internal/indexer/postgres/affected_manifests_e2e_test.go
```diff
@@ -0,0 +1,294 @@
...
+			if err != nil {
+				t.Fatalf("could not decode ir: %v", err)
+			}
+		}
+		{
+			err := json.NewDecoder(vrFD).Decode(&vr)
+			if err != nil {
+				t.Fatalf("could not decode vr: %v", err)
+			}
+		}
```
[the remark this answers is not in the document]
ldelossa: these braces create a new lexical scope so the same block can be used inside, why? because I am lazy and like to copy and paste, and I don't like that a small change is necessary when err is defined in the same scope but you'd like to reuse the var.

libindex/handler.go
ldelossa: yup. thanks.

libindex/libindex.go
ldelossa: my thinking is creating equal sized concurrency groups proportionate to the length of the slice coming in allows a balance between concurrency and database connection usage (to some max).

internal/indexer/postgres/affectedmanifest.go
ldelossa: gotta thanks the IRC folks for this one.

internal/indexer/postgres/affected_manifests_e2e_test.go
```diff
@@ -0,0 +1,294 @@
...
+		// file name of vuln report in ./testdata
+		vrFName string
+	}{
+		// these fixtures have a subtly. these fixtures
+		// were generated against the same database
+		// to esure all ids are sequentially increasing.
+		//
+		// if you are to add fixtures here you must generate
+		// this current set + your new fixtures against the database
+		// to ensure no record id overlap.
```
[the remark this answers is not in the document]
ldelossa: no caps or punc in comments, but I'm fine with the re-wording.

internal/indexer/postgres/affectedmanifest.go
ldelossa: yeah I think you are right.

internal/indexer/postgres/indexmanifest.go
ldelossa: I don't think it can *hurt*, but haven't measured. in theory queries like this are good candidates for prepared statements, as we don't want to do the parsing and planning every single time we send the query to PG.

internal/indexer/postgres/store.go
ldelossa: this is following the pattern of all other methods in that file.

PR #182, opened 2020-05-26
Configuration groundwork

hdonnay:
These configuration knobs lay the groundwork for having scanners that are configurable and make RPCs, and for making updaters that are configurable.

I'd be fine with dropping the `alpine` commit out of this stack, it's there as a PoC.

### DIFF
--- a/internal/indexer/ecosystem.go
+++ b/internal/indexer/ecosystem.go
@@ -1,6 +1,10 @@
 package indexer
 
-import "context"
+import (
+	"context"
+
+	"github.com/rs/zerolog"
+)
 
 // Ecosystems group together scanners and a Coalescer which are commonly used together.
 //
@@ -17,7 +21,11 @@ type Ecosystem struct {
 }
 
 // EcosystemsToScanners extracts and dedupes multiple ecosystems and returns their discrete scanners
-func EcosystemsToScanners(ctx context.Context, ecosystems []*Ecosystem) ([]PackageScanner, []DistributionScanner, []RepositoryScanner, error) {
+func EcosystemsToScanners(ctx context.Context, ecosystems []*Ecosystem, disallowRemote bool) ([]PackageScanner, []DistributionScanner, []RepositoryScanner, error) {
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "internal/indexer/EcosystemsToScanners").
+		Logger()
+	ctx = log.WithContext(ctx)
 	ps := []PackageScanner{}
 	ds := []DistributionScanner{}
 	rs := []RepositoryScanner{}
@@ -29,10 +37,18 @@ func EcosystemsToScanners(ctx context.Context, ecosystems []*Ecosystem) ([]Packa
 			return nil, nil, nil, err
 		}
 		for _, s := range pscanners {
-			if _, ok := seen[s.Name()]; !ok {
-				ps = append(ps, s)
-				seen[s.Name()] = struct{}{}
+			n := s.Name()
+			if _, ok := seen[n]; ok {
+				continue
 			}
+			seen[n] = struct{}{}
+			if _, ok := s.(RPCScanner); ok && disallowRemote {
+				log.Info().
+					Str("scanner", n).
+					Msg("disallowed by configuration")
+				continue
+			}
+			ps = append(ps, s)
 		}
 
 		dscanners, err := ecosystem.DistributionScanners(ctx)
@@ -40,10 +56,18 @@ func EcosystemsToScanners(ctx context.Context, ecosystems []*Ecosystem) ([]Packa
 			return nil, nil, nil, err
 		}
 		for _, s := range dscanners {
-			if _, ok := seen[s.Name()]; !ok {
-				ds = append(ds, s)
-				seen[s.Name()] = struct{}{}
+			n := s.Name()
+			if _, ok := seen[n]; ok {
+				continue
 			}
+			seen[n] = struct{}{}
+			if _, ok := s.(RPCScanner); ok && disallowRemote {
+				log.Info().
+					Str("scanner", n).
+					Msg("disallowed by configuration")
+				continue
+			}
+			ds = append(ds, s)
 		}
 
 		rscanners, err := ecosystem.RepositoryScanners(ctx)
@@ -51,10 +75,18 @@ func EcosystemsToScanners(ctx context.Context, ecosystems []*Ecosystem) ([]Packa
 			return nil, nil, nil, err
 		}
 		for _, s := range rscanners {
-			if _, ok := seen[s.Name()]; !ok {
-				rs = append(rs, s)
-				seen[s.Name()] = struct{}{}
+			n := s.Name()
+			if _, ok := seen[n]; ok {
+				continue
 			}
+			seen[n] = struct{}{}
+			if _, ok := s.(RPCScanner); ok && disallowRemote {
+				log.Info().
+					Str("scanner", n).
+					Msg("disallowed by configuration")
+				continue
+			}
+			rs = append(rs, s)
 		}
 	}
 	return ps, ds, rs, nil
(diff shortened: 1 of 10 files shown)


internal/indexer/ecosystem.go
```diff
@@ -29,32 +29,47 @@ func EcosystemsToScanners(ctx context.Context, ecosystems []*Ecosystem) ([]Packa
...
 		for _, s := range pscanners {
-			if _, ok := seen[s.Name()]; !ok {
-				ps = append(ps, s)
-				seen[s.Name()] = struct{}{}
+			n := s.Name()
+			if _, ok := seen[n]; ok {
+				continue
 			}
+			seen[n] = struct{}{}
+			if _, ok := s.(RPCScanner); airgap && ok {
```
ldelossa, line 37: can we add an INFO() log to each scanner name which is excluded due to airgap?

internal/indexer/layerscanner/layerscanner.go
```diff
@@ -69,7 +70,7 @@ func (ls *layerScanner) Scan(ctx context.Context, manifest claircore.Digest, lay
 
 	ls.cc = make(chan struct{}, ccMin)
 
-	ps, ds, rs, err := indexer.EcosystemsToScanners(ctx, ls.Opts.Ecosystems)
+	ps, ds, rs, err := indexer.EcosystemsToScanners(ctx, ls.Opts.Ecosystems, ls.Opts.Airgap)
```
ldelossa, line 73: part of me wants the airgap filtering to be it's own method and not be done in EcosystemsToScanner as it'll read a little more declaratively in the construction of the Indexer. what do you think?

libvuln/init.go
```diff
@@ -28,6 +29,12 @@ func initUpdaters(ctx context.Context, opts *Opts, db *sqlx.DB, store vulnstore.
 			eC <- fmt.Errorf("duplicate updater found in UpdaterFactory. all names must be unique: %s", u.Name())
 			return
 		}
+		if c, ok := u.(driver.Configurable); ok {
```
ldelossa, line 32: I think this is going to take a little more thinking about. This is going to be an issue where we generate the names dynamically like Oracle (generating the date).

internal/indexer/versionedscanner.go
```diff
@@ -15,6 +21,23 @@ type VersionedScanner interface {
 	Kind() string
 }
 
+// ConfigDeserializer can be thought of as a curried Unmarshal function.
```
ldelossa, line 24:
```suggestion
// ConfigDeserializer can be thought of as an Unmarshal function.
```
don't believe curried is the right word here as it's just a closure over another method.

libvuln/driver/updater.go
```diff
@@ -29,19 +31,32 @@ type Parser interface {
...
 }
 
 // Unchanged is returned by Fetchers when the database has not changed.
 var Unchanged = errors.New("database contents unchanged")
 
-// Fingerprint is some identifiying information about a vulnerability database.
+// Fingerprint is some identifying information about a vulnerability database.
 type Fingerprint string
+
+// ConfigUnmarshaler can be thought of as a curried Unmarshal function, or a
```
ldelossa, line 53:
```suggestion
// ConfigUnmarshaler can be thought of as an Unmarshal function, or a
```

same

libvuln/opts.go
```diff
@@ -59,6 +60,11 @@ type Opts struct {
 	//
 	// This list will me merged with the default matchers.
 	Matchers []driver.Matcher
+	// Configs is a map of functions for configuration of Updaters.
+	Configs map[string]driver.ConfigUnmarshaler
```
ldelossa, line 64: May as well call this UpdaterConfigs if we ever need to pass Matcher configs down the line.

internal/indexer/versionedscanner.go
hdonnay: `Unmarshal` usually has the signature `func([]byte, interface{}) error`, so I was trying to point out it's that same concept with the `[]byte` closed over from a different scope.

internal/indexer/layerscanner/layerscanner.go
```diff
@@ -185,74 +186,53 @@ func (ls *layerScanner) scanDists(ctx context.Context, layer *claircore.Layer, s
...
+		r.dists, err = s.Scan(ctx, l)
+	case indexer.RepositoryScanner:
+		r.repos, err = s.Scan(ctx, l)
+	default:
+		panic(fmt.Sprintf("programmer error: unknown type %T used as scanner", s))
 	}
+	return err
+}
 
+func (r *result) Store(ctx context.Context, store indexer.Store, s indexer.VersionedScanner, l *claircore.Layer) error {
```
ldelossa, line 222: why exported?

internal/indexer/layerscanner/layerscanner.go
```diff
@@ -83,100 +55,128 @@ func (ls *layerScanner) Scan(ctx context.Context, manifest claircore.Digest, lay
...
-	if v == nil {
-		log.Debug().Msg("scan returned a nil")
-		return nil
+	ds = ds[:i]
+
+	sem := semaphore.NewWeighted(int64(ls.cLevel))
+	g, ctx := errgroup.WithContext(ctx)
+	// Scan is a closure to capture the values in the loops and then call the
+	// scan method.
+	scan := func(l *claircore.Layer, s indexer.VersionedScanner) func() error {
```
ldelossa, line 144: can you avoid this scan variable shadow with ls.scan?
put the sem on the layerScanner struct and make it sharable and then use it in ls.scan
seems a bit silly to have ls.scan and then a scan closure just to capture a sem that can be shared in he ls.scan method anyway.

hdonnay: It's not to capture the semaphore, it's to capture the loop variables.

ldelossa: can you make it so the process of performing a scan is self contained in layerScanner.Scan?
its simpler then trying to hop around between the Scan function, the scan variable, the function that is returned by the scan variable, and the ls.scan function.

internal/indexer/layerscanner/layerscanner.go
```diff
@@ -83,100 +56,128 @@ func (ls *layerScanner) Scan(ctx context.Context, manifest claircore.Digest, lay
...
-			ss := s
-			g.Go(func() error {
-				return ls.scanDists(gctx, ll, ss)
-			})
+	// Filter configures the provided scanner and reports if it should be
+	// filtered out of the slice or not.
+	//
+	// This is structured into a closure so that we can type-erase to the
+	// common VersionedScanner interface and only write all this logic once.
+	filter := func(s indexer.VersionedScanner) bool {
```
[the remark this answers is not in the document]
hdonnay: I don't think so.

internal/indexer/layerscanner/layerscanner.go
```diff
@@ -185,74 +186,53 @@ func (ls *layerScanner) scanDists(ctx context.Context, layer *claircore.Layer, s
...
+	repos []*claircore.Repository
+}
 
-	if len(v) > 0 {
-		log.Debug().Int("count", len(v)).Msg("scan returned repos")
-		err = ls.Store.IndexRepositories(ctx, v, layer, s)
-		if err != nil {
-			return fmt.Errorf("scanner: %v error: %v", s.Name(), err)
-		}
+func (r *result) Do(ctx context.Context, s indexer.VersionedScanner, l *claircore.Layer) error {
```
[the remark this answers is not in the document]
hdonnay: Well, the type it's implemented on isn't exported, but it's ExportCased just because it feels nice.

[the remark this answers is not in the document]
ldelossa: Thinks is better to be private to keep it internal usage

PR #189, opened 2020-06-02
updatediffs: fix broken query, add cctool subcommand

hdonnay:
This also normalizes the json names of the update structures.

### DIFF
--- a/internal/vulnstore/postgres/getupdateoperationdiff.go
+++ b/internal/vulnstore/postgres/getupdateoperationdiff.go
@@ -114,11 +114,11 @@ func populateRefs(ctx context.Context, diff *driver.UpdateDiff, pool *pgxpool.Po
 	const query = `SELECT updater, fingerprint, date FROM update_operation WHERE ref = $1;`
 	var err error
 
-	diff.B.Ref = cur
+	diff.Cur.Ref = cur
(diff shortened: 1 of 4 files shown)


libvuln/http/updatediff.go
```diff
@@ -0,0 +1,95 @@
...
+		Error error `json:"error"`
+	}{
+		Error: err,
+	}); err != nil {
+		panic(err)
+	}
+}
+
+// UpdateDiffs assumes it's anchored to a path segment ending in "/".
+func UpdateDiffs(l *libvuln.Libvuln) http.Handler {
```
ldelossa, line 27: Wouldn't it make sense to split this out?

```
const (
  updatesPath = "api/v1/updates"
  updatesDiff = "api/v1/updates/diff"
)

func UpdatesHandler() http.Handlerfunc {
  do the only thing you do when you go to the root of updates
}
func UpdateDiffsHandler() http.HandlerFunc {
  return func(w http.ResponseWriter, r *http.Request) {
     switch r.Method {
     case http.MethodGet:
       do your get stuff
     case http.MethodPut:
       do your put stuff
     case http.Delete:
       do your delete stuff
```

Now as *soon* as I look at the handler, I know what code is running for what http method.

PR #190, opened 2020-06-03
add affected manifest type, update plumbing of the type

ldelossa:
this commit introduces a new type to support the notifications
implementation along with tests and plumbing.


affectedmanifests.go
```diff
@@ -0,0 +1,58 @@
...
+func (a *AffectedManifests) Sort() {
+	for hash, ids := range a.VulnerableManifests {
+		sort.Slice(ids, func(i, j int) bool {
+			id1, id2 := ids[i], ids[j]
+			v1, v2 := a.Vulnerabilities[id1], a.Vulnerabilities[id2]
+			// reverse this since we want descending sort
+			if v1.NormalizedSeverity > v2.NormalizedSeverity {
+				return true
+			}
+			return false
```
hdonnay, line 54:
```suggestion
			return v1.NormalizedSeverity > v2.NormalizedSeverity
```

affectedmanifests.go
```diff
@@ -0,0 +1,58 @@
...
+		a.VulnerableManifests[hash] = append(a.VulnerableManifests[hash], v.ID)
+	}
+	a.mu.Unlock()
+}
+
+// Sort will sort each array in the VulnerableManifests map
+// by Vulnerability.NormalizedSeverity in Desc order.
+//
+// Sort should not be ran concurrently.
+func (a *AffectedManifests) Sort() {
```
hdonnay, line 45: This should just take the lock, then there's not a way to use the struct unsafely.

affectedmanifests.go
```diff
@@ -0,0 +1,58 @@
...
+		sort.Slice(ids, func(i, j int) bool {
+			id1, id2 := ids[i], ids[j]
+			v1, v2 := a.Vulnerabilities[id1], a.Vulnerabilities[id2]
+			// reverse this since we want descending sort
+			if v1.NormalizedSeverity > v2.NormalizedSeverity {
+				return true
+			}
+			return false
+		})
+		a.VulnerableManifests[hash] = ids
```
hdonnay, line 56: This is unnecessary, sort functions work in-place.

affectedmanifests.go
```diff
@@ -0,0 +1,58 @@
...
+	"sort"
+	"sync"
+)
+
+// AffectedManifests describes a set of manifests affected by
+// a set of Vulnerabilities.
+type AffectedManifests struct {
+	mu sync.Mutex
+	// map of vulnerabilities keyed by the vulnerability's ID
+	Vulnerabilities map[string]Vulnerability `json:"vulnerabilities"`
```
hdonnay, line 13: The value should be `*Vulnerability`, otherwise any growing the map will cause a *lot* of copying.

affectedmanifests.go
ldelossa: I didn't want to promote the use of Sort in multiple goroutines as it could potentially take some time and block others from doing work. what do you think about that?

affectedmanifests.go
ldelossa: was wondering that, will remove.

affectedmanifests.go
ldelossa: should I be concerned with polluting the heap vs copy speed here?

affectedmanifests.go
hdonnay: I think having some methods be concurrent-safe and some not is surprising. Either callers handle it or they don't, it shouldn't be a mix.

affectedmanifests.go
hdonnay: I mean, we have to get the Vulnerability from somewhere to pass in. The entrypoint into the affected manifests has them in a slice, but then ends up copying them all around. This and the `Add` method should just take pointers, because the caller has packed them into a slice for us.

ldelossa: I think I see what you're saying, either way I am making copies when Add is called, those copies may as well be pointers, and then when the map needs to grow it's less expensive.

ldelossa:
we discussed the naming a bit out of band, I think affected vs effected could be debating given the order in which we think about it. we have a bunch of technical writing saying affected now and I'm not sure its a big enough point to go and change all that at this point.

PR #196, opened 2020-07-01
dynamic updaters

hdonnay:
This PR provides for having all updaters dynamically constructed upon use.

This allows databases to be discovered at runtime (e.g. RHEL's use of a Pulp repository) and for graceful degradation when distributions fall out of support (e.g. Ubuntu's non-LTS releases).


internal/updater/executor.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/rs/zerolog"
+
+	"github.com/quay/claircore/internal/vulnstore"
+	"github.com/quay/claircore/internal/vulnstore/postgres"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+const tryLock = `SELECT pg_try_advisory_xact_lock($1);`
+
+func crushkey(key string) int64 {
```
ldelossa, line 21: can we please not do this? It is not helpful when troubleshooting to look at pg_advisory locks and have to decode to know if something is keeping a lock around needlessly.

hdonnay:
The way it's implemented, the pulp object isn't coupled to rhel at all. It's the rhel object that's coupled to the pulp object, because that's how the databases are delivered. I like the separate pulp object because it's a nice, self-contained bit of code.

The `init` function is only used to preserve the "default" behavior. I'll rework that to construct the factories in the libvuln constructor.

ldelossa:
But what's the purpose of self containing it, is what I'm failing to see. All you do is feed it an method implemented somewhere else to construct what you need. It being collapsed would make a lot more sense and pulp being placed inside rhel/ it will never be used outside RHEL.

PR #201, opened 2020-07-10
updaters: handle "unchanged" better

hdonnay:
The debian and ubunut updaters weren't actually checking the passed-in fingerprint, which was resulting in spurious updates in the database.

The alpine updater wasn't using an etag, which the remote server seems to support, so this change uses that.


hdonnay:
still need to spot-check the other updaters

alpine/fetcher.go
```diff
@@ -58,18 +58,16 @@ func (u *Updater) Fetch(ctx context.Context, hint driver.Fingerprint) (io.ReadCl
 
 	var r io.Reader = res.Body
 	if _, err := io.Copy(tf, r); err != nil {
+		tf.Close()
```
ldelossa, line 61:
```suggestion
		if tf != nil {
		  tf.Close()
		}
```

I've gotten into the habit of checking nil of these things first, may not be necessary.

pkg/ovalutil/fetcher.go
```diff
@@ -144,12 +145,32 @@ func (f *Fetcher) Fetch(ctx context.Context, hint driver.Fingerprint) (io.ReadCl
...
+func (f fingerprint) Set(h http.Header) {
+	if f.Etag != "" {
+		h.Set("if-none-match", f.Etag)
+	}
+	if f.Date != "" {
+		h.Set("if-modified-since", f.Date)
+	}
+}
+
+func (f *fingerprint) From(h http.Header) {
```
ldelossa, line 168: ```FromHeader``` ?

alpine/fetcher.go
hdonnay: In this case, tmp.File works such that if it didn't return an error, it can't return a nil pointer.

PR #202, opened 2020-07-16
matcher: Introduce Remote Matcher interface

arajkumar:
RemoteMatcher is an additional interface that a Matcher can implement.

When called the interface can invoke the remote matcher using the RESTful API
to fetch new vulnerabilities associated with the given IndexRecords.

The information retrieved from this interface **won't be persisted** into ClairCore database.


arajkumar:
https://github.com/arajkumar/claircore/pull/3 is a WIP PR which implements the RemoteMatcher to fetch Python vulnerabilities from CRDA platform.

internal/matcher/controller.go
```diff
@@ -45,6 +45,11 @@ func (mc *Controller) Match(ctx context.Context, records []*claircore.IndexRecor
 		return map[string][]*claircore.Vulnerability{}, nil
 	}
 
+	remoteMatcher, matchedVulns, err := mc.queryRemoteMatcher(ctx, interested)
+	if remoteMatcher {
+		return matchedVulns, err
```
ldelossa, line 50:
```suggestion
		if err != nil {
		  log.Error().Err(err).Msg("remote matcher error, returning empty results")
		}
		return map[string][]*claircore.Vulnerability, nil
```

Because http endpoints can be transient and unreachable on the network, we probably do not want the entire matching operation to fail, like it would here if you were to return he error upstream. 
Instead lets log the error and return an empty result set.

internal/matcher/controller.go
```diff
@@ -45,6 +45,11 @@ func (mc *Controller) Match(ctx context.Context, records []*claircore.IndexRecor
 		return map[string][]*claircore.Vulnerability{}, nil
 	}
 
+	remoteMatcher, matchedVulns, err := mc.queryRemoteMatcher(ctx, interested)
```
ldelossa, line 48: Let's derive a ctx here with some safe deadline and pass it to the mc.queryRemoteMatcher method.

ldelossa: @arajkumar - since these are HTTP requests, lets bump it up to 60s for now.

arajkumar: Sure, just fixed it. PTAL.

PR #203, opened 2020-07-17
remotematcher: Implement RemoteMatcher for CRDA

arajkumar:
This supplements https://github.com/quay/claircore/pull/202.

### Tasks
- [x] Add unit tests
- [x] Add concurrency control
- [x] Configurable HTTP client
- [x] Configurable Service host

### DIFF
--- a/crda/remotematcher.go
+++ b/crda/remotematcher.go
@@ -1,0 +1,251 @@
+package crda
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"golang.org/x/sync/errgroup"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/quay/zlog"
+	"go.opentelemetry.io/otel/baggage"
+	"go.opentelemetry.io/otel/label"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+var (
+	_ driver.Matcher       = (*Matcher)(nil)
+	_ driver.RemoteMatcher = (*Matcher)(nil)
+)
+
+const (
+	// Bounded concurrency limit.
+	defaultRequestConcurrency = 10
+	defaultURL                = "https://f8a-analytics-2445582058137.production.gw.apicast.io/?user_key=9e7da76708fe374d8c10fa752e72989f"
+	defaultPath               = "/api/v2/component-analyses/pypi/%s/%s"
+)
+
+// Matcher attempts to correlate discovered python packages with reported
+// vulnerabilities.
+type Matcher struct {
+	client             *http.Client
+	url                *url.URL
+	requestConcurrency int
+}
+
+// Build struct to model CRDA V2 ComponentAnalysis response which
+// delivers Snyk sourced Vulnerability information.
+type Vulnerability struct {
+	ID       string   `json:"vendor_cve_ids"`
+	CVSS     string   `json:"cvss"`
+	CVES     []string `json:"cve_ids"`
+	Severity string   `json:"severity"`
+	Title    string   `json:"title"`
+	URL      string   `json:"url"`
+	FixedIn  []string `json:"fixed_in"`
+}
+
+type Analyses struct {
+	Vulnerabilities []Vulnerability `json:"vulnerability"`
+}
+
+type VulnReport struct {
+	RecommendedVersion string   `json:"recommended_versions"`
+	Message            string   `json:"message"`
+	Analyses           Analyses `json:"component_analyses"`
+}
+
+// Option controls the configuration of a Matcher.
+type Option func(*Matcher) error
+
+// NewMatcher returns a configured Matcher or reports an error.
+func NewMatcher(opt ...Option) (*Matcher, error) {
+	m := Matcher{}
+	for _, f := range opt {
+		if err := f(&m); err != nil {
+			return nil, err
+		}
+	}
+
+	if m.url == nil {
+		var err error
+		m.url, err = url.Parse(defaultURL)
+		if err != nil {
+			return nil, err
+		}
+	}
+	if m.client == nil {
+		m.client = http.DefaultClient
+	}
+	// defaults to a sane concurrency limit.
+	if m.requestConcurrency < 1 {
+		m.requestConcurrency = defaultRequestConcurrency
+	}
+
+	return &m, nil
+}
+
+// WithClient sets the http.Client that the matcher should use for requests.
+//
+// If not passed to NewMatcher, http.DefaultClient will be used.
+func WithClient(c *http.Client) Option {
+	return func(m *Matcher) error {
+		m.client = c
+		return nil
+	}
+}
+
+// WithHost sets the server host name that the matcher should use for requests.
+//
+// If not passed to NewMatcher, defaultHost will be used.
+func WithURL(uri string) Option {
+	u, err := url.Parse(uri)
+	return func(m *Matcher) error {
+		if err != nil {
+			return err
+		}
+		m.url = u
+		return nil
+	}
+}
+
+// WithRequestConcurrency sets the concurrency limit for the network calls.
+//
+// If not passed to NewMatcher, a defaultRequestConcurrency will be used.
+func WithRequestConcurrency(requestConcurrency int) Option {
+	return func(m *Matcher) error {
+		m.requestConcurrency = requestConcurrency
+		return nil
+	}
+}
+
+// Name implements driver.Matcher.
+func (*Matcher) Name() string { return "crda" }
+
+// Filter implements driver.Matcher.
+func (*Matcher) Filter(record *claircore.IndexRecord) bool {
+	return record.Package.NormalizedVersion.Kind == "pep440"
+}
+
+// Query implements driver.Matcher.
+func (*Matcher) Query() []driver.MatchConstraint {
+	panic("unreachable")
+}
+
+// Vulnerable implements driver.Matcher.
+func (*Matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, vuln *claircore.Vulnerability) (bool, error) {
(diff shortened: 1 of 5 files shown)


crda/remotematcher.go
```diff
@@ -0,0 +1,258 @@
...
+
+var (
+	_ driver.Matcher       = (*Matcher)(nil)
+	_ driver.RemoteMatcher = (*Matcher)(nil)
+)
+
+const (
+	// Bounded concurrency limit.
+	defaultRequestConcurrency = 10
+	defaultURL                = "https://f8a-analytics-preview-2445582058137.production.gw.apicast.io/?user_key=3e42fa66f65124e6b1266a23431e3d08"
```
hdonnay, line 28: Is this okay for the default for everyone that'll ever use this? It seems odd to have "preview" in the subdomain and a "user_key" URL parameter.

crda/remotematcher.go
```diff
@@ -0,0 +1,258 @@
...
+	log.Debug().
+		Int("records", len(records)).
+		Msg("packages")
+
+	ctrlC, errorC := m.fetchVulnerabilities(ctx, records)
+	results := make(map[string][]*claircore.Vulnerability)
+	for r := range ctrlC {
+		for _, vuln := range r {
+			results[vuln.Package.ID] = append(results[vuln.Package.ID], vuln)
+			log.Debug().
```
ldelossa, line 159: would it make sense, for sake of "nice logging" to just log the number of items in the "results" array. this may get a little verbose.

crda/remotematcher.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+	}()
+	return ctrlC, errorC
+}
+
+func (m *Matcher) componentAnalyses(ctx context.Context, record *claircore.IndexRecord) ([]*claircore.Vulnerability, error) {
+	reqUrl := url.URL{
+		Scheme:   m.url.Scheme,
+		Host:     m.url.Host,
+		Path:     fmt.Sprintf("/api/v2/component-analyses/pypi/%s/%s", record.Package.Name, record.Package.Version),
```
ldelossa, line 211: probably want this string path as a constant declare higher in the file.

crda/remotematcher.go
```diff
@@ -0,0 +1,258 @@
...
+		Method:     http.MethodGet,
+		Header:     http.Header{"User-Agent": {"claircore/crda/RemoteMatcher"}},
+		URL:        &reqUrl,
+		Proto:      "HTTP/1.1",
+		ProtoMajor: 1,
+		ProtoMinor: 1,
+		Host:       reqUrl.Host,
+	}
+	// A request shouldn't go beyound 10s.
+	tctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
ldelossa, line 225: @hdonnay thoughts on 10 second timeout here?

crda/remotematcher.go
```diff
@@ -0,0 +1,258 @@
...
+)
+
+var (
+	_ driver.Matcher       = (*Matcher)(nil)
+	_ driver.RemoteMatcher = (*Matcher)(nil)
+)
+
+const (
+	// Bounded concurrency limit.
+	defaultRequestConcurrency = 10
```
ldelossa, line 27: this this be based on cpu count ? 

```
runtime.NumCPU()
```

crda/remotematcher.go
```diff
@@ -0,0 +1,258 @@
...
+		for _, vuln := range r {
+			results[vuln.Package.ID] = append(results[vuln.Package.ID], vuln)
+			log.Debug().
+				Str("package", vuln.Package.Name).
+				Str("version", vuln.Package.Version).
+				Str("id", vuln.ID).
+				Msg("vulns")
+		}
+	}
+	select {
```
ldelossa, line 166: I'm not sure this select statement is going to work as intended. Checking "ok" only tells you if the error channel has closed, not whether an error has returned or not. 

Since your code guarantees errorC will be closed, I think you simply want to block on the error channel, then check the error.

```
err := <-errorC // guaranteed to have an err or be closed
if err != nil {
   	log.Error().Err(err).Msg("access to component analyses failed")
}
```

Also to guarantee  that ctrlC is closed before errorC you can swap your defers on lines 185,186
```
defer close(errorC)
defer close(ctrlC)
```

these defers stack, so ctrlC will be closed, and then errorC. its not explicitly necessary but it aligns with the order of operations in QuayRemoteMatcher.

ldelossa (changes requested):
@hdonnay @arajkumar how will this work with airgap usage? did we build this into claircore to not call remote matchers if airgap is enabled? 

Other then this question, some review comments to look over.

crda/remotematcher.go
arajkumar: @ldelossa This bounded concurrency number is for http requests. Since it is an I/O bound operation, IMHO it may not be helpful to tying it with `runtime.NumCPU()`.

crda/remotematcher.go
arajkumar: +1, I will fix it ASAP.

crda/remotematcher.go
arajkumar: Ack!

crda/remotematcher.go
arajkumar: Ack!

crda/remotematcher.go
arajkumar: @hdonnay I had used preview/staging deployment to avoid messing up with our reports. 
"user_key" is not a secrete key, it is a publicity known 3scale gateway key.

crda/remotematcher.go
arajkumar: Fixed it.

crda/remotematcher.go
ldelossa: I see, you want to get them all in flight first, and then let the scheduler park them until io is ready? I could see the logic in that.

crda/remotematcher.go
```diff
@@ -0,0 +1,254 @@
...
+			results[vuln.Package.ID] = append(results[vuln.Package.ID], vuln)
+		}
+	}
+	select {
+	case err := <-errorC: // guaranteed to have an err or be closed
+		// Don't propagate error, log and move on.
+		if err != nil {
+			log.Error().Err(err).Msg("call to component analyses has failed")
+		}
+	default:
```
ldelossa, line 168: @arajkumar I believe you want to remove the select switch and "default" case all together here. It is no longer necessary, you do not want to have a "default" case, you want to unconditionally block on the errorC until it is either closed or an error is present on the channel.

ldelossa, line 168: @arajkumar I believe you want to remove the select switch and "default" case all together here. It is no longer necessary, you do not want to have a "default" case, you want to unconditionally block on the errorC until it is either closed or an error is present on the channel.

crda/remotematcher.go
```diff
@@ -0,0 +1,254 @@
...
+		Method:     http.MethodGet,
+		Header:     http.Header{"User-Agent": {"claircore/crda/RemoteMatcher"}},
+		URL:        &reqUrl,
+		Proto:      "HTTP/1.1",
+		ProtoMajor: 1,
+		ProtoMinor: 1,
+		Host:       reqUrl.Host,
+	}
+	// A request shouldn't go beyound 10s.
+	tctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
ldelossa, line 221: @arajkumar would it be safe to roll this back to 1 second timeout? do you know how long an average API request to your api will take?

ldelossa (changes requested):

crda/remotematcher.go
arajkumar: @ldelossa On avg it should take between 2 - 3 seconds, How about 5 seconds?

arajkumar: I can make it configurable.

ldelossa: Lets keep it out of config for now, but 5sec sounds good

arajkumar:
> how will this work with airgap usage? did we build this into claircore to not call remote matchers if airgap is enabled

I would prefer to leave this upto the configuration instead of turning it off by default for Airgap use. There is a [PR in pipeline](https://github.com/arajkumar/claircore/pull/4) which will make remote matcher configurable., I will publish once this PR is merged into mainline.

crda/remotematcher.go
arajkumar: Changed it to 5 sec.

crda/remotematcher.go
arajkumar: Fixed it.

crda/remotematcher.go
```diff
@@ -0,0 +1,251 @@
...
+
+// Vulnerable implements driver.Matcher.
+func (*Matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, vuln *claircore.Vulnerability) (bool, error) {
+	// RemoteMatcher can match Package and Vulnerability.
+	panic("unreachable")
+}
+
+// QueryRemoteMatcher implements driver.RemoteMatcher.
+func (m *Matcher) QueryRemoteMatcher(ctx context.Context, records []*claircore.IndexRecord) (map[string][]*claircore.Vulnerability, error) {
+	log := zerolog.Ctx(ctx).With().
```
ldelossa, line 147: @arajkumar can you rebase and use our new logging api:
https://quay.github.io/claircore/contributor/logging.html

ldelossa (changes requested):
Rebase to newest claircore and implement our new logging api, then I think this is good to go.

We still must find a way to disable remote matchers from running in airgap mode gracefully before this gets released into the wild.

crda/remotematcher.go
arajkumar: Fixed it.  Hope it looks better now.

ldelossa (approved):
LGTM - contingent on tests passing.

PR #204, opened 2020-07-21
Support of embedded content sets in RHEL repo scanner (matcher variant)

Allda:
This PR adds support of embedded content-set and translates them to CPE based repositories.
New features:

    Usage of a new content manifest file
    Mapping file between content-sets and CPEs is updated in regular intervals
    Configurable RHEL repo scanner
    Fewer queries to container API
    Translation of content-sets to CPEs is done in Matcher


rhel/matcher.go
```diff
@@ -1,37 +1,61 @@
...
+
 // Matcher implements driver.Matcher.
-type Matcher struct{}
+type Matcher struct {
+	mapping *repo2cpe.RepoCPEMapping
+}
 
 var _ driver.Matcher = (*Matcher)(nil)
 
+func NewMatcher() *Matcher {
```
ldelossa, line 25: When will this constructor return? 

Will it return only after the content mapping file is loaded into memory?
Will it return before the mapping file is loaded into memory?

If it returns before the mapping file is loaded into memory, calls to `m.mapping.RepositoryToCPE` should *not* error. If the mapping file is not in memory, the matching process should gracefully return no matches. 

Can you confirm we come up with a good error scenario for this?

pkg/omnimatcher/omnimatcher.go
```diff
@@ -42,11 +44,12 @@ func New(m []driver.Matcher) OmniMatcher {
 }
 
 // Vulnerable will call each Matcher's Vulnerable method until one returns true.
-func (om OmniMatcher) Vulnerable(record *claircore.IndexRecord, vuln *claircore.Vulnerability) bool {
+func (om OmniMatcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, vuln *claircore.Vulnerability) (bool, error) {
 	for _, m := range om {
-		if m.Vulnerable(record, vuln) {
-			return true
+		match, _ := m.Vulnerable(ctx, record, vuln)
```
ldelossa, line 49: probably should not ignore this error, 

however this ties into my other comments of making sure rhel matcher handles not having content set mapping file in memory without erroring.

rhel/repositoryscanner.go
```diff
@@ -119,16 +121,32 @@ func (r *RepositoryScanner) Scan(ctx context.Context, l *claircore.Layer) (repos
...
+	contentManifestData := contentmanifest.ContentManifest{}
+	err = json.NewDecoder(buf).Decode(&contentManifestData)
+	if err != nil {
+		return nil, err
+	}
+	return contentManifestData.ContentSets, nil
 }
 
-func (r *RepositoryScanner) containerAPI(ctx context.Context, log *zerolog.Logger, l *claircore.Layer) ([]string, error) {
+func (r *RepositoryScanner) getContentSetsUsingContainerAPI(ctx context.Context, log *zerolog.Logger, l *claircore.Layer) ([]string, error) {
```
ldelossa, line 149: cc @hdonnay when airgap gets designed, this RepoScanner will need to be configured *NOT* to call this method if airgap is enabled.

rhel/repo2cpe/updater.go
```diff
@@ -0,0 +1,164 @@
...
+type LocalUpdaterJob struct {
+	URL             string
+	Mapping         MappingFile
+	Client          *http.Client
+	lastUpdateDate  time.Time
+	lastHeaderQuery time.Time
+}
+
+// NewLocalUpdaterJob creates new LocalUpdaterJob
+func NewLocalUpdaterJob(url string, client *http.Client) *LocalUpdaterJob {
```
ldelossa, line 29: I am not seeing where this becomes "long-running" or triggered on an interval.

rhel/matcher.go
Allda: `m.mapping.RepositoryToCPE` updates a mapping if it is not loaded yet. The constructor doesn't load any data.

docs/vulnerability_matching.md
```diff
@@ -1,15 +1,15 @@
 # Vulnerability Matching
 
 This is a high level overview of the interfaces that must be implemented for end-to-end vulnerability matching.  
-A description of the end-to-end process is described as well.
```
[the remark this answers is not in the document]
Allda: My editor automatically strips trailing whitespace so that's why it is changed. When I compare the original file with spaces and my file without spaces I don't see any difference in how it is rendered. Are you sure those spaces need to be there?

https://github.com/quay/claircore/blob/master/docs/vulnerability_matching.md
https://github.com/quay/claircore/blob/f9b2dc2cf9f0650da29d72b2d28124d5951471e4/docs/vulnerability_matching.md

pkg/omnimatcher/omnimatcher.go
Allda: Fixed

rhel/repo2cpe/updater.go
Allda: `Update` function is called in `RepositoryToCPE` and it updates mapping if needed (based on Last-modified http flag)

docs/vulnerability_matching.md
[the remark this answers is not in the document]
ldelossa: For github markdown yes.

rhel/matcher.go
ldelossa: Yes, but what if this "Update" fails because the mapping file cannot be reached? The matching process should not fail. You are now going to an external source, and if thats not available, we shouldnt error the matching process, 

Is that handles gracefully?

rhel/repo2cpe/updater.go
ldelossa: @allda

When Clair is in the process of matching, its going to call more then just RHEL matcher concurrently. 

If RHEL matcher cannot reach mapping file, because maybe the server is down, the RHEL marcher should not return an error, if it does the entire match process will fail. 

Can you confirm that, if RHEL matcher needs to update the mapping file, but cannot reach the endpoint where it exists, RHEL matcher will gracefully return no results and not return an Error?

docs/vulnerability_matching.md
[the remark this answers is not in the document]
ldelossa: https://meta.stackexchange.com/questions/40976/what-is-the-reason-for-the-top-secret-two-space-newline-markdown-weirdness

It adds the <br> element in the rendered html.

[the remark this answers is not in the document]
Allda: ah ok, I will revert the change

ldelossa:
@allda needs rebase

PR #205, opened 2020-07-22
De-duplicate list of vulnerabilities

Allda:
When package is associated with multiple repositories a database query
returns duplicated vulnerability. This function only appends
vulnerability when it is no present in slice.


Allda:
@ldelossa Fixed. A map is used as a set instead.

PR #206, opened 2020-07-24
Add ability to override oval v2 using env var

Allda:
For integration testing purposes we should be able to point claircore to
different oval stream using env var.


ldelossa:
@allda needs rebase

PR #208, opened 2020-08-03
CPE mapping in indexer + rhel matching changes

Allda:
The previous implementation of CPE mapping in the matcher has a lot of corner cases. So we decided to put a mapping code into indexer instead. Mapping is fetched in the background in regular intervals and it is used during repo scan.

This pull request also contains several changes that fixed rhel matching.


Allda:
@ldelossa @hdonnay Please review this pull request. Changes are based on our chat from last week.

PR #209, opened 2020-08-04
Add package_kind to vulnerability query

Allda:
By adding package kind to database vulnerability query we query only
real affected packages. Without this code Clair produce false negatives
in case source package and binary package has same name.

When vulnerability is issued with source package all binary package that
have come from same source are marked as vulnerable.


Allda:
@ldelossa I change a query to doesn't require package kind if it is not defined in vulnerability. This way updaters that support package kinds have more precise matching results and updaters that doesn't support it yet are not affected by this change.

PR #210, opened 2020-08-05
Hank/updater config

hdonnay:
These commits allow for configuring updaters and running updaters disconnected from a database.


internal/updater/offline.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// Offline is a controller for running Updaters in an offline matter.
+type Offline struct {
+	// Output is used as the sink for calls to Run.
+	//
+	// A call to Run will write gzipped json objects to the Writer.
+	// See the 'jsonblob' package for the details of the json format.
+	//
+	// This package does not take care of any framing for the writer.
```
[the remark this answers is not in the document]
hdonnay: Yes, it just dumps gzip data to the writer. It doesn't write any header or length information or anything.

internal/updater/common.go
```diff
@@ -0,0 +1,105 @@
...
+type Controller interface {
+	// Run has the Controller execute all the Updaters passed on the channel,
+	// until it's closed. The method runs synchronously and only returns after
+	// the channel is closed or the context is canceled.
+	//
+	// Any spawned goroutines should inherit the passed-in Context.
+	//
+	// A call to Run should be thought of as one execution of the Updaters; that
+	// is, the caller should not use a single channel if it's looping, it should
+	// instead call Run on each iteration of the loop.
```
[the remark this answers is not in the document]
hdonnay: yes

internal/vulnstore/jsonblob/jsonblob.go
```diff
@@ -0,0 +1,199 @@
...
+	ch := make(chan *Entry)
+
+	go func() {
+		defer close(ch)
+		var err error
+		var de diskEntry
+		cur := uuid.Nil
+		var e *Entry
+		dec := json.NewDecoder(r)
+		for err = dec.Decode(&de); err == nil; err = dec.Decode(&de) {
```
ldelossa, line 44: just a quick comment on the logic in this loop would go a long way to quickly understanding the channel usage, packing of entry and concerns over memory here.

internal/updater/controller.go
```diff
@@ -0,0 +1,105 @@
...
+	defer e.Unlock()
+	var b strings.Builder
+	b.WriteString("updating errors:\n")
+	for n, err := range e.m {
+		fmt.Fprintf(&b, "\t%s: %v\n", n, err)
+	}
+	return errors.New(b.String())
+}
+
+// DriveUpdater drives the updater.
```
ldelossa, line 59: won't having this uppercase confuse godoc ?

internal/updater/controller.go
```diff
@@ -0,0 +1,105 @@
...
+type Controller interface {
+	// Run has the Controller execute all the Updaters passed on the channel,
+	// until it's closed. The method runs synchronously and only returns after
+	// the channel is closed or the context is canceled.
+	//
+	// Any spawned goroutines should inherit the passed-in Context.
+	//
+	// A call to Run should be thought of as one execution of the Updaters; that
+	// is, the caller should not use a single channel if it's looping, it should
+	// instead call Run on each iteration of the loop.
```
ldelossa, line 26: "it should instead call Run with a new channel on each iteration of the loop" Is that the point you are trying to convey?

internal/updater/online.go
```diff
@@ -18,11 +15,11 @@ import (
 	distlock "github.com/quay/claircore/pkg/distlock/postgres"
 )
 
-// Executor is a helper for managing locks and state for updaters.
+// Online is a helper for managing locks and state for updaters.
```
ldelossa, line 18: Can we make this "Online is a controller" like the Offline controller comments say?

libvuln/opts.go
```diff
@@ -87,6 +87,8 @@ type Opts struct {
 	// UpdaterConfigs is a map of functions for configuration of Updaters.
 	UpdaterConfigs map[string]driver.ConfigUnmarshaler
 
+	UpdaterFilter func(name string) (keep bool)
```
ldelossa, line 90: should this be private to /libvuln ? external clients will never set this.

libvuln/libvuln.go
```diff
@@ -57,13 +59,17 @@ func New(ctx context.Context, opts *Opts) (*Libvuln, error) {
...
+	if err != nil {
+		return nil, err
+	}
 
 	// Run updaters synchronously, initially.
-	if err := l.RunUpdaters(ctx, opts.UpdateWorkers, setFuncs...); err != nil {
+	if err := l.RunUpdaters(ctx, setFuncs...); err != nil {
 		return nil, err
 	}
 	if !opts.DisableBackgoundUpdates {
```
ldelossa, line 71: typo: "DisableBackgroundUpdates"

ldelossa, line 71: Previously, you would disable background updaters by specifying an empty list in the Udpater's config. Is an bool desired over that concept? If so have we made sure that the previous logic does not interfere with this logic? For example if this bool is set but the list is not-empty does the correct thing happen?

libvuln/updates.go
```diff
@@ -0,0 +1,79 @@
...
+	}, nil
+}
+
+type UpdateDriver struct {
+	exe    updater.Controller
+	client *http.Client
+	config map[string]driver.ConfigUnmarshaler
+}
+
+func (d *UpdateDriver) RunUpdaters(ctx context.Context, fs ...driver.UpdaterSetFactory) error {
```
ldelossa, line 43: libvuln.loopUpdaters and this method feel like they belong closer together. Consider the design of this separation and if its cohesive to have the summation of the looping functionality split between objects.

cmd/cctool/updaters.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	u, err := libvuln.NewOfflineUpdater(nil, nil, out)
+	if err != nil {
+		return err
+	}
+
+	ufs := []driver.UpdaterSetFactory{
+		&ubuntu.Factory{Releases: ubuntu.Releases},
+		driver.UpdaterSetFactoryFunc(alpine.UpdaterSet),
```
ldelossa, line 80: should we punt this wrapping to the actual packages? 

```
package alpine

NewFactory() ... {
  return driver.UpdaterSetFactoryFunc(alpine.UpdaterSet)
}
```

PR #212, opened 2020-08-11
Implement oval parsing for debian and ubuntu

ldelossa:
This PR implements correct OVAL parsing for debian and ubuntu moving
away from incorrect regex parsing of the oval data.

goval-parser is bumped to obtain the necessary datastructures.

### DIFF
--- a/pkg/ovalutil/dpkg.go
+++ b/pkg/ovalutil/dpkg.go
@@ -1,0 +1,136 @@
+package ovalutil
+
+import (
+	"context"
+
+	"github.com/quay/goval-parser/oval"
+	"github.com/rs/zerolog"
+
+	"github.com/quay/claircore"
+)
+
+// DpkgDefsToVulns iterates over the definitions in an oval root and assumes DpkgInfo objects and states.
+//
+// Each Criterion encountered with an EVR string will be translated into a claircore.Vulnerability
+func DpkgDefsToVulns(ctx context.Context, root *oval.Root, protoVulns ProtoVulnsFunc) ([]*claircore.Vulnerability, error) {
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "ovalutil/DpkgDefsToVulns").
+		Logger()
+	ctx = log.WithContext(ctx)
+	vulns := make([]*claircore.Vulnerability, 0, 10000)
+	pkgcache := map[string]*claircore.Package{}
+	cris := []*oval.Criterion{}
+	for _, def := range root.Definitions.Definitions {
+		// create our prototype vulnerability
+		protoVulns, err := protoVulns(def)
+		if err != nil {
+			log.Debug().
+				Err(err).
+				Str("def_id", def.ID).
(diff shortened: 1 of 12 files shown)


ubuntu/parser.go
```diff
@@ -0,0 +1,46 @@
...
+func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vulnerability, error) {
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "debian/Updater.Parse").
+		Logger()
+	ctx = log.WithContext(ctx)
+	log.Info().Msg("starting parse")
+	defer r.Close()
+	root := oval.Root{}
+	if err := xml.NewDecoder(r).Decode(&root); err != nil {
+		return nil, fmt.Errorf("debian: unable to decode OVAL document: %w", err)
```
hdonnay, line 24:
```suggestion
		Str("component", "ubuntu/Updater.Parse").
		Logger()
	ctx = log.WithContext(ctx)
	log.Info().Msg("starting parse")
	defer r.Close()
	root := oval.Root{}
	if err := xml.NewDecoder(r).Decode(&root); err != nil {
		return nil, fmt.Errorf("ubuntu: unable to decode OVAL document: %w", err)
```

pkg/ovalutil/dpkg.go
```diff
@@ -0,0 +1,135 @@
...
+				// that object, so just skip tests with a single object reference
+				// and no associated state object.
+				continue
+			}
+			if len(test.ObjectRefs) != len(test.StateRefs) {
+				log.Debug().Str("test_ref", criterion.TestRef).Msg("object refs and state refs are not in pairs. moving to next criterion")
+				continue
+			}
+			// look at each object,state pair the test references
+			// and create a vuln if an evr tag if found
```
hdonnay, line 56:
```suggestion
			// and create a vuln if an evr tag is found
```

This typo probably also exists on the RPM version :(

pkg/ovalutil/dpkg.go
```diff
@@ -0,0 +1,135 @@
...
+					vuln := *protoVuln
+					vuln.FixedInVersion = state.EVR.Body
+					name := object.Name
+
+					// if the dpkginfo_object>name field has a var_ref it indicates
+					// a variable lookup for all packages affected by this vuln is necessary.
+					//
+					// if the name.Ref field is empty it indicates a single package is affected
+					// by the vuln and that package's name is in name.Body.
+					if len(name.Ref) > 0 {
```
hdonnay, line 88: This would be easier to read if written into early return (continue, in this case) form instead of if-else

hdonnay (dismissed):
Some typos in log/error strings, but code looks good.

pkg/ovalutil/dpkg.go
ldelossa: fixed in both.

hdonnay (dismissed):
LGTM

PR #214, opened 2020-08-26
airgap enablement: round 2

hdonnay:


hdonnay:
>Does default need to be it's own package ? Could updater package just export this method?
Having the registry import the default set of updaters means those must be imported. Having it separated out means a package can use the registry without directly depending on any of the updaters.

>Do you feel strongly about implementing the registry as several package level functions working on a shared object? It would follow our pattern more to make a Registry struct with methods reflecting the package level functions, and optionally returning a "DefaultRegistry" instead of having to call "Set" before hand.
Yeah, I think having multiple registries defeats the point of using a registry. If you need to pass something in then just require it be configured beforehand.

>Having a registry also allows a nice place to take the config's "UpdaterSet" name and return the factory, should this be implemented on the registry as well?
Isn't this just a map access?

cmd/libvulnhttp/main.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/rs/zerolog/log"
 
 	"github.com/quay/claircore/libvuln"
+	"github.com/quay/claircore/updater/defaults"
```
ldelossa, line 16: underscore desired before path string ?

ldelossa:
@hdonnay my only nit is considering a name for "defaults" package which makes it immediately clear it's for the registry in updater package.

maybe "updater/registrydefaults"

cmd/libvulnhttp/main.go
hdonnay: I have the server explicitly check for errors down below, so it needs the name.

PR #218, opened 2020-09-04
updater: do not float error

ldelossa:
this commit stops an error returned by an updater from cascading to the
caller.


ldelossa:
Made obsolete by #219

PR #223, opened 2020-09-11
postgres: manage the number of update_operations

hdonnay:
Previously, this table could grow without bound, which is suboptimal.


ldelossa (approved):

ldelossa:
@hdonnay feel free to admin merge this, its block on some tests that aren't in the stale github workflow

PR #226, opened 2020-09-14
RHEL: Use last-modified to cache data

Allda:
Sometime E-tag doesn't work as expected so adding last-modified makes
sure we don't update same file multiple times if it hasn't changed.


ldelossa (approved):
LGTM

PR #228, opened 2020-09-17
rpm: Reduce database file to Packages + fix race condition

Allda:
Packages database is only file that is necessary for getting list of
installed rpms. Other files makes query more efficient.

This commit was created based on couple of images where some rpm
database file were missing and rpm scanner skipped the database.

### DIFF
--- a/rpm/packagescanner.go
+++ b/rpm/packagescanner.go
@@ -16,7 +16,6 @@ import (
 	"strings"
 
 	"github.com/rs/zerolog"
-	"golang.org/x/sync/errgroup"
 
 	"github.com/quay/claircore"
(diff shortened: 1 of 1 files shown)


Allda:
The second part of this PR fixes a race condition that was caused by cmd.Run()

ldelossa (approved):
This looks fine to me. I do think, since we only need to search for a single berkley db, the entire confidence score bits can simply be removed.

Refactor at a later time?

hdonnay:
Yeah, only looking for a single file called `Packages` means we should probably be less fail-happy in the rpm invocation section, which is some more involved code-touching.

PR #229, opened 2020-09-18
Makefile: handle SELinux permissions for volume in docker-compose

jan-zmeskal:
When setting up local dev environment on SELinux-enabled system, `libindexhttp` and `libvulnhttp` containers exit with error code and "Permission denied" in log. This is because by default a repository cloned into your home does not have the SELinux type required by docker. This can be solved by adding `z` suffix to the volume mount. Docker then takes care about applying the proper SELinux type to the mounted volume. More info [here](https://prefetch.net/blog/2017/09/30/using-docker-volumes-on-selinux-enabled-servers/).

It seems that this does not negatively affect non-SELinux systems. Adam tested it on Arch Linux and no problems were encountered. However, if someone else can also verify on non-SELinux OS, it would be great.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -38,7 +38,7 @@ services:
       LAYER_SCAN_CONCURRENCY: 10
       LOG_LEVEL: "debug"
     volumes:
-      - "./:/src/claircore/"
+      - "./:/src/claircore/:z"
     command:
       [ "bash", "-c", "cd /src/claircore/cmd/libindexhttp; go run -mod vendor ." ]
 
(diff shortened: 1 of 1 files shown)


ldelossa:
Very cool, ill check on MacOS and report back.

ldelossa:
@jan-zmeskal this looks good, works on macOS. 

If you have time would you mind checking out the local dev environment in ClairV4? https://github.com/quay/clair/tree/development-4.0

I had recently needed to disable SELinux to allow the traefik loadbalancer container to talk to /var/run/docker.socket mounted from the host. If we could handle that better, I'd love to. Right now there's just an entry in our pending documentation that says "you may need to disable SELinux"

ldelossa (approved):
Thanks.

jan-zmeskal:
@ldelossa Sure, I will definitely take a look at ClairV4. Sooner or later I'll be setting that dev env as well, so I'll keep this in mind,

ldelossa:
Thank you, appreciated.

PR #234, opened 2020-09-25
Remove cache from oval rpm parser

Allda:
The cache structure didn't work as expected in some cases. Cache key was
made of package name and module, but Package structure has much more
fields that may affect cache hit/miss. By removing this cache we fix the
bug that incorrectly filled Package with other properties.

This bug was discovered because architecture of some packages was
incorrectly overridden by package from different vulnerability.

### DIFF
--- a/pkg/ovalutil/rpm.go
+++ b/pkg/ovalutil/rpm.go
@@ -32,7 +32,6 @@ func RPMDefsToVulns(ctx context.Context, root *oval.Root, protoVulns ProtoVulnsF
 		Logger()
 	ctx = log.WithContext(ctx)
 	vulns := make([]*claircore.Vulnerability, 0, 10000)
-	pkgcache := map[string]*claircore.Package{}
 	cris := []*oval.Criterion{}
 	for _, def := range root.Definitions.Definitions {
 		// create our prototype vulnerability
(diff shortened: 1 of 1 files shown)


ldelossa:
@Allda

Why remove this over making a valid key for the package cache.

I'm generally not opposed to removing it, as we haven't measured that its particularly necessary yet. However, allocating less is typically a plus, so I must ask is it better to just make a unique key that works, or remove the cache al together?

Allda:
> @Allda
> 
> Why remove this over making a valid key for the package cache.
> 
> I'm generally not opposed to removing it, as we haven't measured that its particularly necessary yet. However, allocating less is typically a plus, so I must ask is it better to just make a unique key that works, or remove the cache al together?

I decided to remove it because a key to cache would need to be made of all Package properties and if a new property will be added it will be really easy to forget to update a key. Do you have any idea how to generate a key based on struct content?

ldelossa:
@Allda we hash a vuln like this: https://github.com/quay/claircore/blob/6295f370f5aaca200966187585424da8db4983cf/internal/vulnstore/postgres/updatevulnerabilities.go#L148

A similar function can be created for a package.

Allda:
@ldelossa Would it make even sense? To be able create a hash from Package structure we would have to create a Package object with all properties anyway which makes me think how much space we would save by using cache.

ldelossa:
that's fair. 

I'm okay with moving this for now. We haven't measured to determine if it's actually necessary and the package cache was devised optimistically.

ldelossa (approved):

PR #235, opened 2020-09-25
updaters: consolidate into manager

ldelossa:
this commit consolidates updater filtering, getting defaults from the
registry, configuring factories, configuring updaters, and running
updaters in a loop into a single cohesive unit.

### DIFF
--- a/libvuln/jsonblob/jsonblob.go
+++ b/libvuln/jsonblob/jsonblob.go
@@ -1,0 +1,224 @@
+package jsonblob
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"io"
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+// New constructs an empty Store.
+func New() (*Store, error) {
+	s := Store{}
+	s.ops = make(map[string][]driver.UpdateOperation)
+	s.entry = make(map[uuid.UUID]*Entry)
+	return &s, nil
+}
+
+// A Store buffers update operations.
+type Store struct {
+	sync.RWMutex
+	entry  map[uuid.UUID]*Entry
+	ops    map[string][]driver.UpdateOperation
+	latest uuid.UUID
+}
+
+// Load reads in all the records serialized in the provided Reader.
+func Load(ctx context.Context, r io.Reader) (*Loader, error) {
+	l := Loader{
+		dec: json.NewDecoder(r),
+		cur: uuid.Nil,
+	}
+	return &l, nil
+}
+
+// Loader is an iterator struct that returns Entries.
+//
+// Users should call Next until it reports false, then check for errors via Err.
+type Loader struct {
+	err error
+	e   *Entry
+
+	dec  *json.Decoder
+	next *Entry
+	de   diskEntry
+	cur  uuid.UUID
+}
+
+// Next reports whether there's a Entry to be processed.
+func (l *Loader) Next() bool {
+	if l.err != nil {
+		return false
+	}
+
+	for l.err = l.dec.Decode(&l.de); l.err == nil; l.err = l.dec.Decode(&l.de) {
+		id := l.de.Ref
+		// If we just hit a new Entry, promote the current one.
+		if id != l.cur {
+			l.e = l.next
+			l.next = &Entry{}
+			l.next.Updater = l.de.Updater
+			l.next.Fingerprint = l.de.Fingerprint
+			l.next.Date = l.de.Date
+		}
+		l.next.Vuln = append(l.next.Vuln, l.de.Vuln)
+		l.de.Vuln = nil // Needed to ensure the Decoder allocates new backing memory.
+
+		// If this was an initial diskEntry, promote the ref.
+		if id != l.cur {
+			l.cur = id
+			// If we have an Entry ready, report that.
+			if l.e != nil {
+				return true
+			}
+		}
+	}
+	l.e = l.next
+	return true
+}
+
+// Entry returns the latest loaded Entry.
+func (l *Loader) Entry() *Entry {
+	return l.e
+}
+
+// Err is the latest encountered error.
+func (l *Loader) Err() error {
+	// Don't report EOF as an error.
+	if errors.Is(l.err, io.EOF) {
+		return nil
+	}
+	return l.err
+}
+
+// Store writes out the Store to the provided Writer. It's the inverse of Load.
+func (s *Store) Store(w io.Writer) error {
+	s.RLock()
+	defer s.RUnlock()
+	enc := json.NewEncoder(w)
+	enc.SetEscapeHTML(false)
+	for id, e := range s.entry {
+		for _, v := range e.Vuln {
+			if err := enc.Encode(&diskEntry{
+				CommonEntry: e.CommonEntry,
+				Ref:         id,
+				Vuln:        v,
+			}); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Entry is a record of all information needed to record a vulnerability at a
+// later date.
+type Entry struct {
+	CommonEntry
+	Vuln []*claircore.Vulnerability
+}
+
+// CommonEntry is an embedded type that's shared between the "normal" Entry type
+// and the on-disk json produced by a Store's Load method.
+type CommonEntry struct {
+	Updater     string
+	Fingerprint driver.Fingerprint
+	Date        time.Time
+}
+
+// DiskEntry is a single vulnerability. It's made from unpacking an Entry's
+// slice and adding a uuid for grouping back into an Entry upon read.
+type diskEntry struct {
+	CommonEntry
+	Ref  uuid.UUID
+	Vuln *claircore.Vulnerability
+}
+
+// Entries returns a map containing all the Entries stored by calls to
+// UpdateVulnerabilities.
+//
+// It is unsafe for modification because it does not return a copy of the map.
+func (s *Store) Entries() map[uuid.UUID]*Entry {
+	s.RLock()
+	defer s.RUnlock()
+	return s.entry
+}
+
+// UpdateVulnerabilities records all provided vulnerabilities.
+func (s *Store) UpdateVulnerabilities(_ context.Context, updater string, fingerprint driver.Fingerprint, vulns []*claircore.Vulnerability) (uuid.UUID, error) {
+	now := time.Now()
+	e := Entry{
+		Vuln: vulns,
+	}
+	e.Date = now
+	e.Updater = updater
+	e.Fingerprint = fingerprint
+	ref := uuid.New() // God help you if this wasn't unique.
+	s.Lock()
+	defer s.Unlock()
+	s.latest = ref
+	s.entry[ref] = &e
+	s.ops[updater] = append([]driver.UpdateOperation{{
+		Ref:         ref,
+		Date:        now,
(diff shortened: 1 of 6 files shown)


pkg/distlock/locker.go
```diff
@@ -19,3 +19,26 @@ type Locker interface {
 	// Unlock should return the distributed lock to the implemented synchronization point.
 	Unlock() error
 }
+
+type LockerMock struct{}
```
ldelossa, line 23: This will be used in Clair, when Clairctl utilizes a jsonblob store.

libvuln/jsonblob/jsonblob.go
```diff
@@ -0,0 +1,224 @@
+package jsonblob
```
ldelossa, line 1: this is moved to support jsonblob store construction in clairctl.

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+		updaters = append(updaters, set.Updaters()...)
+	}
+
+	// reconfigure updaters
+	for _, u := range updaters {
+		f, fOK := u.(driver.Configurable)
+		cfg, cfgOK := m.configs[u.Name()]
+		if fOK && cfgOK {
+			if err := f.Configure(ctx, cfg, nil); err != nil {
+				log.Warn().Err(err).Str("updater", u.Name()).Msg("Failed configuring updaters. Updater will be excluded from batch.")
```
hdonnay, line 152:
```suggestion
				log.Warn().Err(err).Str("updater", u.Name()).Msg("configuration failed, excluding from current run")
```
I just like the log messages terse, which I feel like we can get away with because of all the context added by the logging library.

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+
+	updaters := []driver.Updater{}
+	// constructing updater sets may require network access,
+	// depending on the factory.
+	// if construction fails we will simply ignore those updater
+	// sets.
+	for _, factory := range m.factories {
+		set, err := factory.UpdaterSet(ctx)
+		if err != nil {
+			log.Error().Err(err).Msg("Error while calling factory for updater set. Updaters in this factory will be excluded from batch.")
```
hdonnay, line 140:
```suggestion
			log.Error().Err(err).Msg("excluding facotry")
```

The log message already indicates there's an error.

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+		cfg, cfgOK := m.configs[u.Name()]
+		if fOK && cfgOK {
+			if err := f.Configure(ctx, cfg, nil); err != nil {
+				log.Warn().Err(err).Str("updater", u.Name()).Msg("Failed configuring updaters. Updater will be excluded from batch.")
+				continue
+			}
+		}
+	}
+
+	log.Info().Int("total", len(updaters)).Int("batchSize", m.batchSize).Msg("Batching updaters...")
```
hdonnay, line 158:
```suggestion
	log.Info().Int("total", len(updaters)).Int("batchSize", m.batchSize).Msg("running updaters")
```
The ellipsis adds nothing, and the code isn't about to batch the updaters, it's about to run the updaters in batches.

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+				if !ok {
+					log.Debug().Str("updater", u.Name()).Msg("Lock acquisition failed. Another process is running this updater.")
+					return
+				}
+				defer lock.Unlock()
+			}
+
+			err = m.driveUpdater(ctx, u)
+			if err != nil {
+				errChan <- fmt.Errorf("%v: %v\n", u.Name(), err)
```
hdonnay, line 188:
```suggestion
				errChan <- fmt.Errorf("%v: %w", u.Name(), err)
```

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+
+			if m.pool != nil {
+				lock := postgres.NewPool(m.pool, 0)
+				ok, err := lock.TryLock(ctx, u.Name())
+				if err != nil {
+					errChan <- err
+					return
+				}
+				if !ok {
+					log.Debug().Str("updater", u.Name()).Msg("Lock acquisition failed. Another process is running this updater.")
```
hdonnay, line 180: I don't think a debug log for a common case needs to be so verbose.

```suggestion
					log.Debug().Str("updater", u.Name()).Msg("lock acquisition failed")
```

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+
+	log.Info().Int("total", len(updaters)).Int("batchSize", m.batchSize).Msg("Batching updaters...")
+
+	sem := semaphore.NewWeighted(int64(m.batchSize))
+	errChan := make(chan error, len(updaters)+1) // +1 for a potential ctx error
+	for i := range updaters {
+
+		err := sem.Acquire(ctx, 1)
+		if err != nil {
+			return fmt.Errorf("Failed to acquire semaphore")
```
hdonnay, line 166: I think this shouldn't bare-return, and instead exit the loop and do the normal error flow.

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+			if err != nil {
+				errChan <- fmt.Errorf("%v: %v\n", u.Name(), err)
+			}
+		}(updaters[i])
+	}
+
+	// acquire m.batchSize to wait on any in-flight updaters.
+	if err := sem.Acquire(ctx, int64(m.batchSize)); err != nil {
+		return fmt.Errorf("Failed to acquire final semaphore")
+	}
```
hdonnay, line 196: I don't think this accounts for cancellation. If the context is cancelled while the goroutine is waiting here, any errors won't be handled.

Alternatively, the error tracking can be much simpler and just be "an error occurred" and rely on the inner `driveUpdater` function to log errors, and any other handling is disallowed.

libvuln/updates/options.go
```diff
@@ -0,0 +1,75 @@
...
+		err := updater.Configure(context.TODO(), m.factories, cfgs, m.client)
+		if err != nil {
+			panic("updater could not be configured")
+		}
+	}
+}
+
+// WithOutOfTree allows callers to provide their own out-of-tree
+// updaters.
+func WithOutOfTree(outOfTree []driver.Updater) ManagerOption {
```
hdonnay, line 64: This should exist as an argument to NewManager xor this option.

pkg/distlock/locker.go
hdonnay: Does this need to be here? clairctl can just provide a bogus implementation.

hdonnay (changes requested):

pkg/distlock/locker.go
ldelossa: fair, I'll gut it.

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+	// max in-flight updaters.
+	batchSize int
+	// update interval used once Manager.Start is invoked, otherwise
+	// this field is not used.
+	interval time.Duration
+	// configs provided to updaters once constructed.
+	configs Configs
+	// provided to construct distributed locks.
+	// TODO(louis): this will be replaced by newest distlock implementation.
+	pool   *pgxpool.Pool
```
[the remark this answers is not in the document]
ldelossa: Right now, our distlock implementation is not shareable. A pool needs to be provided so that a distlock can be constructed for every run. Runs may happen in parallel. 

When the new distlock implementation is worked in, this will be replaced with a sharable distlock.

libvuln/updates/manager.go
ldelossa: As a debugger, that doesn't provide me with much context to why the lock failed. Most ops engineers are going to look at our debug code and not have the context we do as the authors.

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+		batchSize: DefaultBatchSize,
+		interval:  DefaultInterval,
+		client:    client,
+	}
+
+	// out-of-tree updaters is a required dependency
+	// even tho it could be empty.
+	//
+	// this avoids a dependecy issue which would force ManagerOption(s)
+	// to be ran in a specific order.
```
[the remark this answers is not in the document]
ldelossa: The original error scenario this was intended to avoid was: 

- user constructs manager placing WithConfigs option *before* OutOfTree option
- WithConfig option calls updater.Configure with default set of factories, excluding the synthentic "outOfTree" factory we create
- outOfTree factory set never gets configured. 

taking a look again tho, it appears we always use a static factory set for outOFtree, which will never be configurable. 

I guess this bears the question then, will we ever have out of tree factories which are configurable? If so, this should stay as is. While we do hard code it in our code, anyone who refactors or uses the manager in another place is susceptible to this issue if outOfTree is not a required dependency to the constructor.

libvuln/updates/options.go
```diff
@@ -0,0 +1,75 @@
...
+// Note: this option is optimal when ran after WithEnabled option. However,
+// this option has no strict depedency on others.
+func WithConfigs(cfgs Configs) ManagerOption {
+	return func(m *Manager) {
+		m.configs = cfgs
+		err := updater.Configure(context.TODO(), m.factories, cfgs, m.client)
+		if err != nil {
+			panic("updater could not be configured")
+		}
+	}
```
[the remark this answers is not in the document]
ldelossa: I'm cool with that, and I think that allows for outOfTree to remain an option, and not a dependency.

libvuln/updates/manager.go
ldelossa: So you are suggesting ripping out the error channel? 

IIRC this was your suggestion, so that error messages for the updaters are not interleaved with other messages, and a final log message with all errors that occured during the process is provided, which I do like. 

Getting rid of the error channel does sound nice too, as it simplifies things.

libvuln/updates/manager.go
```diff
@@ -0,0 +1,256 @@
...
+	configs Configs
+	// provided to construct distributed locks.
+	// TODO(louis): this will be replaced by newest distlock implementation.
+	pool   *pgxpool.Pool
+	client *http.Client
+	store  vulnstore.Updater
+}
+
+// NewManager will return a manager ready to have its Start or Run methods called.
+func NewManager(ctx context.Context, store vulnstore.Updater, pool *pgxpool.Pool, outOfTree []driver.Updater, client *http.Client, opts ...ManagerOption) (*Manager, error) {
```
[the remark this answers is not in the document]
ldelossa: yup, lets decide which one makes sense by answering my other comments, and nix the extraneous one.

libvuln/updates/manager.go
ldelossa: updated to 
```
log.Debug().Str("updater", u.Name()).Msg("another process running updater, excluding from run.")
```

libvuln/updates/manager.go
ldelossa: Updated to just conditionally wait for the acquire to succeed. This should be guaranteed, as when the ctx is canceled, all in-flight go routines are guaranteed to release their sems.

libvuln/updates/manager.go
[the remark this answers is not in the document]
ldelossa: alright @hdonnay I removed the required outOfTree in favor of the options. 
If we decide that the synthetic outOfTree factory needs configuration at some point, it should become a dependency again to avoid the configuration ordering issue.

hdonnay (approved):
LGTM

PR #241, opened 2020-09-30
Affected manifest: Update query and logic that determines afected manifests

Allda:
The pull request fixes 2 bugs in omnimatcher logic.

Add missing properties into affected manifest query
These properties were missing in queries that search for affected
manifest which caused incorrect API response.

-----
Use mather's Filter() in omnimatcher. 
The previous implementation used all matchers to find out whether
package is vulnerable. This approach didn't work because each matcher is
interested only in specific subset packages given by mather's filter.

This commit adds filer to omnimatcher to make sure only the right matchers
are used.


ldelossa (dismissed):
LGTM.

ldelossa:
@Allda needs a rebase, @ me or hank when done.

ldelossa (approved):

PR #245, opened 2020-10-07
repo2cpe: add errorchecking

hdonnay:


rhel/repositoryscanner.go
```diff
@@ -86,7 +86,7 @@ func NewRepositoryScanner(ctx context.Context, c *http.Client, cs2cpeURL string)
 				}
 			}
 			u := repo2cpe.NewLocalUpdaterJob(cs2cpeURL, nil)
-			u.Start(context.TODO())
+			u.Start(ctx)
```
ldelossa, line 89: Is this going to work? 

The ctx being provided to `NewRepositoryScanner` will be http request scoped. Once the request is finished the local updater background job is going to be cancelled, which is not desired.

ldelossa (changes requested):

ldelossa (approved):

ldelossa (approved):

PR #248, opened 2020-10-19
updaters: do not kill loop on error

ldelossa:
this removes the case where a single updater error kills the long
running update loop


Allda (approved):
LGTM

PR #257, opened 2020-10-30
etc: update podman yaml

hdonnay:
Previous versions of the postgres container were happy to have the sever
run as root, but the new entrypoint drops permissions. The generated
yaml enshrined not using the container's entrypoint. This commit undoes
that.


ldelossa (dismissed):
LGTM

ldelossa (approved):

Issue #254, opened 2020-10-29
False positive for jq (CVE-2016-4074)


SofyaTavrovskaya:
### Description of Problem / Feature Request

Clair v4.0.0-rc.19 detects already fixed vulnerability CVE-2016-4074 in jq 1.6-r0:
`ucp-azure-ip-allocator:3.3.0 found jq 1.6-r0 CVE-2016-4074 (fixed: 1.6_rc1-r0)`

The reason seems to be that Clair wrongly interprets the version and as the fixed version is a release candidate and the installed version is already the newest version that contains the fix.

Maybe it more reasonable to use in https://github.com/quay/claircore/blob/04cb53cd0090dc60cf240ac4dd46db297ee8b18f/alpine/matcher.go#L6 github.com/knqyf263/go-apk-version module, which can resolve and work with alpine rc versions. 

### Expected Outcome
`ucp-azure-ip-allocator:3.3.0 ok`

### Actual Outcome

`ucp-azure-ip-allocator:3.3.0 found jq 1.6-r0 CVE-2016-4074 (fixed: 1.6_rc1-r0)`

### Environment

- Clair version/image: v4.0.0-rc.19
- Clair client name/version: clairctl version 1.2.8
- Host OS: "Ubuntu 20.04.1 LTS"
- Kernel (e.g. `uname -a`): Linux 5.4.0-52-generic
- Kubernetes version (use `kubectl version`): NA
- Network/Firewall setup: NA

ldelossa:
@SofyaTavrovskaya I've transferred this issue to ClairCore as the issue exists in this repository. 

I will build a clair version which utilizes the apk version parser you are specifying here and ping you with a container that you can test with.

ldelossa:
@SofyaTavrovskaya I have built a new container which utilizes the apk version package you specified. Can you please test this with

```
quay.io/projectquay/clair:apk-parser
```

SofyaTavrovskaya:
@ldelossa Thank you.
I will test it, but I can't find tag apk-parser in registry [](https://quay.io/repository/projectquay/clair)

ldelossa:
@SofyaTavrovskaya whoops, looks like my original push failed. It is now up on quay.io

SofyaTavrovskaya:
@ldelossa I've tested on alpine image, and it works.
`sofya@sofya-HP-ProBook-440-G6:~/clair/cmd/clairctl$ ./clairctl report ucp-azure-ip-allocator:3.3.0
[...]

PR #258, opened 2020-11-02
matcher: add apk specific version parser for alpine

ldelossa:
this commit adds an apk specific version parser for alpine matcher.

Closes: #254


hdonnay (approved):
Approved, but I think this needs to actually be rebased to not mess up the commit CI stuff.

hdonnay (approved):

PR #263, opened 2020-11-17
ubuntu: attempt to add normalized severity

hdonnay:


ldelossa (changes requested):
Looks like this is a best effort approach, from the title of your commit? just curious.

also this will need a quick doc's update here: https://github.com/quay/claircore/blob/master/docs/concepts/severity_mapping.md

only blocking on doc's update.

ldelossa (approved):

PR #264, opened 2020-11-18
Louis/indexer hash dedupe with migration

ldelossa:
# What
This pull adds commits necessary to reduce the copying of layer and manifest hash strings across multiple tables. 

# Why
Reduce database table size, especially in regards to the "manifest_index" table.


hdonnay (changes requested):
The changes in `test/postgres` seem wrong, they just change the column names but don't actually change the queries from using the digests.

ldelossa:
@hdonnay 

files in `test/postgres` fixed.

libindex/migrations/migration2.go
```diff
@@ -0,0 +1,105 @@
...
+	ALTER TABLE dist_scanartifact ADD FOREIGN KEY (layer_id) REFERENCES layer(id);
+	UPDATE dist_scanartifact AS ds SET layer_id = (SELECT id FROM layer WHERE hash = ds.layer_hash);
+
+	ALTER TABLE indexreport ADD COLUMN manifest_id bigint;
+	ALTER TABLE indexreport ADD FOREIGN KEY (manifest_id) REFERENCES manifest(id);
+	UPDATE indexreport AS ir SET manifest_id = (SELECT id FROM manifest WHERE hash = ir.manifest_hash);
+
+	ALTER TABLE manifest_index ADD COLUMN manifest_id bigint;
+	ALTER TABLE manifest_index ADD FOREIGN KEY (manifest_id) REFERENCES manifest(id);
+
```
Allda, line 52: There is a missing UPDATE command for the manifest_index table

ldelossa: good catch.

Allda (approved):
Based on the migration I did yesterday + tests of the basic use case I can confirm it works fine.

Allda (approved):

hdonnay (approved):

PR #265, opened 2020-11-19
updater: remove updater diff limit

Allda:
The limit can be removed because it is no longer needed and not having a
limit didn't cause issues as we thought at a beginning.


ldelossa (approved):
:+1:

PR #266, opened 2020-11-19
Desqlx

hdonnay:
This removes lingering usages of the sqlx package.


internal/indexer/postgres/affectedmanifest.go
```diff
@@ -120,6 +130,9 @@ func affectedManifests(ctx context.Context, pool *pgxpool.Pool, v claircore.Vuln
 		pkgsToFilter = append(pkgsToFilter, pkg)
 	}
 	log.Debug().Int("count", len(pkgsToFilter)).Msg("packages to filter")
+	if err := rows.Err(); err != nil {
```
ldelossa, line 133: is this check redundant? you exhaust the rows in the loop above, is it still necessary for a final check of rows.Err here?

internal/indexer/postgres/affectedmanifest.go
```diff
@@ -142,13 +155,13 @@ func affectedManifests(ctx context.Context, pool *pgxpool.Pool, v claircore.Vuln
...
+	// containing each unique manifest.
 	//
-	// since this loop opens a db conn each iteration the returned rows
+	// Since this loop opens a db conn each iteration the returned rows
 	// handle must be closed manually and not deferred else a buildup of db conns
-	// may occur until the method exits
+	// may occur until the method exits.
 	set := map[string]struct{}{}
 	out := []claircore.Digest{}
 	for _, record := range filteredRecords {
```
ldelossa, line 167: As a boyscout task, what do you think about wrapping the code inside the for loop as a closure? Closing each iteration's row can happen as a defer then.

internal/indexer/postgres/distributionsbylayer.go
```diff
@@ -33,47 +33,37 @@ func distributionsByLayer(ctx context.Context, db *sqlx.DB, hash claircore.Diges
...
-	for _, scnr := range scnrs {
-		var scannerID int64
-		err := db.Get(&scannerID, selectScanner, scnr.Name(), scnr.Version(), scnr.Kind())
+	scannerIDs := make([]int64, len(scnrs))
+	for i, scnr := range scnrs {
+		err := s.pool.QueryRow(ctx, selectScanner, scnr.Name(), scnr.Version(), scnr.Kind()).
+			Scan(&scannerIDs[i])
 		if err != nil {
-			return nil, fmt.Errorf("store:distributionseByLayer failed to retrieve scanner ids for scnr %v: %v", scnr, err)
+			return nil, fmt.Errorf("store:distributionseByLayer failed to retrieve scanner ids for scanner %v: %v", scnr, err)
```
ldelossa, line 52: house keeping: these error messages use a "id:" prefix which I don't think took off in the rest of the code base. Feel free to clean-up these error messages to make them uniform with the rest of the code.

internal/indexer/postgres/layerscanned.go
```diff
@@ -2,52 +2,61 @@ package postgres
...
+	name = $1 AND version = $2 AND kind = $3;
+`
 		selectScanned = `
-		SELECT layer.hash
-		FROM layer
-                 JOIN scanned_layer ON scanned_layer.layer_id = layer.id
-		WHERE layer.hash = $1
-		  AND scanned_layer.scanner_id = $2;
-		`
+SELECT
```
ldelossa, line 27: dope.

internal/indexer/postgres/packagesbylayer.go
```diff
@@ -102,7 +96,7 @@ func packagesByLayer(ctx context.Context, db *sqlx.DB, hash claircore.Digest, sc
 			&pkg.Module,
 			&pkg.Arch,
 
-			&spkg.ID,
+			&srcID,
```
ldelossa, line 99: was this a bug previously?

internal/indexer/postgres/setindexfinished.go
```diff
@@ -4,83 +4,74 @@ import (
...
-		INSERT
-		INTO scanned_manifest (manifest_id, scanner_id)
-		VALUES ((SELECT manifest_id FROM manifests),
-				$2);
-		`
-		upsertIndexReport = `
-		with manifests as (
-			select id as manifest_id from manifest
-			where hash = $1
+		cte = `
```
ldelossa, line 13: I'm not a fan of building the query like this. It makes it difficult to copy and paste into a database IDE for troubleshooting. It's also a little confusing to piece together the elements.

internal/indexer/postgres/setlayerscanned.go
```diff
@@ -2,51 +2,49 @@ package postgres
...
+ON CONFLICT
+	(layer_id, scanner_id)
+DO
+	NOTHING;
 		`
-	)
 
 	// get scanner id
-	row := pool.QueryRow(ctx, selectScanner, scnr.Name(), scnr.Version(), scnr.Kind())
+	row := s.pool.QueryRow(ctx, selectScanner, scnr.Name(), scnr.Version(), scnr.Kind())
```
ldelossa, line 34: do you want to replace this block with the store.selectScanners method?

libindex/controllerfactory.go
```diff
@@ -7,21 +7,19 @@ import (
...
 // controllerFactory is the default ControllerFactory
 func controllerFactory(ctx context.Context, lib *Libindex, opts *Opts) (*controller.Controller, error) {
-	sc := dlpg.NewLock(lib.db, opts.ScanLockRetry)
 	ft := fetcher.New(lib.client, opts.LayerFetchOpt)
 
 	// convert libindex.Opts to indexer.Opts
 	sOpts := &indexer.Opts{
 		Store:         lib.store,
-		ScanLock:      sc,
+		ScanLock:      lib.locks,
```
ldelossa, line 22: Is this an issue? A single lock cannot be shared and must be created on controller construction. This is a fallout of our previous lock design, its a lock per go routine. 

I belelieve LibIndex needs to be configured with a handle to the pool, and a NewPool lock needs to be constructed by this factory, or else attempting to lock for the same manifest will cause an error.

ldelossa (changes requested):

internal/indexer/postgres/affectedmanifest.go
hdonnay: I think so. This is how errors that occur while reading rows after a successful query are reported.

internal/indexer/postgres/packagesbylayer.go
hdonnay: I think so. I think we only didn't notice because of our... weak support for source packages.

ldelossa: yeah think so lol.

internal/indexer/postgres/affectedmanifest.go
ldelossa: wouldn't you encounter any possible errors on the Scan calls here: https://github.com/quay/claircore/pull/266/files/6c513e0421fcf451f93a880e8d91aef20d562804#diff-fe8b20110d1dedeb23c0974c2532aedf03f71871bfce78e7d485fcc8751ac417R109

internal/indexer/postgres/setlayerscanned.go
hdonnay: I meant to rework this because it's doing something odd and only selecting one scanner, so I'll do that.

internal/indexer/postgres/affectedmanifest.go
hdonnay: I'm pretty sure this is needed to catch an error where the connection drops after the Scan call. The Next call just returns a bool, and the Err call retrieves a stashed error.

ldelossa: ahhh, I see.

ldelossa (approved):
LGTM :100:

PR #272, opened 2020-12-01
fetch: turn layer fetcher into a generic fetcher

hdonnay:
I'd like to add onto this PR to ultimately make it so that we're talking to quay.io for integration tests, and this is the first step.


hdonnay:
Moving our test layers to quay.io is a bigger project that can be done piecemeal, it doesn't need to be in a single PR. So, I'm just going to undraft this.

ldelossa (approved):

PR #273, opened 2020-12-02
ovalutil: attempt lookups on single criterion

hdonnay:
The code was changed in an attempt to capture definitions that only have
existence checks, but inadvertently started ignoring definitions with
a single version check.


ldelossa:
superseded by https://github.com/quay/claircore/pull/274

PR #279, opened 2020-12-08
Add exception for rhel-7-alt OVAL stream

jan-zmeskal:
See the comment above the exception for explanation of this PR.

There was also an alternate approach: Creating 
```go
var ignoredStreams = []string{
   "RHEL7-rhel-7-alt",
}
```
and then checking for each image whether it is ignored or not. Even though this would be better from maintainability point of view, I decided to go for this simple approach. It's mainly because:

- This will save a lot of iterations
- With high probability there'll be no other ignored stream

### DIFF
--- a/rhel/updaterset.go
+++ b/rhel/updaterset.go
@@ -150,6 +150,14 @@ func (f *Factory) UpdaterSet(ctx context.Context) (driver.UpdaterSet, error) {
 		case strings.Contains(p, "RHEL8"):
 			r = RHEL8
 		case strings.Contains(p, "RHEL7"):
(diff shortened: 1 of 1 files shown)


jan-zmeskal:
@Allda @ldelossa @hdonnay PTAL

ldelossa (approved):
LGTM. 

If this becomes more of a need in the future, we'll probably want to create a black list for specific repos, but we'll address that if another one of these PR's pop up.

jan-zmeskal:
> If this becomes more of a need in the future, we'll probably want to create a black list for specific repos, but we'll address that if another one of these PR's pop up.

Agreed!

PR #280, opened 2020-12-10
alpine: switch to JSON security DB

hdonnay:
The YAML format seems to get corrupted annoyingly often. The JSON
database seems to not have these problems, so use that.


ldelossa (approved):
dope.

PR #283, opened 2020-12-14
libvulnhttp: add DisableBackgroundUpdates config option

jan-zmeskal:
This option is useful when you develop `libvuln` library, want to keep it running but don't want regular updates to get in your way.

Signed-off-by: Jan Zmeskal <jzmeskal@redhat.com>

### DIFF
--- a/cmd/libvulnhttp/main.go
+++ b/cmd/libvulnhttp/main.go
@@ -19,12 +19,13 @@ import (
 // Config this struct is using the goconfig library for simple flag and env var
 // parsing. See: https://github.com/crgimenes/goconfig
 type Config struct {
-	HTTPListenAddr string `cfgDefault:"0.0.0.0:8081" cfg:"HTTP_LISTEN_ADDR"`
-	MaxConnPool    int    `cfgDefault:"100" cfg:"MAX_CONN_POOL" cfgHelper:"the maximum size of the connection pool used for database connections"`
-	ConnString     string `cfgDefault:"host=localhost port=5434 user=claircore dbname=claircore sslmode=disable" cfg:"CONNECTION_STRING" cfgHelper:"Connection string for the provided DataStore"`
-	Run            string `cfg:"RUN" cfgDefault:"." cfgHelper:"Regexp of updaters to run."`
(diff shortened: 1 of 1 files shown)


ldelossa (changes requested):
I believe this is redundant. 

If you provide a empty array to UpdaterSet configuration, it instructs libvuln to not run any updater. UpdaterSet config param is a tristate: nil - use defaults, len(0) array - run no updaters, len(>0) - run only updaters provided. 

Please correct me if I'm wrong.

jan-zmeskal:
@ldelossa That is correct. However with this PR I consider scenario where you don't have config file parsing capabilities provided by `clair` CLI command. This is used when you run `cmd/libvulnhttp/main.go` directly. 

What you say could also be applied for example to [Run option](https://github.com/quay/claircore/blob/master/cmd/libvulnhttp/main.go#L25). You could achieve the same by specifying `UpdaterSet`, but when running `libvulnhttp` directly, you don't have those capabilities.

ldelossa:
@jan-zmeskal possible to accomplish this without a new config value?

jan-zmeskal:
@ldelossa I mean I can always change the code directly for example [here](https://github.com/quay/claircore/blob/f639452bfe3872d730e11462e117f122a5fcde7a/cmd/libvulnhttp/main.go#L94) when I'm developing. It just isn't the best experience. Also I though that not wanting regular updates when running `libvulnhttp` might be a legit option. However, if we want to stay clear of introducing new config options for this binary, we can close this PR and I'll just work this around by changing the code.

One other possibility is letting the user set `UpdateInterval` as env variable and changing how `UpdateInterval == 0` is treated [here](https://github.com/quay/claircore/blob/master/libvuln/opts.go#L122). However that seems like an even bigger intervention and does not solve the problem of introducing new config option to `libvulnhttp`.

Maybe there's another way, but I don't see it.

ldelossa:
Okay I see your reasoning here. Looks good to me.

ldelossa (approved):

ldelossa (approved):

ldelossa:
@jan-zmeskal needs a rebase.

jan-zmeskal:
@ldelossa done

ldelossa (approved):

PR #287, opened 2021-01-05
indexer: filter scanners during manifest check

ldelossa:
This commit adds a filtering step in the CheckManifest state.
Now, if a manifest reports being analyzed by a particular scanner that
scanner will be omitted from all subsequent states.

This is safe to do, if a manifest was analyzed by a particular scanner,
this also means all it's layers were analyzed by the particular scanner.
Thus, the scanner need not be included in the rest of the indexing
process.

This fixes a unique constraint issue in the IndexFinished state.
Previous to this commit, an attempt to persist duplicate scanner ids was
made in the aforementioned state.

### DIFF
--- a/internal/indexer/controller/checkmanifest.go
+++ b/internal/indexer/controller/checkmanifest.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/quay/claircore/internal/indexer"
 	"github.com/rs/zerolog"
 )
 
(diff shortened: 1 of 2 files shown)


jan-zmeskal:
@ldelossa LGTM, great idea with setting `filtered`'s capacity and not length. I tested this locally and got good results.

hdonnay (approved):
very clever

LGTM

PR #289, opened 2021-01-11
vet fixes

hdonnay:
These are the things that `go vet` complained about. I think the only controversial change (that could be dropped and thought about more) is the `AffectedManifests` commit.

### DIFF
--- a/libindex/libindex.go
+++ b/libindex/libindex.go
@@ -186,7 +186,7 @@ func (l *Libindex) IndexReport(ctx context.Context, hash claircore.Digest) (*cla
 }
 
 // AffectedManifests retrieves a list of affected manifests when provided a list of vulnerabilities.
-func (l *Libindex) AffectedManifests(ctx context.Context, vulns []claircore.Vulnerability) (claircore.AffectedManifests, error) {
+func (l *Libindex) AffectedManifests(ctx context.Context, vulns []claircore.Vulnerability) (*claircore.AffectedManifests, error) {
 	const (
 		maxGroupSize = 100
 	)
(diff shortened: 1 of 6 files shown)


libindex/libindex.go
```diff
@@ -186,7 +186,7 @@ func (l *Libindex) IndexReport(ctx context.Context, hash claircore.Digest) (*cla
 }
 
 // AffectedManifests retrieves a list of affected manifests when provided a list of vulnerabilities.
-func (l *Libindex) AffectedManifests(ctx context.Context, vulns []claircore.Vulnerability) (claircore.AffectedManifests, error) {
+func (l *Libindex) AffectedManifests(ctx context.Context, vulns []claircore.Vulnerability) (*claircore.AffectedManifests, error) {
```
ldelossa, line 189: looks good, but was expecting to see some changes in the notifier code, not necessary? Auto de-ref may take care of it without changes.

ldelossa (approved):

libindex/libindex.go
hdonnay: What notifier code? This is only used in the libvuln HTTP handler, as far as I can tell.

ldelossa: Sorry, I meant I was expecting to see a linked pr: notifier in Clair utilizes this: https://github.com/quay/clair/blob/main/notifier/processor.go

ldelossa (approved):

ldelossa:
**Note: only backport bug fixes, do not backport api change in libindex/libindex.go**

PR #290, opened 2021-01-12
logging: move to zlog

hdonnay:
This PR moves over to a log package that does logging context differently.

Instead of using the Context to pass down a logger with added context, use a logging facade that understands the opentelemetry labeling system and uses that to generate context on each log event.

While this is in draft, I'm going to try to tackle each package in its own commit, for rebasing sanity.


ldelossa:
Like it :+1:

hdonnay:
I think this will need some admin mojo for the merge because it drops go1.13 support.

ldelossa:
@hdonnay yeah, in settings for this repo the Test(1.13) check can be thrown away.

hdonnay:
[I did some editing,](https://github.com/quay/claircore/pull/290/files?file-filters%5B%5D=.md&file-filters%5B%5D=.mod&file-filters%5B%5D=.sum&file-filters%5B%5D=.yml#diff-528ae1267b7d0f475e5f4d9e57ea1828e721867ecb513b927588f9756102c781) anything specific to change?

ldelossa:
Whoops, nope just missed that file.

ldelossa (approved):

PR #300, opened 2021-01-22
Fast init

hdonnay:
image

### DIFF
--- a/internal/vulnstore/postgres/initialized.go
+++ b/internal/vulnstore/postgres/initialized.go
@@ -1,0 +1,29 @@
+package postgres
+
+import (
+	"context"
+	"sync/atomic"
+)
+
+func (s *Store) Initialized(ctx context.Context) (bool, error) {
+	const query = `
+SELECT EXISTS(SELECT 1 FROM vuln LIMIT 1);
+`
+	ok := atomic.LoadUint32(&s.initialized) != 0
+	if ok {
+		return true, nil
+	}
+
+	if err := s.pool.QueryRow(ctx, query).Scan(&ok); err != nil {
+		return false, err
+	}
+	// There were no rows when we looked, so report that. Don't update the bool,
+	// because it's in the 'false' state.
+	if !ok {
+		return false, nil
+	}
+	// If this fails, it means a concurrent goroutine already swapped. Any
+	// subsequent calls will see the 'true' value.
+	atomic.CompareAndSwapUint32(&s.initialized, 0, 1)
+	return true, nil
+}
--- a/internal/vulnstore/postgres/store.go
+++ b/internal/vulnstore/postgres/store.go
@@ -19,6 +19,8 @@ import (
 // store implements all interfaces in the vulnstore package
 type Store struct {
 	pool *pgxpool.Pool
+	// Initialized is used as an atomic bool for tracking initialization.
+	initialized uint32
 }
 
 func NewVulnStore(pool *pgxpool.Pool) *Store {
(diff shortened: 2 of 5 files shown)


libvuln/libvuln.go
```diff
@@ -162,3 +169,17 @@ func (l *Libvuln) GCFull(ctx context.Context) (int64, error) {
 
 	return i, err
 }
+
+// Initialized reports whether the backing vulnerability store is initialized.
+func (l *Libvuln) Initialized(ctx context.Context) (bool, error) {
+	ok := l.initialized.Load().(bool)
```
ldelossa, line 175: instead of having l.initialized, can store.Initialized just return an atomic bool?

internal/vulnstore/postgres/store.go
```diff
@@ -18,7 +19,9 @@ import (
 
 // store implements all interfaces in the vulnstore package
 type Store struct {
-	pool *pgxpool.Pool
+	pool        *pgxpool.Pool
+	initMu      sync.RWMutex
```
ldelossa, line 23: can't this be dropped to just an atomic bool? its okay if more the one go routine writes "true" to it, nothing will flip it back to "false" and all writes of "true" are atomic to any viewers.

ldelossa (commented):
I'm curious if the whole concept here can be driven with a single atomic boolean. 

the store gets an atomic boolean called "initialized" 

Calls to store.Initialized first check atomic bool, if false, check db and flip. 

libvuln.Initialized just asks store.Initialized for a pointer to the atomic value, eliminating the need for a second one in libvuln.

hdonnay:
I don't think it can be done Correctly with an atomic value, but I think this way works.

internal/vulnstore/postgres/initialized.go
```diff
@@ -0,0 +1,29 @@
...
+		return false, err
+	}
+	// There were no rows when we looked, so report that. Don't update the bool,
+	// because it's in the 'false' state.
+	if !ok {
+		return false, nil
+	}
+	// If this fails, it means a concurrent process already swapped. Any
+	// subsequent calls will see the 'true' value.
+	atomic.CompareAndSwapUint32(&s.initialized, 0, 1)
```
ldelossa, line 27: dope

internal/vulnstore/postgres/store.go
hdonnay: That's not correct, because the calls may be concurrent. I think my latest change does it correctly, though.

internal/vulnstore/postgres/initialized.go
ldelossa: tho you may want to change that log line from "process" to "goroutine" as this value is process specific.

internal/vulnstore/postgres/store.go
```diff
@@ -18,7 +18,8 @@ import (
 
 // store implements all interfaces in the vulnstore package
 type Store struct {
-	pool *pgxpool.Pool
+	pool        *pgxpool.Pool
+	initialized uint32 // bool for tracking initialization
```
ldelossa, line 22: nit: make a comment about this being an atomic value?

ldelossa:
LGTM.

hdonnay:
Does this need changes in the test binaries? This change means `libvulnhttp` doesn't block on start anymore.

ldelossa:
I *think* most of our tests focus on specific components, and do not create a libvuln instance directly. It's possible tests in Clair will require a change. 

I do not see the `libvuln.New` used in any test files, only libvulnhttp/main.go

ldelossa (approved):

PR #307, opened 2021-02-08
vulnstore: chunked vuln cleanup

ldelossa:
this commit moves to a chunking strategy for cleaning up
vulnerabilities.

the previous methods take too long to complete on our largest two
tables.

### DIFF
--- a/cmd/libvulnhttp/main.go
+++ b/cmd/libvulnhttp/main.go
@@ -82,7 +82,7 @@ func confToLibvulnOpts(conf Config) *libvuln.Opts {
 		Migrations:               true,
 		UpdaterSets:              nil,
 		DisableBackgroundUpdates: conf.DisableBackgroundUpdates,
-		UpdateInterval:           30 * time.Second,
+		UpdateInterval:           1 * time.Minute,
 		UpdateWorkers:            10,
 		UpdateRetention:          10,
 	}
--- a/internal/vulnstore/postgres/gc.go
+++ b/internal/vulnstore/postgres/gc.go
@@ -1,0 +1,279 @@
+package postgres
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"runtime"
+	"strings"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgtype"
+	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"golang.org/x/sync/semaphore"
+)
+
+// GC is split into two phases, first it will identify any update operations
+// which are older then the provided keep value and delete these.
+//
+// Next it will perform chunked deletions of any vulns from the vuln table
+// which are not longer referenced by update operations.
+//
+// The GC is throttled to not overload the database with cascade deletes.
+// If a full GC is required run this method until the returned int64 value
+// is 0.
+func (s *Store) GC(ctx context.Context, keep int) (int64, error) {
+	const (
+		// GCThrottle sets a limit for the number of deleted update operations
+		// (and subsequent cascade deletes in the uo_vuln table) that can occur in a GC run.
+		GCThrottle = 50
+	)
+
+	// obtain update operations which need deletin'
+	ops, totalOps, err := eligibleUpdateOpts(ctx, s.pool, keep)
+	if err != nil {
+		return 0, err
+	}
+
+	// delete em', but not too many...
+	if totalOps >= GCThrottle {
+		ops = ops[:GCThrottle]
+	}
+
+	deletedOps, err := s.DeleteUpdateOperations(ctx, ops...)
+	if err != nil {
+		return totalOps - deletedOps, err
+	}
+
+	// get all updaters we know about.
+	updaters, err := distinctUpdaters(ctx, s.pool)
+	if err != nil {
+		return totalOps - deletedOps, err
+	}
+
+	// issue concurrent chunked deletion for known updaters
+	// limit concurrency by available CPUs.
+	cpus := int64(runtime.NumCPU())
+	sem := semaphore.NewWeighted(cpus)
+
+	errC := make(chan error, len(updaters))
+
+	for _, updater := range updaters {
+		err = sem.Acquire(ctx, 1)
+		if err != nil {
+			break
+		}
+		go func(u string) {
+			defer sem.Release(1)
+			err := chunkedCleanup(ctx, s.pool, u)
+			if err != nil {
+				errC <- err
+			}
+		}(updater)
+	}
+
+	// unconditionally wait for all in-flight go routines to return.
+	// the use of context.Background and lack of error checking is intentional.
+	// all in-flight go routines are guarantee to release their sems.
+	sem.Acquire(context.Background(), cpus)
+
+	close(errC)
+	if len(errC) > 0 {
+		b := strings.Builder{}
+		b.WriteString("encountered the following errors during gc: \n")
+		for e := range errC {
+			b.WriteString(e.Error() + "\n")
+		}
+		return totalOps - deletedOps, errors.New(b.String())
+	}
+	return totalOps - deletedOps, nil
+}
+
+// distinctUpdaters returns all updaters which have registered an update
+// operation.
+func distinctUpdaters(ctx context.Context, pool *pgxpool.Pool) ([]string, error) {
+	const (
+		// will always contain at least two update operations
+		selectUpdaters = `
+SELECT DISTINCT(updater) FROM update_operation;
+`
+	)
+	rows, err := pool.Query(ctx, selectUpdaters)
+	if err != nil {
+		return nil, fmt.Errorf("error selecting distinct updaters: %v", err)
+	}
+	defer rows.Close()
+
+	var updaters []string
+	for rows.Next() {
+		var updater string
+		err := rows.Scan(&updater)
+		switch err {
+		case nil:
+			// hop out
+		default:
+			return nil, fmt.Errorf("error scanning updater: %v", err)
+		}
+		updaters = append(updaters, updater)
+	}
+	if rows.Err() != nil {
+		return nil, rows.Err()
+	}
+	return updaters, nil
+}
+
+// eligibleUpdateOpts returns a list of update operation refs which exceed the specified
+// keep value.
+func eligibleUpdateOpts(ctx context.Context, pool *pgxpool.Pool, keep int) ([]uuid.UUID, int64, error) {
+	const (
+		// this query will return rows of UUID arrays.
+		// each returned array are the UUIDs which exceed the provided keep value
+		updateOps = `
+WITH ordered_ops AS (
+    SELECT array_agg(ref ORDER BY date DESC) AS refs FROM update_operation GROUP BY updater
+)
+SELECT ordered_ops.refs[$1:]
+FROM ordered_ops
+WHERE array_length(ordered_ops.refs, 1) > $2;
+`
+	)
+
+	// gather any update operations exceeding our keep value.
+	// keep+1 is used because PG's array slicing is inclusive,
+	// we want to grab all items once after our keep value.
+	m := []uuid.UUID{}
+	rows, err := pool.Query(ctx, updateOps, keep+1, keep)
+	switch err {
+	case nil:
+	default:
+		return nil, 0, fmt.Errorf("error querying for update operations: %v", err)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		// pgx will not scan directly into a []uuid.UUID
+		tmp := pgtype.UUIDArray{}
+		err := rows.Scan(&tmp)
+		if err != nil {
+			return nil, 0, fmt.Errorf("error scanning update operations: %w", err)
+		}
+		for _, u := range tmp.Elements {
+			m = append(m, u.Bytes) // this works since [16]byte value is assignable to uuid.UUID
+		}
+	}
+	if rows.Err() != nil {
+		return nil, 0, rows.Err()
+	}
+	return m, int64(len(m)), nil
+}
+
+func chunkedCleanup(ctx context.Context, pool *pgxpool.Pool, updater string) error {
+	const (
+		paginatedSelect = `
+SELECT id FROM vuln WHERE vuln.updater = $1 AND id > $2 ORDER BY id ASC LIMIT 10000;
+`
+		shouldDelete = `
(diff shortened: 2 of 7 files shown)


cmd/libvulnhttp/main.go
```diff
@@ -82,7 +82,7 @@ func confToLibvulnOpts(conf Config) *libvuln.Opts {
 		Migrations:               true,
 		UpdaterSets:              nil,
 		DisableBackgroundUpdates: conf.DisableBackgroundUpdates,
-		UpdateInterval:           30 * time.Second,
+		UpdateInterval:           1 * time.Minute,
```
ldelossa, line 85: libvuln's opt parses defaults anything under 1 minute to the default, 30 minutes. changed this for testing purposes.

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"golang.org/x/sync/semaphore"
+)
+
+func (s *Store) GC(ctx context.Context, keep int) (int64, error) {
+	const (
+		// GCThrottle sets a limit for the number of deleted update operations
+		// (and subsequent cascade deletes in the uo_vuln table) that can occur in a GC run.
+		GCThrottle = 50
```
ldelossa, line 19: increased this, I'm a little fearful of a lot of cascade deletes happening at once, they will cause individual row locks on the table - not sure how that plays out in other parts of the app, but feel limiting them is a good idea. not immediately sure what that limit should be.

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// loop is only terminated by returning from the function.
+	for {
+		// idempotently reslice for next chunk operation.
+		ids = ids[:0]
+		toDelete = toDelete[:0]
+
+		// collect ids which the provided updater created.
+		// note arguments via closure
+		err := func() error {
```
ldelossa, line 178: wrapped everything in closures as a cheap/easy way to always just defer the con or batch resolution closing and not run into confusion.

internal/vulnstore/postgres/store.go
```diff
@@ -86,83 +84,3 @@ func (s *Store) Get(ctx context.Context, records []*claircore.IndexRecord, opts
 	}
 	return vulns, nil
 }
-
```
ldelossa, line 89: Winded up being too much code to just shove into the same store.go file, GC code now lives in its own file.

internal/vulnstore/postgres/gc.go
hdonnay: yeah, not obvious. We should figure out a way to A/B test this.

ldelossa: Should it stay a constant in that case? Or should this become a config

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// unconditionally wait for all in-flight go routines to return.
+	// the use of context.Background and lack of error checking is intentional.
+	// all in-flight go routines are guarantee to release their sems.
+	sem.Acquire(context.Background(), int64(runtime.NumCPU()))
+
+	close(errC)
+	for e := range errC {
+		if e != nil {
+			return totalOps - deletedOps, fmt.Errorf("received error issuing deletes: %v", err)
```
hdonnay, line 78: This only ever returns the first error in the channel, which isn't _necessarily_ the first error encountered. I think it'd be worth the cycles to build a comprehensive error message in this path.

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+		var updater string
+		err := rows.Scan(&updater)
+		switch err {
+		case nil:
+			// hop out
+		default:
+			return nil, fmt.Errorf("error scanning updater: %v", err)
+		}
+		updaters = append(updaters, updater)
+	}
```
hdonnay, line 110: need to check `rows.Err()`

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+		tmp := pgtype.UUIDArray{}
+		err := rows.Scan(&tmp)
+		if err != nil {
+			return nil, total, fmt.Errorf("error scanning update operations: %w", err)
+		}
+		for _, u := range tmp.Elements {
+			total++
+			m = append(m, u.Bytes) // this works since [16]byte value is assignable to uuid.UUID
+		}
+	}
```
hdonnay, line 154: need to check `rows.Err()`

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+	return m, total, nil
+}
+
+func chunkedCleanup(ctx context.Context, pool *pgxpool.Pool, updater string) error {
+	const (
+		paginatedSelect = `
+SELECT id FROM vuln WHERE vuln.updater = $1 and id > $2 ORDER BY id ASC LIMIT 10000
```
hdonnay, line 161:
```suggestion
SELECT id FROM vuln WHERE vuln.updater = $1 AND id > $2 ORDER BY id ASC LIMIT 10000;
```

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+	// loop is only terminated by returning from the function.
+	for {
+		// idempotently reslice for next chunk operation.
+		ids = ids[:0]
+		toDelete = toDelete[:0]
+
+		// collect ids which the provided updater created.
+		// note arguments via closure
+		err := func() error {
+			rows, _ := pool.Query(ctx, paginatedSelect, updater, largestID)
```
hdonnay, line 184: Check the error

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+			defer rows.Close()
+			for rows.Next() {
+				var id int64
+				err := rows.Scan(&id)
+				if err != nil {
+					return fmt.Errorf("encountered error while scanning vuln id: %v", err)
+				}
+				ids = append(ids, id)
+			}
```
hdonnay, line 193: check `rows.Err()`

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+		toDelete = toDelete[:0]
+
+		// collect ids which the provided updater created.
+		// note arguments via closure
+		err := func() error {
+			rows, _ := pool.Query(ctx, paginatedSelect, updater, largestID)
+			defer rows.Close()
+			for rows.Next() {
+				var id int64
+				err := rows.Scan(&id)
```
hdonnay, line 188: This can scan into the backing array directly by reslicing and taking the last element. Or by having the reslice above be to `len == cap`, counting the iterations here, then slicing to the correct length.

hdonnay (changes requested):

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+	)
+
+	// obtain update operations which need deletin'
+	ops, totalOps, err := eligibleUpdateOpts(ctx, s.pool, keep)
+	if err != nil {
+		return 0, err
+	}
+
+	// delete em', but not too many...
+	if totalOps >= GCThrottle {
```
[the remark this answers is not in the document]
ldelossa: I think the normal compare is a bit cleaner, just an opinion.

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+WITH ordered_ops AS (
+    SELECT array_agg(ref ORDER BY date DESC) AS refs FROM update_operation GROUP BY updater
+)
+SELECT ordered_ops.refs[$1:]
+FROM ordered_ops
+WHERE array_length(ordered_ops.refs, 1) > $2;
+`
+	)
+
+	var total int64
```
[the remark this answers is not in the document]
ldelossa: done

libvuln/updates/manager.go
```diff
@@ -206,15 +203,17 @@ func (m *Manager) Run(ctx context.Context) error {
 
 	// unconditionally wait for all in-flight go routines to return.
 	// the use of context.Background and lack of error checking is intentional.
-	// all in-flight go routines are gauranteed to release their sems.
+	// all in-flight go routines are guarantee to release their sems.
```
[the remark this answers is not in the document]
ldelossa: meant to fix the spelling, butchered it more.

internal/vulnstore/postgres/gc.go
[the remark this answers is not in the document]
ldelossa: I want to keep totalOps around for later, but slice the returned ops array. That's why I return both the total count, and the original array. 

I'm partial to how that reads, as the ops becomes the working set, and totalOps is a immutable count of total operations that would have been done, if GC was not throttled.

internal/vulnstore/postgres/gc.go
ldelossa: Still a little confused why that's necessary.

Isn't this only useful if we *don't* defer the rows.Close() ? rows.err is set on Close() https://github.com/jackc/pgx/blob/master/rows.go#L132 

In other cases rows.err is returned to you directly: https://github.com/jackc/pgx/blob/master/conn.go#L608

https://github.com/jackc/pgx/blob/master/rows.go#L76

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return totalOps - deletedOps, err
+	}
+
+	// get all updaters we know about.
+	updaters, err := distinctUpdaters(ctx, s.pool)
+
+	// issue concurrent chunk'd deletion for known updaters
+	// limit concurrency by available CPUs.
+	sem := semaphore.NewWeighted(int64(runtime.NumCPU()))
```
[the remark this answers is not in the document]
ldelossa: ohhh good catch.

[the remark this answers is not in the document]
ldelossa: can see this, "why does clair break on my vm when I change the CPU core settings during runtime"

internal/vulnstore/postgres/gc.go
ldelossa: Do we care about "first" error, since they can all run in parallel? I'm not necessary seeing why we care about "first" error - order isn't really of concern AFAICT.

internal/vulnstore/postgres/gc.go
ldelossa: explained out of band. the checks are needed.

internal/vulnstore/postgres/gc.go
ldelossa: going to skip on this one unless the var creation becomes an issue after profiling, think it adds just a bit more to grok on an already pretty gnarly function.

internal/vulnstore/postgres/gc.go
hdonnay: I think we need to figure out a system to observe it first, then we can work in a toggle. This is a purely internal data structure, so I'm not worries.

internal/vulnstore/postgres/gc.go
hdonnay: Right, but as it stands, this just returns one of the errors, not all of them. The first one to occur is probably going to have the most informative error, but this just grabs a random one.

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,278 @@
...
+// operation.
+func distinctUpdaters(ctx context.Context, pool *pgxpool.Pool) ([]string, error) {
+	const (
+		// will always contain at least two update operations
+		selectUpdaters = `
+SELECT DISTINCT(updater) FROM update_operation;
+`
+	)
+	rows, err := pool.Query(ctx, selectUpdaters)
+	defer rows.Close()
```
hdonnay, line 98: Shouldn't this be after checking the returned error?

internal/vulnstore/postgres/gc.go
ldelossa: Alright, probably easiest to round up all the errors into a channel then, and build an error string like we do with updaters. 

Dont want to add a sync.Once here if i can avoid it.

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,278 @@
...
+
+	deletedOps, err := s.DeleteUpdateOperations(ctx, ops...)
+	if err != nil {
+		return totalOps - deletedOps, err
+	}
+
+	// get all updaters we know about.
+	updaters, err := distinctUpdaters(ctx, s.pool)
+	if err != nil {
+		return totalOps - deletedOps, nil
```
hdonnay, line 50: So if it was unable to get the updaters, the error is just swallowed?

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,278 @@
...
+			return nil
+		}()
+		if err != nil {
+			return err
+		}
+
+		// delete any vulns accumulated
+		// note arguments via closure
+		err = func() error {
+			if len(toDelete) <= 0 {
```
hdonnay, line 254: How can the length be less than zero?

internal/vulnstore/postgres/gc.go
```diff
@@ -0,0 +1,278 @@
...
+		shouldDelete = `
+SELECT NOT EXISTS(SELECT 1 FROM uo_vuln WHERE vuln = $1);
+`
+		deleteVuln = `
+DELETE FROM vuln WHERE id = $1;
+`
+	)
+
+	ids := make([]int64, 0, 10000)
+	toDelete := make([]int64, 0, 10000)
```
hdonnay, line 179: this allocates 160MB per call, it might be better to let the `id` slice get grown as it goes around the loop and have the function at L225 create the batch used in the function at L253.

hdonnay (changes requested):
The error handling in the GC method should aggregate all the errors, not just grab whichever one got into the channel first.

I think there's a few odd error paths, and perhaps a more memory efficient way to issue the deletes.

internal/vulnstore/postgres/gc.go
ldelossa: nope that's a bug.

hdonnay (approved):
LGTM

hdonnay (approved):

Issue #308, opened 2021-02-10
There is now way to limit number of indexer's DB connections


jan-zmeskal:
So I recently deployed clair from [this commit](https://github.com/quay/clair/commit/eb519e0752d3cf7f5f8daeefd4ad9bd29cbfa8c2) to quite a big environment. The reason why I mention the commit is that before the deployment, I did not have @hdonnay 's postgres refactor deployed. I don't know if that's significant information, but I think it might given [these changes](https://github.com/quay/claircore/pull/266/files#diff-a7d7db14845dc0c99ee7c38e117721015b9406ad32a63544ade17750142b0834). What happened is that indexer pods used up way more DB connections than I expected. Details of the issue are described in my message to @ldelossa.
![image](https://user-images.githubusercontent.com/22600243/107488618-e617ab80-6b87-11eb-8917-08e016ff407a.png)

I read the code for a bit since then and as far as I can understand, there's no real way to ultimately limit the overall number of how many DB connections one indexer instance creates. Keep in mind that my understanding might be wrong, but let me walk you through it. 

First the easy part. One connection pool is created during [store initialization](https://github.com/quay/claircore/blob/533316c18965c9007f1fb60bead2f33cf012d715/libindex/init.go#L25). The number of max connections is hard-coded to 30. As described in quay/clair/issues/1179, it'd be nice to have some control over that. But that's not the main problem because we could just scale down the number of indexer pods to deal with that. I think the catch is in different place.

When a POST request is made to `index_report` endpoint, [Index](https://github.com/quay/claircore/blob/533316c18965c9007f1fb60bead2f33cf012d715/libindex/libindex.go#L91) method of `Libindex` is eventually called. That in turn makes call to [controllerFactory](https://github.com/quay/claircore/blob/533316c18965c9007f1fb60bead2f33cf012d715/libindex/controllerfactory.go#L20) function. And each call to that function creates a new connection pool with options acquired from connstring. If an option is not specified in connstring, it falls back to some sane defaults chosen by library authors. For [number of maximum connections](https://github.com/jackc/pgx/blob/3568b908c03beaf0846015658ca9fd3123c51823/pgxpool/pool.go#L15), it's 4. Furthermore, value of `defaultMaxConnIdleTime` is 30 minutes. With all this being said, I think it's apparent now that when there's a lot of index requests, one might run out of DB connections.

Now please correct me if I'm wrong in my analysis. But if not, I think we need a mechanism how to limit the overall number of DB connections indexer may make.

[...]

PR #310, opened 2021-02-10
libindex: limit MaxConns in controller pool to 1

jan-zmeskal:
With the default number of connections controller pool can make set to 4, indexer tends to get quite greedy and might bully matcher and notifier. For in-depth discussion of this issue, see #308

### DIFF
--- a/libindex/controllerfactory.go
+++ b/libindex/controllerfactory.go
@@ -3,6 +3,7 @@ package libindex
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/jackc/pgx/v4/pgxpool"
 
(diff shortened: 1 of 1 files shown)


ldelossa (approved):
@jan-zmeskal lets merge this maybe after 24 hours of no issues in your testing env?

jan-zmeskal:
@ldelossa Sounds like a plan. I'll keep you posted.

jan-zmeskal:
So far it seems to be okay. Indexer uses ~ 40 connections when nothing much is happening. During the night, security data got updated so it seems there was a surge in index requests. During that time, we received an alert from notifier. So my best guess is that during the peak, indexer ate up the connections and notifier had none left. That's of course not good, but from what I can gather it happened only during the peak, which is improvement.

PR #311, opened 2021-02-10
vulnstore: update-diff optimize

ldelossa:
small optimization seems to shave ~800ms off our largest update diff
query.


hdonnay (approved):
LGTM

I rebased and pushed because I forgot to hit the button.

PR #313, opened 2021-02-12
rhel: fix cpe mapping type assertion

ldelossa:
this commit makes sure a nil pointer of type *repo2cpe.MappingFile is
loaded into the atomic.Value.

type assertions of the atomic.Value should never fail after this.
Loads() of the atomic.Value before a mapping file is downloaded should
return a typed nil.


hdonnay (approved):
LGTM

I didn't re-read all of the, but I guess there's a case where the request can error and nothing gets stored and this fixes it.

ldelossa:
Yup that's correct, if the first "Update" of a mapping file fails, an untyped nil is present in the atomic.Value - and then type assertion fails.

hdonnay (approved):

Issue #236, opened 2020-09-28
Support vulnerability listing for maven central repo artifacts


arajkumar:
Maven is a package repository for Java. 

The `mvn` tool which packs the maven projects usually adds metadata fields(dependency name, version) into `META-INF/maven/<groupId>/<artifactId>` which could be found on every `jar` file.

There are two files with in every maven jar which would give us the necessary information like `package|version`,
- META-INF/maven/<groupId>/<artifactId>/pom.properties
- META-INF/maven/<groupId>/<artifactId>/pom.xml

> The pom.xml and pom.properties files are packaged up in the JAR so that each artifact produced by Maven is self-describing and also allows you to utilize the metadata in your own application if the need arises. One simple use might be to retrieve the version of your application. Operating on the POM file would require you to use some Maven utilities but the properties can be utilized using the standard Java API and look like the following:

```
#Generated by Maven
#Tue Oct 04 15:43:21 GMT-05:00 2005
version=1.0-SNAPSHOT
groupId=com.mycompany.app
artifactId=my-app
```

Reference: https://maven.apache.org/guides/getting-started/#how-do-i-add-resources-to-my-jar

arajkumar:
I would be happy to contribute if there are no concerns on this ask. @ldelossa @hdonnay

ldelossa:
@arajkumar does maven provide its own security db? If so can you link us to it? 

This seems like a fine thing to do. 

You will need to write a package scanner which finds and peeks into .jar files. IIRC jar is just a zip formatted archive so that seems fine. 

Writing a parser for the xml items would be necessary.

arajkumar:
> @arajkumar does maven provide its own security db? If so can you link us to it?
As now my plan to use CRDA remote matcher for maven, similar to what we are trying for Python #203.

But in the long run we have to plug it with NVD(CPE as matching criteria) for non CRDA use cases.,(similar to https://github.com/jeremylong/DependencyCheck).

ldelossa:
Okay, just keep in mind that we are unable to support client notifications unless we index the database of vulnerabilities. Any CRDA provided results will not provide notifications to clients when contents has changed (since we cannot detect the change)

arajkumar:
Trivy is using Github Advisory as a data source: https://github.com/advisories?query=ecosystem%3Amaven

Gitlab also has public security advisory database for maven: https://gitlab.com/gitlab-org/security-products/gemnasium-db/-/tree/master/maven

PR #329, opened 2021-02-22
indexer: Implement package indexer for maven

arajkumar:
Maven is a package repository for Java.

The mvn tool which packs the maven projects usually adds metadata fields(dependency name, version) into `META-INF/maven/<groupId>/<artifactId>` which could be found on every jar file.

There are two files with in every maven jar which would give us the necessary information like package|version,

```
META-INF/maven///pom.properties
META-INF/maven///pom.xml
```
>[1] The pom.xml and pom.properties files are packaged up in the JAR so that each artifact produced by Maven is self-describing and also allows you to utilize the metadata in your own application if the need arises. One simple use might be to retrieve the version of your application. Operating on the POM file would require you to use some Maven utilities but the properties can be utilized using the standard Java API and look like the following:

```
version=1.0-SNAPSHOT
groupId=com.mycompany.app
artifactId=my-app
```

[1] https://maven.apache.org/guides/getting-started/#how-do-i-add-resources-to-my-jar

`github.com/aquasecurity/go-dep-parser/pkg/jar` has been used to parse jar/war/ear files.

Fixes https://github.com/quay/claircore/issues/236

Signed-off-by: Arunprasad Rajkumar <ar.arunprasad@gmail.com>


arajkumar:
Trivy is using Github Advisory as a data source: https://github.com/advisories?query=ecosystem%3Amaven

Gitlab also has public security advisory database for maven: https://gitlab.com/gitlab-org/security-products/gemnasium-db/-/tree/master/maven

java/packagescanner.go
```diff
@@ -0,0 +1,111 @@
...
+	libs, err := jar.Parse(r, jar.WithFilePath(name))
+	if err != nil {
+		return nil, err
+	}
+	packages := make([]*claircore.Package, len(libs))
+	for i, l := range libs {
+		packages[i] = &claircore.Package{
+			Name:           l.Name,
+			Version:        l.Version,
+			PackageDB:      "maven:" + name,
```
ldelossa, line 105: Referencing the python package scanner, is there any path fixing that needs to occur here, similar to:
https://github.com//ldelossa/claircore/blob/b165424edcf12301661573730041058b505d716c/python/packagescanner.go#L81 ?

ldelossa (changes requested):

java/packagescanner.go
arajkumar: It is not necessary here, we just check at the extension without any path.

ldelossa: Does this result in the PackageDB field showing where the java package exists in the file system?

arajkumar: Yeah, that is a valid question. I assumed it should have full file path.

However the python implementation seem to be recording the directory which holds the package in filesystem. Let me fix it.

ldelossa: Yes, this field should be the file system location where the package was found. It's used for uniqueness so we can find the same package across multiple file system locations.

ldelossa (approved):

PR #346, opened 2021-03-22
crda: simplify remotematcher implementaion

arajkumar:
This PR simplifies the crda remotematcher using POST api which takes care of concurrent processing at the server end.

Signed-off-by: Arunprasad Rajkumar <ar.arunprasad@gmail.com>


ldelossa (approved):
I'm not too sure why there are changes to struct fields since this seems to be using a new API feature. But I don't see anything wrong in this PR. LGTM.

arajkumar:
Thanks @ldelossa 
>I'm not too sure why there are changes to struct fields since this seems to be using a new API feature

It is a known difference, though the structure should be same, we made some incompatible changes in the response structure between GET and POST.

PR #350, opened 2021-03-24
docs: note default updater URLs

hdonnay:
Generated a list to look at with this command:

    git grep '["`]https\?://' -- ':/*.go' ':!*_test.go' :^test :^cmd


ldelossa (changes requested):
I think providing the script to gen that list in the commit is going to be lost.

possible to cook up a go generate file for this? 

At the least, place the script to get the urls somewhere more visible.

hdonnay:
It's not an automated process to create this list, so I can't just write a script to do it.

We could (and probably should) have a magic comment and a script that pulls it out into the documentation, but that's more work than just documenting the state of things.

ldelossa:
Ohh I see, thought I could just take that script and run it at root and generate the list.

ldelossa (approved):

PR #351, opened 2021-03-24
python: force re-fetch/parse

ldelossa:
recent changes to the python updater were made but
pyup databse updates slowly.

this migration forces a re-fetch/parse of the pyup
database, fixing python matching in the upgrade path.


hdonnay (approved):
LGTM

PR #353, opened 2021-03-24
ratelimiter

hdonnay:
This is going to be changes to make sure that if a caller (like clair) configures an http.Client, it's actually used everywhere.


hdonnay:
This looks worse on first blush than it actually is.

ldelossa:
Can you test this by setting the update interval rather low and running a clair cluster for somewhere around 24 hours? Threading this code thru then entire app warrants that degree of testing IMO.

ldelossa:
> Can you test this by setting the update interval rather low and running a clair cluster for somewhere around 24 hours? Threading this code thru then entire app warrants that degree of testing IMO.

Waiting to see how this plays out, since this is a pretty over arching change that effects all updaters

hdonnay:
ran a build of clair with this changeset for 24h without incident. I can post logs, but they're extremely uneventful.

Latest push is a rebase onto current `main`.

ldelossa (approved):

ldelossa (approved):

PR #355, opened 2021-03-25
Louis/backport 351

ldelossa:
This backports #351.

It was necessary to pull in "migration2.go" from commit: 78564563897a3f1573987ae05fb6ecb6539abe57 in order to not break the pending upgrade path from Clair 4.0 -> Clair 4.1


hdonnay (approved):
LGTM

We discussed a migration policy that should make these issues simpler in the future

PR #360, opened 2021-03-29
introspection: datastore metrics

ldelossa:
implements prometheus database metrics for the indexer store
per our specification: https://github.com/quay/clair/discussions/1167


internal/indexer/postgres/affectedmanifest.go
```diff
@@ -91,6 +131,11 @@ WHERE
 	// collect all packages which may be affected
 	// by the vulnerability in question.
 	pkgsToFilter := []claircore.Package{}
+
+	done := metric.WithDuration(ctx, func(d time.Duration) {
+		affectedManifestsCounter.WithLabelValues("selectPackages").Add(1)
+		affectedManifestsDuration.WithLabelValues(("selectPackages")).Observe(d.Seconds())
+	})
```
hdonnay, line 138:
```suggestion

	start := time.Now()
```

internal/indexer/postgres/affectedmanifest.go
```diff
@@ -99,6 +144,7 @@ WHERE
 	default:
 		return nil, fmt.Errorf("failed to query packages associated with vulnerability %+v: %v", v, err)
 	}
+	done()
```
hdonnay, line 147:
```suggestion
	defer func(d time.Duration) {
		affectedManifestsCounter.WithLabelValues("selectPackages").Add(1)
		affectedManifestsDuration.WithLabelValues(("selectPackages")).Observe(d.Seconds())
	})(time.Now())
```

hdonnay (commented):
I think doing it like this is much simpler.

It doesn't leak contexts, doesn't spawn a goroutine, and only does `defer`+`time.Now()` in the main path and does the submission during the defers being run.

internal/indexer/postgres/affectedmanifest.go
hdonnay: This is wrong, the `d.Seconds()` should be `d.Sub(start).Seconds()`

hdonnay (dismissed):
LGTM

Just to double-check, this doesn't record any stats about queries that error out, is that on purpose?

hdonnay (dismissed):

internal/indexer/postgres/setindexreport.go
```diff
@@ -31,10 +56,16 @@ DO
...
+
+	start := time.Now()
 	_, err := s.pool.Exec(ctx, query, ir.Hash, jsonbIndexReport(*ir))
 	if err != nil {
 		return fmt.Errorf("failed to upsert index report: %v", err)
 	}
+	defer func(d time.Duration) {
+		setIndexReportCounter.WithLabelValues("query").Add(1)
+		setIndexReportDuration.WithLabelValues("query").Observe(d.Seconds())
+	}(time.Since(start))
```
hdonnay, line 68: leftover defer

hdonnay (changes requested):

hdonnay (approved):

PR #363, opened 2021-04-08
add go binary package scanner

hdonnay:
Some outstanding questions:

How should the dependency and main module be represented?

Should we scan for a `go.mod` file?

Will the main module ever have a version?


scanner/gobin/gobin.go
```diff
@@ -0,0 +1,326 @@
...
+	for h, err = tr.Next(); err == nil; h, err = tr.Next() {
+		fi := h.FileInfo()
+		m := fi.Mode()
+		switch {
+		case !m.IsRegular():
+			continue
+		case m.Perm()&0555 == 0:
+			// Not executable
+			continue
+		case fi.Size() > toobig:
```
ldelossa, line 211: Curious if an arbitrary file size limit is necessary here? 

We don't care about reading the entire binary file, only the ELF header. With some finagling of the ELF header fields I think you can get the size of the elf header: 

```
      e_phnum
              This member holds the number of entries in the program
              header table.  Thus the product of e_phentsize and e_phnum
              gives the table's size in bytes.  If a file has no program
              header, e_phnum holds the value zero.
```

I'm thinking at that point, you don't care about the full Golang binary, you just slice off the ELF header bits.

hdonnay: Could also reduce the likelihood of that by peaking for the elf/pe header

ldelossa: That's along the lines I'm getting at here, if you can deduce the header sizes, you should be able to peak, and then only read out the portion of the file you need. I had the assumption this data is kept in something like PT_NOTE, but not too sure. I imagine the headers give you some information on where to peak/truncate/offset into to find what you need. 

I don't think this arbitrary limit is an issue, but if you can nix it somehow, it would be nice.

arajkumar:
@hdonnay Any specific reason for not using `https://github.com/aquasecurity/go-dep-parser/tree/main/pkg/gobinary` for parsing the go exe?

ldelossa:
Note from comm dev: maybe worth standarizing the string encoding of a packagedb:package when the package is solely housed in the database (not on filesystem).

scanner/gobin/exe.go
```diff
@@ -0,0 +1,60 @@
...
+	vs := map[string]string{
+		"runtime": bi.GoVersion,
+	}
+	*out = append(*out, &claircore.Package{
+		Kind:      claircore.BINARY,
+		PackageDB: pkgdb,
+		Name:      bi.Main.Path,
+		Version:   bi.Main.Version,
+	})
+	if ev.Enabled() {
```
crozzy, line 42: Is this to save on allocations if debugging isn't enabled?

crozzy (commented):
Nice, a few comments, nothing blocking that I can see

scanner/gobin/gobin.go
```diff
@@ -0,0 +1,153 @@
...
+		return s.err
+	}
+	if _, err := s.File.Seek(0, io.SeekStart); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *spoolfile) setup() {
+	f, err := os.CreateTemp("", "gobin.spool.*")
```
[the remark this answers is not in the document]
hdonnay: I don't think this makes a lot of sense, as these only briefly exist on disk. They're unlinked in the same function.

scanner/gobin/exe.go
hdonnay: Yes -- no point in building a big map if it's not going to be printed.

scanner/gobin/gobin.go
```diff
@@ -0,0 +1,153 @@
...
+		fi, err := d.Info()
+		if err != nil {
+			return err
+		}
+		m := fi.Mode()
+		switch {
+		case !m.IsRegular():
+			return nil
+		case m.Perm()&0o555 == 0:
+			// Not executable
```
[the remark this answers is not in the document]
hdonnay: Yes

scanner/gobin/gobin.go
```diff
@@ -0,0 +1,153 @@
+// Package gobin implements a package scanner that pulls go runtime and
```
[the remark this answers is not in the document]
hdonnay: Moved it -- need to eventually get things out of the root

updater/osv/cvss.go
```diff
@@ -21,11 +21,11 @@ func fromCVSS(s string) (sev claircore.Severity, err error) {
 	// Giant switch ahoy
 	var ns [8]float64
 	for _, m := range ms[1:] {
-		vec := strings.SplitN(m, ":", 2) // TODO(hank) go1.18: use strings.Cut
-		if len(vec) != 2 {
+		n, v, ok := strings.Cut(m, ":")
```
crozzy, line 24: :D

crozzy (approved):

PR #366, opened 2021-04-16
[WIP] Bounded concurrency affected manifests

crozzy:
This concurrency pattern may help with DB load, particularly burst load. Instead of starting 100 goroutines, waiting for them to all finish, then firing off 100 more, this change will block creating new goroutines until a currently running one has returned. 

This means you aren't waiting for the slowest routine in every 100. In this case, where it's a DB lookup with the same query, all the routines probably complete within a similar time bound, but it may help spread the DB load rather than being spikey.

TODO
- [x] Find a good way to benchmark
- [x] Clean up tests
- [x] Find a good concurrency number
- [x] Discuss concurrency number as opt (as the pg instance is variable depending on deployment)

### DIFF
--- a/libindex/libindex.go
+++ b/libindex/libindex.go
@@ -178,8 +178,7 @@ func (l *Libindex) IndexReport(ctx context.Context, hash claircore.Digest) (*cla
 
 // AffectedManifests retrieves a list of affected manifests when provided a list of vulnerabilities.
 func (l *Libindex) AffectedManifests(ctx context.Context, vulns []claircore.Vulnerability) (*claircore.AffectedManifests, error) {
-	// This concurrency number could be another opt for Libindex
-	sem := semaphore.NewWeighted(50)
+	sem := semaphore.NewWeighted(20)
 	ctx = baggage.ContextWithValues(ctx,
 		label.String("component", "libindex/Libindex.AffectedManifests"))
 
--- a/libindex/libindex.go
+++ b/libindex/libindex.go
@@ -178,6 +178,7 @@ func (l *Libindex) IndexReport(ctx context.Context, hash claircore.Digest) (*cla
 
 // AffectedManifests retrieves a list of affected manifests when provided a list of vulnerabilities.
 func (l *Libindex) AffectedManifests(ctx context.Context, vulns []claircore.Vulnerability) (*claircore.AffectedManifests, error) {
+	// This concurrency number could be another opt for Libindex
 	sem := semaphore.NewWeighted(50)
 	ctx = baggage.ContextWithValues(ctx,
 		label.String("component", "libindex/Libindex.AffectedManifests"))
--- a/libindex/libindex.go
+++ b/libindex/libindex.go
@@ -6,7 +6,6 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
-	"math"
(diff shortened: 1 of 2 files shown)


ldelossa:
I dig this, a lot of the original chunked-loop code has been moving to using semaphores else where.
Discuss concurrency number as opt (as the pg instance is variable depending on deployment)
>  Find a good concurrency number

Thoughts on runtime.NumCPU ? 
A consistent multiplier can be provided if we want to over-commit go routines to available cores. 

> Discuss concurrency number as opt (as the pg instance is variable depending on deployment)

My feelings lately have been to scale concurrency limits based on the system, and not allow those
parameters to be tuned externally (with some max limit to bound away spamming Postgres). 

I can be persuaded otherwise tho with valid use cases where the tuning is useful, and doesn't 
cause potential arm to the operators.

crozzy:
> I dig this, a lot of the original chunked-loop code has been moving to using semaphores else where.
> 
> > Find a good concurrency number
> 
> Thoughts on runtime.NumCPU ?
> A consistent multiplier can be provided if we want to over-commit go routines to available cores.

My thought was that; as the work is IO (looking up from the DB) tying the # of goroutines to CPU might not be a good metric (as the context switching is going to happen on the IO blocking). With that said, I guess whatever the Clair recommendations are for Postgres max_connections (or similar config) would be closer to the bottleneck number. This could be an incorrect hypothesis though.

> 
> > Discuss concurrency number as opt (as the pg instance is variable depending on deployment)
> 
> My feelings lately have been to scale concurrency limits based on the system, and not allow those
> parameters to be tuned externally (with some max limit to bound away spamming Postgres).
> 
> I can be persuaded otherwise tho with valid use cases where the tuning is useful, and doesn't
> cause potential harm to the operators.

I agree with this sentiment, it's probably infinitely more frustrating to debug bad configs than live with a 1 size fits all number. My kneejerk is to make everything configurable (and avoid the job of deciding on that number :laughing: ).

ldelossa:
> My thought was that; as the work is IO (looking up from the DB) tying the # of goroutines to CPU might not be a good metric (as the context switching is going to happen on the IO blocking). With that said, I guess whatever the Clair recommendations are for Postgres max_connections (or similar config) would be closer to the bottleneck number. This could be an incorrect hypothesis though.

ldelossa:
Unintentionally closed, my bad. 

> My thought was that; as the work is IO (looking up from the DB) tying the # of goroutines to CPU might not be a good metric (as the context switching is going to happen on the IO blocking). With that said, I guess whatever the Clair recommendations are for Postgres max_connections (or similar config) would be closer to the bottleneck number. This could be an incorrect hypothesis though.

We set a max conn pool limit to 30 in libindex. 

My thinking was using core count, as we want each "parked" goroutine to have an available cpu with little contention, as contention would mean conns being held for longer then necessary. 

I think I undersatnd your point about context_switching tho. 

We could set a static value for now, somewhere between the average core count of a machine, and our 30 max conn limit. 

We also may want to do some analysis on that max conn pool limit under load. This has not been done far warning.

ldelossa:
@crozzy needs rebase, but will approve after.

ldelossa (approved):
Looks good. Nice work.

PR #372, opened 2021-04-21
chore: use map to get unique CPEs instead of slice

RTann:
This was kind of bothering me, so I figured I might as well make a PR. I figured using a set and then converting to a slice would be better than searching through a slice each time.

Signed-off-by: RTann <rtann@redhat.com>


ldelossa:
I'm not sure what's more expensive here, hassling the key values repeatedly, or iterating. Interesting is very fast.

hdonnay:
Going to close this, because I don't feel strongly about using the map instead of a slice.

Thank you for the contribution, though.

PR #377, opened 2021-04-28
Louis/enrichment datamodel

ldelossa:
Draft pull request to implement Enrichment data model changes per:
https://github.com/quay/clair/discussions/1229

### DIFF
--- a/internal/vulnstore/postgres/getupdateoperations.go
+++ b/internal/vulnstore/postgres/getupdateoperations.go
@@ -76,19 +76,93 @@ var (
 )
 
 // GetLatestUpdateRef implements driver.Updater.
-func (s *Store) GetLatestUpdateRef(ctx context.Context) (uuid.UUID, error) {
-	const query = `SELECT ref FROM update_operation ORDER BY id USING > LIMIT 1;`
+func (s *Store) GetLatestUpdateRef(ctx context.Context, kind driver.UpdateKind) (uuid.UUID, error) {
+	const (
+		query              = `SELECT ref FROM update_operation ORDER BY id USING > LIMIT 1;`
+		queryEnrichment    = `SELECT ref FROM update_operation WHERE kind = 'enrichment' ORDER BY id USING > LIMIT 1;`
+		queryVulnerability = `SELECT ref FROM update_operation WHERE kind = 'vulnerability' ORDER BY id USING > LIMIT 1;`
+	)
 	ctx = baggage.ContextWithValues(ctx,
 		label.String("component", "internal/vulnstore/postgres/getLatestRef"))
 
+	var q string
+	var label string
+	switch kind {
+	case "":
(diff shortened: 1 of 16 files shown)


internal/vulnstore/postgres/getupdateoperations.go
```diff
@@ -180,50 +256,45 @@ func getUpdateOperations(ctx context.Context, pool *pgxpool.Pool, updater ...str
...
-				rows.Close()
-				return nil, fmt.Errorf("failed to scan update operation for updater %q: %w", u, err)
-			}
-		}
-		rows.Close()
-		if err := rows.Err(); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to scan update operation for updater %q: %w", uo.Updater, err)
 		}
-		out[u] = ops
```
hdonnay, line 223: I'm a little worried about excess copying doing it this way, but it's certainly simpler to read.

internal/vulnstore/postgres/getupdateoperations.go
```diff
@@ -180,50 +256,45 @@ func getUpdateOperations(ctx context.Context, pool *pgxpool.Pool, updater ...str
...
-	}
-	if err := tx.Commit(ctx); err != nil {
-		return nil, fmt.Errorf("failed to commit transaction: %w", err)
```
hdonnay, line 226: I see why we wouldn't bother with the commit, but it's a little weird. Using `BeginTx` with a read-only option would call out this being read-only in code.

hdonnay (dismissed):
My one suggestion isn't going to affect correctness, so it's not super important.

internal/vulnstore/postgres/getupdateoperations.go
ldelossa: Where are the concerns?

internal/vulnstore/postgres/getupdateoperations.go
ldelossa: Yeah I'm not sure why we went with a transaction here to begin with, I think we can get rid of it and make this a read only method all together, right?

internal/vulnstore/postgres/getupdateoperations.go
hdonnay: The previous version allocates a local slice, grows it and populates in-place, then adds it to the map, so the slices in the map never need to be updated once they escape. This version allocates a `uo` local, populates it, then potentially reallocates the updater's slice copies it into the slice.

internal/vulnstore/postgres/getupdateoperations.go
hdonnay: I thought the transaction was needed to make sure the same snapshot of the table is used, since we're issuing multiple queries.

hdonnay (approved):

PR #378, opened 2021-04-29
updater: log error from updater.manager.Run

arajkumar:
Error returned by `manager.Run` is not logged anywhere. Logging the same would ease the `Updater` debugging.

Signed-off-by: Arunprasad Rajkumar <arajkuma@redhat.com>


ldelossa:
Noticed this too and is fixed in: https://github.com/quay/claircore/pull/377

arajkumar:
@ldelossa , I see. Let me close this then.

PR #381, opened 2021-05-05
[dpkg scanner] support file names starting with "./"

Bezbran:
replace the suffix of "dpkg/testdata/sha256:3c9020349340788076971d5ea638b71e35233fd8e149e269d8eebfa17960c03f.replacethissuffix" to "dpkg/testdata/sha256:3c9020349340788076971d5ea638b71e35233fd8e149e269d8eebfa17960c03f.layer" and run the new test "TestScannerACMEImage" to see the impact.


dpkg/scanner.go
```diff
@@ -124,7 +124,10 @@ Find:
 		var db io.Reader
 		var h *tar.Header
 		for h, err = tr.Next(); err == nil; h, err = tr.Next() {
-			if h.Name == fn {
+			// in case we have "./" as prefix, the filepath.Dir(h.Name) in "Find" loop will return the path without the prefix
+			// so we must normalize it
+			fileName := strings.TrimPrefix(h.Name, "./")
+			if fileName == fn {
```
hdonnay, line 130:
```suggestion
			// The location from the above loop is cleaned, so make sure to do that.
			if c := filepath.Clean(h.Name); c == fn {
```

dpkg/scanner_test.go
```diff
@@ -825,6 +825,41 @@ func TestScanner(t *testing.T) {
 	}
 }
 
+func TestScannerACMEImage(t *testing.T) {
```
hdonnay, line 828: This test needs to actually fetch the relevant layer or craft a layer (like in `TestExtraMetadata`), checking in the layer won't work. Also, this name doesn't explain what it's actually testing.

hdonnay (changes requested):
This is a nice catch. This needs some changes, though:

- [ ] Layer isn't checked-in: It can be either created for the test or hosted and fetched.
- [ ] Use the `Clean` function.
- [ ] Commits signed-off (e.g. `git commit -s`).
- [ ] Commits formatted for changelog (e.g. `dpkg: clean filenames`).

hdonnay:
I've gone ahead and re-implemented this with the changes I requested. Thanks for the report and patch!

PR #392, opened 2021-05-26
new distlock

hdonnay:
This PR pulls in and integrates the new context-based distlock API.

### DIFF
--- a/internal/distlock/README.md
+++ b/internal/distlock/README.md
@@ -1,0 +1,87 @@
+# DistLock
+Dist lock implements a distributed lock via postgres.
+
+## Goals
+
+* No Transcations.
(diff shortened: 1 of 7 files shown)


internal/distlock/.github/workflows/main.yml
```diff
@@ -0,0 +1,21 @@
+---
```
ldelossa, line 1: I'm not sure if you want to keep these. 

I could not get docker-compose to work for the life of me in the GH-Actions environments. 

If you want to keep this just for completeness that's fine too, I don't think gh actions searches recursively for a .github folder and these files won't mess up the current actions, right?

ldelossa (approved):

PR #394, opened 2021-06-02
crda: remove matcher

hdonnay:
The CRDA matcher seems unable/unwilling to cope with the traffic that
comes from being enabled by default for every usage of claircore.

Planning on merging this in a few days' time unless anyone comes up with a better solution.


arajkumar (approved):
lgtm, @yzainee please ack it.

yzainee:
lgtm

PR #401, opened 2021-06-07
all: use GOMAXPROCS instead of NumCPU

hdonnay:
When running directly on an OS, these values will be the same. But in
containerized environments, the number of CPUs reported is the number on
the host, not the number allocated. Using the GOMAXPROCS means the
container orchestrator can configure things such that the go runtime is
less likely to constantly blow its cpu quota.


crozzy (dismissed):

hdonnay:
I don't understand why the action isn't firing here, but it's green on the fork: https://github.com/hdonnay/claircore/actions/runs/1009410939

PR #408, opened 2021-06-22
indexer: avoid creating a new db pool for the locks

crozzy:
This change instanciates the DB and passes the pool
around to who needs it. It also introduces a factory function
to create new locks as and when needed (with the same resources)

### DIFF
--- a/libindex/controllerfactory.go
+++ b/libindex/controllerfactory.go
@@ -2,16 +2,11 @@ package libindex
 
 import (
 	"context"
-	"fmt"
-	"time"
-
-	"github.com/jackc/pgx/v4/pgxpool"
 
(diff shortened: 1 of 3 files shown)


crozzy:
Still not finished thinking all the way through this so leaving it in draft for now, but feel free to comment

hdonnay (dismissed):
If I'm reading this right, this means we're closing over the pool now instead of constructing a single-connection pool on demand. This sounds right.

crozzy:
> If I'm reading this right, this means we're closing over the pool now instead of constructing a single-connection pool on demand. This sounds right.

exactly

hdonnay (dismissed):

hdonnay (approved):

PR #409, opened 2021-06-24
java: ignore whiteout files

hdonnay:
As described in the OCI image [spec][1], whiteout files are used to
indicate file or directory removals. The jar scanner was not considering
file prefixes and instead only looking at the suffixes.

[1]: https://github.com/opencontainers/image-spec/blob/master/layer.md#representing-changes

### DIFF
--- a/java/common.go
+++ b/java/common.go
@@ -1,0 +1,36 @@
+package java
+
+import (
+	"archive/tar"
+	"context"
+	"path"
+	"strings"
+
+	"github.com/quay/zlog"
+)
+
+// IsArchive reports whether the file described by the passed tar.Header is some
+// form of Java archive.
+//
+// Assumes slash-separated paths.
+func isArchive(ctx context.Context, h *tar.Header) bool {
+	base := path.Base(h.Name)
+Outer:
+	switch {
+	case h.Typeflag != tar.TypeReg:
+	case strings.HasPrefix(base, ".wh."):
+	default:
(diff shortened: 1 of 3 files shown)


java/common.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"archive/tar"
+	"context"
+	"path"
+	"strings"
+
+	"github.com/quay/zlog"
+)
+
+// IsArchive reports whether the file described by the passed tar.Header is some
```
crozzy, line 12: nit: comment capitalized

crozzy (commented):
LGTM

One question, what happens if a file is identified as `isArchive()` is "whitedout" in the next layer? Seems like a corner case, but I was wondering if we'd still identify that the container as having a java archive?

crozzy (approved):

jfrancin:
Can the reviewers who need to look at this please do so? There's a downstream issue in Red Hat's Clair scanner that this resolves. The issue is affecting scans of partner containers - they're blowing up attempting to scan these whiteout files, and it's causing one of our partners much consternation.

hdonnay:
Re: whiteouts: Yeah, currently we don't have a native way of modelling removals up to the manifest level.

java/common.go
hdonnay: I try to write doc comments as complete sentences, and capping the first word is just fallout of that.

PR #410, opened 2021-06-25
DB test harness

hdonnay:
These changes add new API to the `test/integration` package that uses [minimal postgres binaries](https://mvnrepository.com/artifact/io.zonky.test.postgres) to provide a database engine for tests.

Once cached locally, the binaries are used for tests regardless of the `integration` build tag.

This is inspired by [embedded-postgres](https://github.com/fergusstrange/embedded-postgres), but provides extra isolation for when tests are running in parallel and attempts to synchronize around creation of the caches.


test/integration/engine.go
```diff
@@ -0,0 +1,277 @@
...
+		ex = true
+		break
+	}
+	if !ex {
+		t.Fatal("didn't find txz")
+	}
+	t.Log("extraction OK")
+}
+
+func binUncached(t testing.TB) bool {
```
crozzy, line 235: Should we be returning `true` here if we get errors? We kind of have 3 (or more) states:

* Doesn't exist
* Exists and can get a lock (therefore fully available)
* Exists but error opening or getting the lock (I think flock will block until it gets the lock, but I guess the function can still return an error)

test/integration/db.go
```diff
@@ -163,3 +156,63 @@ func (db *DB) Close(ctx context.Context, t testing.TB) {
...
+
+// DBSetup queues setup and teardown for a postgres engine instance. If the
+// "integration" build tag is not provided, then nothing is done. If the
+// environment variable at EnvPGConnString is populated and the "integration"
+// build tag is provided, then the value of that environment variable is used
+// instead of an embedded postgres binary.
+//
+// This function is meant to be called from TestMain like so:
+//
+//	func TestMain(m *testing.M) {
```
crozzy, line 177: Wow, took me a few passes to work out what was going on here

test/integration/db_init.go
```diff
@@ -0,0 +1,74 @@
...
+func init() {
+	dbOS = runtime.GOOS
+	dbVersion = v12
+	dbArch = runtime.GOARCH
+
+	// See if a different version was requested
+	switch e := os.Getenv(EnvPGVersion); e {
+	case "13":
+		dbVersion = v13
+	case "12":
```
crozzy, line 25: No version 12?

test/integration/engine.go
```diff
@@ -0,0 +1,277 @@
...
+	p := filepath.Join(c, "clair-testing", fmt.Sprintf("postgres-%s-%s-%s", dbOS, dbArch, dbVersion))
+	_, err = os.Stat(p)
+	if err != nil {
+		return p, err
+	}
+	return p, nil
+}
+
+func downloadURL() string {
+	u, err := url.Parse(`https://repo1.maven.org`)
```
crozzy, line 267: perhaps a `const` for this and the path?

crozzy (commented):
This will be a great addition, so nice to not rely on external deps for integration tests.

Did you think about packaging the binaries with the code instead of downloading at runtime?

We can also update the comment in the Makefile that expresses the need for an external DB when running `make integration`.

test/integration/db_init.go
hdonnay: It's set above, I'll add a comment.

test/integration/engine.go
hdonnay: I've got it logging the error and then failing the test. I'll add some explicit commentary.

hdonnay:
> Did you think about packaging the binaries with the code instead of downloading at runtime?

I think this would really blow up the repo size, I'd rather keep the run-if-present, fetch-if-integration behavior.

crozzy (approved):

PR #413, opened 2021-07-13
postgres: fix query in GetLatestUpdateRefs

jvulgan:
This commit fixes query in GetLatestUpdateRefs which instead of
vulnerabilities was looking for enrichments.

Signed-off-by: Jakub Vulgan <jvulgan@redhat.com>


hdonnay (dismissed):
Looks like a bad copy-paste, nice catch.

hdonnay (approved):

PR #416, opened 2021-07-15
indexer: Add random number suffix to layer file

crozzy:
I noticed during loadtesting that it was possible to have
a filename collision during layer downloads from distinct indexing
requests that have reference the same layer in their manifest. While
one request finishes and deletes the file, the other request is left
with an unexpected EOF.

### DIFF
--- a/internal/indexer/fetcher/fetcher.go
+++ b/internal/indexer/fetcher/fetcher.go
@@ -9,12 +9,14 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/klauspost/compress/gzip"
 	"github.com/klauspost/compress/zstd"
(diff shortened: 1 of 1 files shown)


crozzy:
I'm not set on this approach, wanted to go for something that was light but useful, but happy to change it if something better is suggested

crozzy:
You can recreate this by indexing two images at the same time with the same base layer:
```
clairctl report --host {HOST} redis:buster
clairctl report --host {HOST} python:slim
```
Output:
```
indexer error: failed to scan all layer contents: unexpected EOF
```

crozzy:
After thinking about this some more I think it's going to be difficult to do this water tight with file-locks. Consider this situation:
Request one:

1. Creates a file
2. Locks the file
3. Writes to the file
4. Unlocks the file
5. Deletes the file

Meanwhile request two:

1. Tries to create file, gets a file already exists error
2. Tries to lock and may or may not get the lock (if request one is between stages 1 and 2 or 4 and 5)
3. If it gets the lock it's going to either fail itself or cause the other request to block
4. If it doesn't get the lock it will get it just before request 1 deletes the file

@hdonnay what am I missing here?

crozzy:
closing as addressed in https://github.com/quay/claircore/pull/453

PR #418, opened 2021-07-23
integration: multiple OS support

hdonnay:
Some linux-specific usage of the syscall package was added with the
embedded postgres addition. This splits out and cordons off all the
syscall usage so that this package builds on linux, darwin, and windows
again.


crozzy (approved):
LGTM

PR #419, opened 2021-07-26
crda: remote matcher with crda api.

rafiu007:
- [x] Add back crda remote matcher
- [x] Add new api
- [x] Add unit tests
- [x] Add configurable api-key
- [x] Add configurable source

 The default api in crda remote matcher will serve 100rpm
 The other key which can be optained by making an api call will be able to serve 1800 rpm to start with . We are positive about 
 increasing it even more.  
 
 Our api will have proper 3scale limits in place and we have removed the scenario which resulted in our system downtime.


crda/remotematcher.go
```diff
@@ -0,0 +1,404 @@
...
+)
+
+const (
+	// Bounded concurrency limit.
+	defaultBatchSize          = 100
+	defaultEndPoint           = "/api/v2/vulnerability-analysis"
+	defaultRequestConcurrency = 10
+	defaultURL                = "https://f8a-analytics-2445582058137.production.gw.apicast.io/api/v2/"
+	defaultSource             = "clair-upstream"
+	defaultKey                = "user_key=9e7da76708fe374d8c10fa752e72989f"
```
arajkumar, line 37: Isn't it defaultSource redundant? You should able to find the source using default user key from backend.
Also there is a `user agent` http header which can be used for this purpose.

crda/remotematcher.go
```diff
@@ -0,0 +1,404 @@
...
+
+var (
+	_ driver.Matcher             = (*Matcher)(nil)
+	_ driver.RemoteMatcher       = (*Matcher)(nil)
+	_ driver.MatcherConfigurable = (*Matcher)(nil)
+)
+
+const (
+	// Bounded concurrency limit.
+	defaultBatchSize          = 100
```
arajkumar, line 32: Reduce the batchSize to 10, having more than 10 would cause higher latency. Ideally the backend shouldn't allow batchSize > 10

crda/remotematcher.go
```diff
@@ -0,0 +1,404 @@
...
+	defaultEndPoint           = "/api/v2/vulnerability-analysis"
+	defaultRequestConcurrency = 10
+	defaultURL                = "https://f8a-analytics-2445582058137.production.gw.apicast.io/api/v2/"
```
arajkumar, line 35: Do you mean to hold just host? Adding endpoint with host defeats the purpose of defaultEndPoint

arajkumar (changes requested):

crda/remotematcher.go
rafiu007: @arajkumar the purpose here is to track users.Since this there is a possiblity that others can also get and use the other key.
A dedicated source is hence required.

crda/remotematcher.go
rafiu007: ACK.

crda/remotematcher.go
rafiu007: In the config the user can send across urlhost, key, source.
the end point is not going to change hence it is kept as not configurable .

crda/matcherfactory.go
```diff
@@ -0,0 +1,110 @@
...
+	zlog.Debug(ctx).Msg("loaded incoming config")
+
+	if fc.URL != "" {
+		u, err := url.Parse(fc.URL)
+		if err != nil {
+			return err
+		}
+		zlog.Info(ctx).
+			Str("url", u.String()).
+			Msg("configured manifest URL")
```
hdonnay, line 85:
```suggestion
			Msg("configured API URL")
```

crda/matcherfactory.go
```diff
@@ -0,0 +1,110 @@
...
+		f.url = u
+	}
+
+	if fc.Source != "" {
+		f.source = fc.Source
+		zlog.Info(ctx).
+			Str("source", fc.Source).
+			Msg("configured source of clair user")
```
hdonnay, line 93:
```suggestion
			Msg("configured source")
```

crda/matcherfactory.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if fc.Key != "" {
+		f.key = fc.Key
+		zlog.Info(ctx).
+			Str("key", f.key).
+			Msg("configured 3-scale-key for crda api")
```
hdonnay, line 100:
```suggestion
			Msg("configured API key")
```

crda/matcherfactory.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if c != nil {
+		zlog.Info(ctx).
+			Msg("configured HTTP client")
```
hdonnay, line 105:
```suggestion
```

The `http.Client` is passed unconditionally now, so this log message will be a lot of noise.

crda/remotematcher.go
```diff
@@ -0,0 +1,404 @@
...
+			Msg("configured source of clair user")
+	}
+	if cfg.Key != "" {
+		m.key = cfg.Key
+		zlog.Info(ctx).
+			Msg("configured 3-scale-key for crda api")
+	}
+	m.client = c
+	zlog.Info(ctx).
+		Msg("configured HTTP client")
```
hdonnay, line 278: Reuse/remove messages from above, as these all seem to be the same tunables.

crda/remotematcher.go
```diff
@@ -0,0 +1,404 @@
...
+
+// QueryRemoteMatcher implements driver.RemoteMatcher.
+func (m *Matcher) QueryRemoteMatcher(ctx context.Context, records []*claircore.IndexRecord) (map[string][]*claircore.Vulnerability, error) {
+	ctx = baggage.ContextWithValues(ctx,
+		label.String("component", "crda/Matcher.QueryRemoteMatcher"))
+	zlog.Debug(ctx).
+		Int("records", len(records)).
+		Msg("request")
+
+	// map Packge{name@version} to Packge to associate it with Vulnerability.
```
hdonnay, line 291:
```suggestion
	// Map Package{name@version} to Package to associate it with Vulnerability.
```

crda/remotematcher.go
```diff
@@ -0,0 +1,404 @@
...
+	for i, ir := range records {
+		request.Packages[i] = Package{
+			Name:    ir.Package.Name,
+			Version: ir.Package.Version,
+		}
+	}
+	// A request shouldn't go beyound 5s.
+	tctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	reqBody, _ := json.Marshal(request)
```
hdonnay, line 384: Check the error.

crda/remotematcher.go
```diff
@@ -0,0 +1,404 @@
...
+	req, err := http.NewRequestWithContext(tctx, http.MethodPost, m.url.String(), bytes.NewBuffer(reqBody))
+	req.Header.Set("User-Agent", "claircore/remote_matcher")
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Source-Type", defaultSource)
```
hdonnay, line 388: This seems to ignore the `Source` member?

crda/remotematcher.go
```diff
@@ -0,0 +1,404 @@
...
+
+	res, err := m.client.Do(req)
+	if res != nil {
+		defer res.Body.Close()
+	}
+	if err != nil {
+		return nil, err
+	}
```
hdonnay, line 396:
```suggestion
	if err != nil {
		return nil, err
	}
	defer res.Body.Close()
```

crda/remotematcher.go
```diff
@@ -0,0 +1,404 @@
...
+	var vulnReport []*VulnReport
+	data, _ := ioutil.ReadAll(res.Body)
+	err = json.Unmarshal(data, &vulnReport)
+	if err != nil {
+		return nil, err
+	}
```
hdonnay, line 402: Don't read the buffer the response without any limit. Better yet, use a `json.Decoder`.

crda/remotematcher_test.go
```diff
@@ -0,0 +1,245 @@
...
+				Name:    p.Name,
+				Version: p.Version,
+			}
+			testLocalPath := fmt.Sprintf("testdata/%s/%s/%s.json", vulnRequest.Ecosystem, p.Name, p.Version)
+			t.Logf("serving request for %v", testLocalPath)
+			jsonOut, err := ioutil.ReadFile(testLocalPath)
+			if err == nil {
+				err = json.Unmarshal(jsonOut, &res)
+				if err != nil {
+					t.Errorf("mock server unmarshall error %v", err)
```
hdonnay, line 81:
```suggestion
					t.Errorf("mock server unmarshal error %v", err)
```

crda/remotematcher_test.go
```diff
@@ -0,0 +1,245 @@
...
+				}
+			}
+			resp = append(resp, res)
+		}
+		out, err := json.Marshal(&resp)
+		if err != nil {
+			t.Errorf("mock server marshall error %v", err)
```
hdonnay, line 88:
```suggestion
			t.Errorf("mock server marshal error %v", err)
```

hdonnay (changes requested):
In addition to the code comments, this is missing needed documentation: the rate limit on the default configuration and how to request a unique API key.

The history also needs a cleanup, but that can be done after code changes and before submission.

crda/remotematcher.go
rafiu007: ACK.

crda/remotematcher.go
rafiu007: Will fix this.

crda/remotematcher.go
rafiu007: ACK

crda/remotematcher.go
rafiu007: used json.Decoder.

crda/remotematcher.go
rafiu007: Removed messages.

crda/remotematcher.go
```diff
@@ -0,0 +1,389 @@
...
+		Msg("response")
+	return results, nil
+}
+
+func (m *Matcher) invokeComponentAnalysesInBatch(ctx context.Context, records []*claircore.IndexRecord) (<-chan []*VulnReport, <-chan error) {
+	inC := make(chan []*claircore.IndexRecord, m.requestConcurrency)
+	ctrlC := make(chan []*VulnReport, m.requestConcurrency)
+	errorC := make(chan error, 1)
+	batchSize := m.batchSize
+	go func() {
```
crozzy, line 330: I may be wrong, but I think this could be unnecessarily complicated. This goroutine exists to allow the function to return before the work is done, but the very next line basically blocks until work is done:
```go
	ctrlC, errorC := m.invokeComponentAnalysesInBatch(ctx, records)
	err := <-errorC // guaranteed to have an err or be closed.
``` 
What would we lose from removing this goroutine and returning `([]*VulnReport, error)`? (To be clear, I still think the errgroup is good).

crda/remotematcher.go
```diff
@@ -0,0 +1,389 @@
...
+	} else {
+		m.key = "user_key=" + m.key
+	}
+
+	m.url = &url.URL{
+		Scheme:     m.url.Scheme,
+		Host:       m.url.Host,
+		Path:       defaultEndPoint,
+		ForceQuery: true,
+		RawQuery:   m.key,
```
crozzy, line 118: IMO the `m.key` variable shouldn't include the string`"user_key="`, it should be added here:
```go
RawQuery:   "user_key=" + m.key,
```

crda/remotematcher.go
```diff
@@ -0,0 +1,389 @@
...
+			Name:    ir.Package.Name,
+			Version: ir.Package.Version,
+		}
+	}
+	// A request shouldn't go beyound 5s.
+	tctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	reqBody, _ := json.Marshal(request)
+	req, err := http.NewRequestWithContext(tctx, http.MethodPost, m.url.String(), bytes.NewBuffer(reqBody))
+	req.Header.Set("User-Agent", "claircore/remote_matcher")
```
crozzy, line 374: I tend to agree with others that the source is essentially a `User-Agent`, but I get that the backend is expecting values like this, could you explain a little how the `key`, `source` and `user-agent` are checked on the backend, and what the limitations are keyed off of?

crozzy (changes requested):
Just a few comments, thanks!

crda/normalizeseverity.go
```diff
@@ -0,0 +1,28 @@
...
+	Low       = "low"
+	Medium    = "medium"
+	High      = "high"
+	Critical  = "critical"
+)
+
+// NormalizeSeverity takes a string[1] and normalizes it to
+// a claircore.Severity.
+// [1] https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/api_specs/v2/stack_analyses.yaml#L178
+func NormalizeSeverity(severity string) claircore.Severity {
```
crozzy, line 15: This function can be private I believe

crda/remotematcher.go
rafiu007: ACK.

crda/remotematcher.go
rafiu007: I believe the User-Agent field can be used to pass source.Will change accordingly.

crda/normalizeseverity.go
rafiu007: ACK.

crda/remotematcher.go
rafiu007: I think the idea is to make the api calls parallelly.Even thought it waits for the error.
I will @arajkumar To explain this better as he had implemented this. :)

crozzy: Gotcha and I agree, the API calls in parallel makes sense, I'm just suggesting that the synchronization takes place in the function rather than the caller needing to be responsible for that.

crda/remotematcher.go
crozzy: @rafiu007 like this I believe
```go
	reqBody, err := json.Marshal(request)
        if err != nil {
             // handle error
        }
	req, err := http.NewRequestWithContext(tctx, http.MethodPost, m.url.String(), bytes.NewBuffer(reqBody))
        if err != nil {
            // handle error
        }
```
(substitute the json.Marshal for the json.Decoder implementation)

rafiu007: Got it. My bad, I misunderstood it for a different part of the code. :)

rafiu007: Done.
Resolving this conversation.

crda/remotematcher.go
rafiu007: Resolved.

hdonnay:
I think this can be safely closed? Feel free to re-open if I don't understand what's going on between this is #423.

PR #423, opened 2021-07-30
crda: remote matcher with new crda api, source, and key

rafiu007:
- [x] Add back crda remote matcher
 - [x] Add new api
 - [x] Add unit tests
 - [x] Add configurable api-key
 - [x] Add configurable source 

The default api in crda remote matcher will serve 100rpm
The other key which can be optained by making an api call will be able to serve 1800 rpm to start with . We are positive about
increasing it even more.

Our api will have proper 3scale limits in place and we have removed the scenario which resulted in our system downtime.

Continued from https://github.com/quay/claircore/pull/419

### DIFF
--- a/crda/remotematcher.go
+++ b/crda/remotematcher.go
@@ -1,0 +1,361 @@
+package crda
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/quay/zlog"
+	"go.opentelemetry.io/otel/baggage"
+	"go.opentelemetry.io/otel/label"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+var (
+	_ driver.Matcher             = (*Matcher)(nil)
+	_ driver.RemoteMatcher       = (*Matcher)(nil)
+	_ driver.MatcherConfigurable = (*Matcher)(nil)
+)
+
+const (
+	// Bounded concurrency limit.
+	defaultBatchSize          = 100
+	defaultEndPoint           = "/api/v2/component-analyses"
+	defaultRequestConcurrency = 10
+	defaultURL                = "https://f8a-analytics-2445582058137.production.gw.apicast.io/?user_key=9e7da76708fe374d8c10fa752e72989f"
+)
+
+var (
+	supportedEcosystems = []string{"pypi", "maven"}
+)
+
+// Matcher attempts to correlate discovered python packages with reported
+// vulnerabilities.
+type Matcher struct {
+	batchSize          int
+	client             *http.Client
+	ecosystem          string
+	requestConcurrency int
+	url                *url.URL
+}
+
+// Build struct to model CRDA V2 ComponentAnalysis response which
+// delivers Snyk sourced Vulnerability information.
+type Vulnerability struct {
+	ID       string   `json:"id"`
+	CVSS     string   `json:"cvss"`
+	CVES     []string `json:"cve_ids"`
+	Severity string   `json:"severity"`
+	Title    string   `json:"title"`
+	URL      string   `json:"url"`
+	FixedIn  []string `json:"fixed_in"`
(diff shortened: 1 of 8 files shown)


crda/remotematcher.go
```diff
@@ -0,0 +1,386 @@
...
+			vulns, err := m.invokeComponentAnalyses(ctx, records[start:end])
+			if err != nil {
+				// log error
+				return nil
+			}
+			ctrlC <- vulns
+			return nil
+		})
+	}
+	g.Wait()
```
crozzy, line 343: Why don't we want to iterate through the channel here and return `[]*VulnReport`?

crda/remotematcher.go
```diff
@@ -0,0 +1,386 @@
...
+		return fmt.Sprintf("%s@%s", Name, Version)
+	}
+	for _, ir := range records {
+		packageVersionToIndexRecord[key(ir.Package.Name, ir.Package.Version)] = ir
+	}
+
+	results := make(map[string][]*claircore.Vulnerability)
+	ctrlC := m.invokeComponentAnalysesInBatch(ctx, records)
+
+	for vrs := range ctrlC {
```
crozzy, line 293: This outer for loop can be encapsulated inside the function `m.invokeComponentAnalysesInBatch()`, it means the channel's lifecycle is contained entirely within 1 function.

crozzy (commented):
I think you've more or less addressed all my comments, just left a couple of comments about returning the channel or not.

crda/remotematcher.go
rafiu007: @crozzy I did try doing that as per your reccomendation but the code erros out. Since ctrlC is still waiting for some api calls to return data its not possible to read fro ctrC and write into another variable.

crda/remotematcher.go
rafiu007: I agree but this was the issue why i chose to not https://github.com/quay/claircore/pull/423#discussion_r680910136

crda/remotematcher.go
crozzy: even after `g.Wait()` has been called? All goroutines should be done at that point, you should be able to close the ctrlC and iterate through the results. What was the error?

rafiu007: `libvulnhttp     | panic: close of closed channel` This was the error.It has been resolved and code updated

crda/remotematcher.go
```diff
@@ -333,15 +332,22 @@ func (m *Matcher) invokeComponentAnalysesInBatch(ctx context.Context, records []
...
 			}
 			ctrlC <- vulns
 			return nil
 		})
 	}
 	g.Wait()
-	return ctrlC
+	close(ctrlC)
+
+	// Not sure how big
```
crozzy, line 345: This comment not longer makes sense here, can you remove it?

rafiu007: yes.

crozzy (dismissed):
This LGTM

rafiu007:
@crozzy Can you please trigger the tests here too?

hdonnay:
@rafiu007 can you please look at this in light of the recent changes in the `java` package? The changes there may make this work differently, and I think you'll figure out if that's the case quicker than I will.

rafiu007:
> @rafiu007 can you please look at this in light of the recent changes in the `java` package? The changes there may make this work differently, and I think you'll figure out if that's the case quicker than I will.

sure @hdonnay I will try this with the changes.

rafiu007:
> @rafiu007 can you please look at this in light of the recent changes in the `java` package? The changes there may make this work differently, and I think you'll figure out if that's the case quicker than I will.

@hdonnay I tried out claircore with the changes. The only issue was with some dependencies , nothing that go mod tidy cant fix.
so ran go mod tidy and everything works. Were you expecting any other issues?

hdonnay:
Well, the `java` package has been changed to use a new, in-tree `jar` analyzer and as part of that, what it reports in the returned `claircore.Package` records may be different:

1. the `RespositoryHint` now has the sha1sum of the containing jar
2. the `PackageDB` now has a few pseudo-URLs it can return: `maven:`, `jar:`, and `file:`

hdonnay (approved):

PR #424, opened 2021-07-30
rpm: permissions madness

hdonnay:
The root of permission errors seems to be from directories not being
writable. This isn't an issue for root, since they're allowed to touch
files however they like no matter what (assuming no xattrs are at play
-- I really hope they're not).  However, when a normal user extracts a
tarball, they can take away their own write bit, which must be added to
remove any files (i.e. write the dirent).

This is compounded by the behavior of `chmod`, which is to operate on the target.
If the walk hasn't fixed the directory of the target yet, it will fail.

### DIFF
--- a/rpm/packagescanner.go
+++ b/rpm/packagescanner.go
@@ -164,11 +164,14 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
 			Stringer("stderr", &errbuf).
 			AnErr("err", err).
 			Msg("error extracting layer")
+		if rmerr := os.RemoveAll(root); rmerr != nil {
+			err = fmt.Errorf("%v (and during cleanup: %v)", err, rmerr)
+		}
 		return nil, fmt.Errorf("rpm: failed to untar: %w", err)
 	}
 	zlog.Debug(ctx).Str("dir", root).Msg("extracted layer")
 	// Immediately fix permissions.
-	if err := filepath.Walk(root, fixDirs(ctx)); err != nil {
+	if err := filepath.Walk(root, fixPerms(ctx, root)); err != nil {
 		if rmerr := os.RemoveAll(root); rmerr != nil {
 			err = fmt.Errorf("%v (and during cleanup: %v)", err, rmerr)
 		}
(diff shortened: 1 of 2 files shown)


rpm/packagescanner.go
```diff
@@ -161,78 +179,146 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
...
+	l := &extractedLayer{root: root}
+	// This takes the extreme measure of panicing the program if the layer
+	// directory is leaked or unsuccessfully cleaned up.
+	//
+	// Since these can use a nontrivial amount of disk space (layers around 1GiB
+	// are not uncommon) the eventual result is the same, but this way operators
+	// are more likely to notice.
+	_, file, line, _ := runtime.Caller(1)
+	runtime.SetFinalizer(l, func(l *extractedLayer) {
+		panic(fmt.Sprintf("%s:%d: layer directory %q not closed", file, line, root))
```
crozzy, line 235: This makes me a little nervous, I understand it's an issue that's probably going to affect the normal running of the application and probably lead to a full disk, but I'm thinking that should be discerned via an errors in the logs. We are sort of assuming an infrastructure by panicking, like we're assuming some kind of orchestration / storage setup.

I think the instance will error out until it's manually restarted with clean storage when the disk is full, but it could service other requests that aren't novel indexing requests. (which I'm not sure is better or worse from a diagnostic POV).

rpm/packagescanner.go
```diff
@@ -34,6 +36,14 @@ var dbnames = map[string]struct{}{
 	"Packages": {},
 }
 
+// Database kinds
+const (
+	_ int = iota
+	backendBDB
+	backendNDB
+	backendSqlite
+)
```
crozzy, line 45: Fine if not, but might be nice to have a dedicated type and then this var is more self-documenting:

```go
oldDBs := make(map[string]backend, 0)
```

rpm/packagescanner.go
```diff
@@ -161,78 +179,146 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
...
+		return err
+	}
+	return nil
 }
 
+func (l *extractedLayer) Close() error {
+	if err := os.RemoveAll(l.root); err != nil {
+		return fmt.Errorf("rpm: will panic soon: unable to remove directory: %w", err)
+	}
+	runtime.SetFinalizer(l, nil)
```
crozzy, line 315: wow, super interesting way of raising a panic if `Close()` hasn't been called

hdonnay: Yeah, I saw it here: https://crawshaw.io/blog/sharp-edged-finalizers

This use-case is a bit different, so maybe it's not appropriate, but between the additional flags telling tar to disregard permissions and ownership, setting permissions before doing work (instead of trying to do it just before cleanup), and the abstraction making the layer a bit easier to reason about, I think leaking this directory is either a programmer error or something seriously odd/missing in our handling.

rpm/packagescanner.go
hdonnay: The issue is that the old (bdb) backend is spread across multiple files. So the idea is to do a linear scan once of files that look like _any_ of the possible DB files and categorize them, then look at the ones that we're pretty sure are database and ignore extraneous files named `Packages` or the like.

hdonnay: I guess this patch is also carved out from some hacking to attempt to support sqlite databases (which wouldn't require extracting an entire layer) so I think this "type" concept is left over

crozzy (approved):

PR #426, opened 2021-08-05
rpm: eagerly change directory permissions

hdonnay:
This should error out earlier in the process and uses some extra tar
flags to make the permissions more amenable.


crozzy (approved):
LGTM

PR #435, opened 2021-08-10
java: check magic number

### LINKED ISSUE #434: failed to scan all layer contents: zip error: zip: not a valid zip file
I've seen this message in logs, and I'm not sure where it comes from.

Needs investigation.

hdonnay:
This change looks at the magic number if something appears to be a jar
based on the filename.

Closes #434

### DIFF
--- a/java/common.go
+++ b/java/common.go
@@ -2,7 +2,10 @@ package java
 
 import (
 	"archive/tar"
+	"bytes"
 	"context"
+	"errors"
+	"io"
 	"path"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


hdonnay:
does this need backports?

java/common.go
```diff
@@ -34,3 +37,29 @@ Outer:
...
+	// (ending in "\x05\x06" and "\x07\x08" respectively) for zips, but if I
+	// find those in use I'm going to delete this entire package and go flame
+	// some people on IRC.
+	zipHeader = []byte{'P', 'K', 0x03, 0x04}
+)
+
+// PeekHeader looks at the magic prefix of the current file pointed at by the
+// tar.Reader and returns either the entirety of the file as an io.Reader or an
+// error.
+func peekHeader(tr *tar.Reader) (io.Reader, error) {
```
crozzy, line 56: What is the point of returning the file as well as the error? Is there another case you're thinking about or is it because we need the contents anyway so might as well return them?

crozzy (dismissed):
LGTM, just one question

crozzy (dismissed):

hdonnay:
Merging via superpowers, because the commit hasn't changed.

Issue #436, opened 2021-08-12
rpm: remove tar


hdonnay:
The call to the external tar command has become more trouble than it's worth, mostly because there's no way to "clean" permissions, so we end up doing that anyway.

Some bug reports indicate that the rpm scanner is filling up disk space because it's reproducing perimssions that are, frankly, nonsense and preventing removal.

PR #437, opened 2021-08-12
rpm: remove tar(1) call

hdonnay:
This commit removes the external call to tar and replaces it with
creating files directly.

The original motivation for the tar call was that recreating the correct
permissions for everything is tricky. I now realize we most certainly do
*not* want to recreate the permissions of the filesystem, because
they're frequently bonkers and require extra work to clean up properly.

The simplified extraction routine implemented here makes sure that all
files remain readable and writable by the user creating them, so that
the Clair process can actually remove them. Any additional PRs to this
logic to support things like BSD suid/guid directory logic will send me
into my cups.

Closes: #436

### DIFF
--- a/rpm/packagescanner.go
+++ b/rpm/packagescanner.go
@@ -61,7 +61,7 @@ func (*Scanner) Kind() string { return pkgKind }
 //
 // A return of (nil, nil) is expected if there's no rpm database.
 //
-// The external commands "tar" and "rpm" are used and expected to be in PATH.
+// The external command "rpm" is used and expected to be in PATH.
 func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircore.Package, error) {
 	if err := ctx.Err(); err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


rpm/packagescanner.go
```diff
@@ -130,53 +130,125 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
...
-	// Unprivileged containers can't call mknod(2), so exclude /dev and
-	// hopefully there aren't any others strewn about.
-	tarcmd := exec.CommandContext(ctx, "tar",
-		"--extract",
-		"--exclude", "dev",
-		"--exclude", ".wh*",
-		"--delay-directory-restore",
-		"--no-same-owner",
-		"--no-same-permissions",
+	tr = tar.NewReader(rd)
```
crozzy, line 141: This looks good, the actual home-rolled extraction seems like something that might be useful in more places, should we consider pulling it out into it's own function?

crozzy (approved):

PR #447, opened 2021-08-17
rpm: normalize link target

hdonnay:
This fixes a bug where links would be created to unprefixed paths, which
would not exist and would fail at the link fixup step.

See-also: https://issues.redhat.com/browse/PROJQUAY-2403


jonathankingfc (approved):
LGTM

PR #455, opened 2021-08-26
ctxlock: context based locks

hdonnay:
This change adds a new lock API that uses a single database connection
to back the locks. The child contexts are canceled if the underlying
connection dies or if the parent is canceled.


pkg/ctxlock/watcher.go
```diff
@@ -0,0 +1,35 @@
...
+	return &watcher{
+		onCancel: onCancel,
+		done:     make(chan struct{}),
+	}
+}
+
+// Watch on the provided channel.
+func (w *watcher) Watch(ch <-chan struct{}) {
+	if ch == nil {
+		panic("nil channel")
```
crozzy, line 20: The panic here seems a little strong for an importable package. I get that it's going to be always called in a goroutine so returned errors aren't so useful, but if the channel is `nil` I have no problem with that just `return`ing.

pkg/ctxlock/watcher.go
```diff
@@ -0,0 +1,35 @@
+package ctxlock
+
+// A Watcher waits on two cancellation sources and makes sure to call the
+// wrapped function as soon as possible.
+type watcher struct {
```
crozzy, line 5: Is there a reason behind the private struct with public methods?

pkg/ctxlock/watcher.go
```diff
@@ -0,0 +1,35 @@
...
+	case <-ch:
+		w.onCancel()
+		<-w.done
+	case <-w.done:
+	}
+}
+
+// Unwatch tears down the watch. It should be called unconditionally.
+func (w *watcher) Unwatch() {
+	w.onCancel()
```
crozzy, line 33: Seems possible (likely) that `onCancel` will be called twice if I'm reading correctly, this should be documented if so

crozzy (commented):
Looks really good, I would always err on the side of the simpler `keyify()` function unless there is significant evidence of the allocations causing perf worries.

pkg/ctxlock/watcher.go
hdonnay: Yeah, this is implicitly depending on the behavior specified in the `context` docs:

> A CancelFunc tells an operation to abandon its work. A CancelFunc does not wait for the work to stop. A CancelFunc may be called by multiple goroutines simultaneously. After the first call, subsequent calls to a CancelFunc do nothing.

I'll add a note and double-check that the `unlock` call does the right thing.

pkg/ctxlock/watcher.go
hdonnay: My argument would be that this is a private function and it's a programmer error if this is called with a nil channel.

pkg/ctxlock/ctxlock.go
```diff
@@ -0,0 +1,354 @@
...
+// advisory lock if needed.
+func (l *Locker) unlock(ctx context.Context, key string, kb []byte, gen int, next context.CancelFunc) context.CancelFunc {
+	const query = `SELECT lock FROM pg_advisory_unlock($1) lock WHERE lock = true;`
+	return func() {
+		defer next()
+		l.rc.L.Lock()
+		defer l.rc.L.Unlock()
+
+		switch {
+		case gen < l.gen:
```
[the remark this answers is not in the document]
hdonnay: Ah, I suppose there is a chance that Close() has been called, but the `run` routine hasn't woken up yet and removed the connection. nice catch.

pkg/ctxlock/ctxlock.go
```diff
@@ -0,0 +1,354 @@
...
+func (l *Locker) ping() {
+	t := time.NewTicker(5 * time.Second)
+	defer t.Stop()
+	leave := false
+	for !leave {
+		<-t.C
+		l.rc.L.Lock()
+		leave = l.gen < 0
+		l.rc.L.Unlock()
+		l.rc.Broadcast()
```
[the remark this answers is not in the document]
hdonnay: I don't think it hurts, it just makes the loop simpler

pkg/ctxlock/watcher.go
hdonnay: not particularly

hdonnay:
the `keyify` allocations amounted to 2 more per call and about 500KiB in a 1 minute benchmark

pkg/ctxlock/watcher.go
crozzy: Sure, but it's fairly standard practice to de-reference a channel (as happens here in `reconnect()`) I'm not saying it's not safe, I think the cond lock probably covers it, just making the point that unlike writing to a closed channel or reading from a non-existent index, this is a situation where there are other options.

I'm happy for it to stay as is, the struct being private ties the logic up in the package.

crozzy (approved):

PR #456, opened 2021-08-26
dpkg: update identification method of dpkg DB dir

crozzy:
The current way to identify a dpkg DB can be too
exclusive in some scenarios and the "available" dir
does not always exist. This change checks the info
dir is there. This could result in some more false
positives given how generic the names are but this
is highly unlikely to lead to false positives due
to DB file format we expect.`

### DIFF
--- a/dpkg/scanner.go
+++ b/dpkg/scanner.go
@@ -27,7 +27,7 @@ import (
 const (
 	name    = "dpkg"
 	kind    = "package"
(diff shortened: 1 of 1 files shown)


hdonnay (approved):

hdonnay (changes requested):
Needs a bump to the version

hdonnay (approved):

PR #457, opened 2021-08-27
bug/livelock

hdonnay:
These are the fixes accumulated while hunting down a live lock.

Going per-commit is highly recommended when reviewing.


libvuln/updates/locks.go
```diff
@@ -2,108 +2,67 @@ package updates
...
+	sync.Mutex
+	wait *sync.Cond
+	m    map[string]struct{}
 }
 
 // LocalLockSource provides locks backed by local concurrency primitives.
-func LocalLockSource() LockSource {
-	return &localLockSource{
-		m: make(map[string]chan struct{}),
+func LocalLockSource() *localLockSource {
```
crozzy, line 17: Maybe we could take this opportunity to update the name to `NewLocalLockSource()`

pkg/ctxlock/ctxlock.go
```diff
@@ -0,0 +1,354 @@
...
+	"go.opentelemetry.io/otel/label"
+)
+
+// TODO(hank) Specify this algorithm to check its soundness.
+
+// New creates a Locker that will pull connections from the provided pool.
+//
+// The provided context is only used for logging and initial setup. Close must
+// be called to release held resources.
+func New(ctx context.Context, p *pgxpool.Pool) (*Locker, error) {
```
crozzy, line 31: More starting a discussion, does it make more sense for the `LockSource` interface definition to be defined here (and add the interface check) and imported? I suppose that would make the dir structure something like:
```
pkg/locksource/locksource.go // interface def
pkg/locksource/ctxlock.go // would need to update to `NewCtxLock`
````
```

crozzy (commented):
Just a couple of small comments, I agree 1.0.1 makes sense considering the public interface change.

internal/indexer/controller/controller.go
```diff
@@ -60,60 +52,78 @@ func New(opts *indexer.Opts) *Controller {
...
+			t := time.NewTimer(w)
+			select {
+			case <-ctx.Done():
+			case <-t.C:
+			}
+			t.Stop()
+			w = jitter()
+			retry = false
+			if e := ctx.Err(); !errors.Is(e, nil) {
+				err = nil // Fall out of the outer loop.
```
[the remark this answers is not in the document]
hdonnay: I think this is an old/odd comment, I'll think through and reword.

pkg/ctxlock/ctxlock.go
hdonnay: I think the structure of packages like this runs into a larger issue of structure with `libvuln` and `libindex` themselves: they don't actually use the store and lock abstractions, and other parts of the code do so inconsistently. I think for separating the interface out, we'd want to rework the main objects to be agnostic and use them, as well. Otherwise, it's all an internal detail and the indirection just gets annoying.

crozzy (approved):

PR #458, opened 2021-08-27
fetcher: use security data from access.redhat.com instead of www.redhat.com

siwinski:
The Red Hat security data owners, access.redhat.com owners, and www.redhat.com owners have coordinated data to be hosted via access.redhat.com instead of www.redhat.com.  There were concerns that merely redirecting all traffic from www.redhat.com to access.redhat.com would break certain tooling so the data is currently made available via both sites.  This pull request is to update Clair to pull data from access.redhat.com instead of www.redhat.com.

When we get to the point that there is no (or minimal) traffic on www.redhat.com we will implement a redirect so all `/security/data/*` traffic is redirected to access.redhat.com... do older versions of Clair support redirects?


siwinski:
I believe I have made all necessary commit updates.  Please let me know if anything needs to be adjusted.

crozzy (approved):

PR #478, opened 2021-09-28
rpm: prevent directory traversal when extracting

hdonnay:
A malicious tar archive with a member consisting of multiple "parent"
elements could cause a file outside of the target directory to be
overwritten.

Fixes: CVE-2021-3762


crozzy (approved):
LGTM

PR #490, opened 2021-10-13
jar: handle manifests with multiple sections

hdonnay:
I misread the jar manifest spec and expected sections to be delimited by
blank lines. On closer inspection, the spec says the sections are
delimited by the "Name" member itself.

This commit implements an io.Reader adapter that only reads the main
section.

### DIFF
--- a/java/jar/jar.go
+++ b/java/jar/jar.go
@@ -366,22 +366,18 @@ var errUnpopulated = errors.New("unpopulated")
 // This extracts "Main Attributes", as defined at
 // https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html.
 //
-// Also note that this spec gives and example that's invalid per their little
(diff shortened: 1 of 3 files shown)


java/jar/jar.go
```diff
@@ -425,6 +418,54 @@ func (i *Info) parseManifest(ctx context.Context, r io.Reader) error {
...
+func (m *mainSectionReader) Read(b []byte) (int, error) {
+	if m.Reader == nil {
+		return 0, io.EOF
+	}
+	var n int
+	l, err := m.ReadSlice('\n')
+	switch {
+	case errors.Is(err, nil):
+	case errors.Is(err, io.EOF):
+		// Hit EOF before a delimiter. Return EOF.
```
crozzy, line 446: I don't really understand this, are we actually going to return an EOF?

PR #501, opened 2021-10-28
Use go.opentelemetry.io/otel v1.1.0

misberner:
See https://github.com/quay/zlog/pull/1 for the motivation. Basically, the dependency on the outdated `otel` version with no stable API blocks upgrades of other libraries like `grpc`, for projects that depend on `claircore`.

The `replace` directive can be removed once https://github.com/quay/zlog/pull/1 is merged.

### DIFF
--- a/dpkg/testdata/texlive.status
+++ b/dpkg/testdata/texlive.status
@@ -7585,7 +7585,7 @@ Version: 2.31.1-0.4ubuntu3.6
 Depends: libc6 (>= 2.25), libuuid1 (>= 2.16)
 Description: block device ID library
  The blkid library allows system programs such as fsck and mount to
- quickly and easily find block devices by filesystem UUID or attribute.
+ quickly and easily find block devices by filesystem UUID or label.
  This allows system administrators to avoid specifying filesystems by
  hard-coded device names and use a logical naming system instead.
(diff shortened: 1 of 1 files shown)


internal/baggageutil/context.go
```diff
@@ -0,0 +1,37 @@
...
+// ContextWithValues adds a list of key value pairs to the context's baggage. kvPairs must be of even length.
+func ContextWithValues(ctx context.Context, kvPairs ...string) context.Context {
+	if len(kvPairs) == 0 {
+		return ctx
+	}
+
+	b := baggage.FromContext(ctx)
+	oldMembers := b.Members()
+	newMembers := make([]baggage.Member, 0, len(oldMembers)+len(kvPairs)/2)
+	newMembers = append(newMembers, oldMembers...)
```
hdonnay, line 19: What's the reasoning for doing this? Preventing a bunch of copies?

If this is used everywhere, it's probably worth being hoisted up into zlog itself.

misberner: Yeah, I wanted to avoid repeated `SetMember` calls, as that copies all existing members each time, leading to quadratic complexity.

hdonnay:
Closing in favor of #564 -- I added a `ContextWithValues` function in to `zlog` itself.

PR #502, opened 2021-11-01
Database ID

hdonnay:
These changes expose a database-specific ID, and then incorporates it into the `state` endpoint calculation.


hdonnay:
TODO before undrafting:

- [ ] check if a new database engine spin up results in the same database ID

hdonnay:
Looks like database oids are deterministic in a new engine.

PR #514, opened 2021-11-10
gc: avoid DB spikes during gc by adding lock

crozzy:
Currently it appear that the GC queries are causing some
pressure on the DB, this is compounded as there is nothing
to stop all matcher nodes from attempting GC at the same time.


crozzy:
image

crozzy:
tacking here: https://github.com/quay/claircore/pull/517

PR #517, opened 2021-11-12
GC: Do query execution all in one rather than separate

crozzy:
This change will delete the orphaned vulns, ie. those
that are no longer tied up the update_operations, in
one query rather than 3 seperate ones.

### DIFF
--- a/internal/vulnstore/postgres/gc.go
+++ b/internal/vulnstore/postgres/gc.go
@@ -10,10 +10,12 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/jackc/pgtype"
-	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/quay/zlog"
+	"go.opentelemetry.io/otel/baggage"
+	"go.opentelemetry.io/otel/label"
 	"golang.org/x/sync/semaphore"
 )
 
@@ -25,7 +27,7 @@ var (
 			Name:      "gc_total",
 			Help:      "Total number of database queries issued in the GC method.",
 		},
-		[]string{"query"},
+		[]string{"query", "success"},
 	)
 	gcDuration = promauto.NewHistogramVec(
 		prometheus.HistogramOpts{
@@ -38,22 +40,22 @@ var (
 	)
 )
 
+const (
+	// GCThrottle sets a limit for the number of deleted update operations
+	// (and subsequent cascade deletes in the uo_vuln table) that can occur in a GC run.
+	GCThrottle = 50
+)
+
 // GC is split into two phases, first it will identify any update operations
 // which are older then the provided keep value and delete these.
 //
-// Next it will perform chunked deletions of any vulns from the vuln table
+// Next it will perform updater based deletions of any vulns from the vuln table
 // which are not longer referenced by update operations.
 //
 // The GC is throttled to not overload the database with cascade deletes.
 // If a full GC is required run this method until the returned int64 value
 // is 0.
 func (s *Store) GC(ctx context.Context, keep int) (int64, error) {
-	const (
-		// GCThrottle sets a limit for the number of deleted update operations
-		// (and subsequent cascade deletes in the uo_vuln table) that can occur in a GC run.
-		GCThrottle = 50
-	)
-
 	// obtain update operations which need deletin'
 	ops, totalOps, err := eligibleUpdateOpts(ctx, s.pool, keep)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/vulnstore/postgres/gc.go
```diff
@@ -199,124 +199,38 @@ WHERE array_length(ordered_ops.refs, 1) > $2;
...
+	row := pool.QueryRow(ctx, getMaxVulnID)
+	err := row.Scan(&maxID)
+	if err != nil && !errors.Is(err, pgx.ErrNoRows) {
+		return err
+	}
 
-			return toDelete, nil
-		}()
+	// loop is terminated once the maximum ID in the DB is reached
+	for startID := uint64(0); startID >= maxID; startID = startID + 10000 {
```
kwestpharedhat, line 224: Can we think of any better ways to paginate to reduce the number of iterations required?

crozzy: Running locally it seems the query is actually move efficient without the bounds, the whole process was taking a prohibitively long time (20+mins), kicking it up to 1,000,000 was slightly faster, removing the bounds, the GC ran in around 2 mins.

crozzy: So there are 300+ updaters currently, which means queries are now 1 to 1 with # of updaters

internal/vulnstore/postgres/gc.go
```diff
@@ -197,126 +197,27 @@ WHERE array_length(ordered_ops.refs, 1) > $2;
...
-		if err != nil {
-			return err
-		}
+	start := time.Now()
+	zlog.Debug(ctx).
+		Str("updater", updater).
+		Msg("Starting clean up")
+	_, err := pool.Exec(ctx, deleteOrphanedVulns, updater)
+	if err != nil {
+		return fmt.Errorf("failed while exec'ing vuln delete: %v", err)
```
hdonnay, line 217:
```suggestion
		return fmt.Errorf("failed while exec'ing vuln delete: %w", err)
```

internal/vulnstore/postgres/gc.go
```diff
@@ -197,126 +197,27 @@ WHERE array_length(ordered_ops.refs, 1) > $2;
...
-			gcDuration.WithLabelValues("deletevuln_batch").Observe(time.Since(start).Seconds())
-			return nil
-		}()
-		if err != nil {
-			return err
-		}
+	start := time.Now()
+	zlog.Debug(ctx).
+		Str("updater", updater).
+		Msg("Starting clean up")
```
hdonnay, line 214:
```suggestion
		Msg("starting clean up")
```

internal/vulnstore/postgres/gc.go
```diff
@@ -197,126 +197,27 @@ WHERE array_length(ordered_ops.refs, 1) > $2;
...
+	_, err := pool.Exec(ctx, deleteOrphanedVulns, updater)
+	if err != nil {
+		return fmt.Errorf("failed while exec'ing vuln delete: %v", err)
 	}
+	gcCounter.WithLabelValues("deleteVulns").Add(1)
+	gcDuration.WithLabelValues("deleteVulns").Observe(time.Since(start).Seconds())
```
hdonnay, line 220:
```suggestion
	gcCounter.WithLabelValues("deleteOrphanedVulns").Add(1)
	gcDuration.WithLabelValues("deleteOrphanedVulns").Observe(time.Since(start).Seconds())
```

hdonnay: Should these counters get tickled before the error is checked? If the query fails, they don't count as written.

hdonnay (changes requested):
This approach looks much simpler and hopefully faster. Some q's around the error handling, though.

internal/vulnstore/postgres/gc.go
crozzy: I think probably not, if there's an error the latency is at best a useless metrics and at worse misleading. The count, should probably include another label that is success/failure (this would also be very useful for update operations).

crozzy:
```suggestion
	gcCounter.WithLabelValues("deleteVulns", "true").Inc()
	gcDuration.WithLabelValues("deleteVulns").Observe(time.Since(start).Seconds())
```

Then:
```go
	if err != nil {
		gcCounter.WithLabelValues("deleteVulns", "false").Inc()
		return fmt.Errorf("failed while exec'ing vuln delete: %w", err)
	}
```

crozzy:
```
11:21PM DBG lock failed error="ctxlock: lock acquisition failed" component=libvuln/updates/Manager.Run key=garbage-collection
11:21PM DBG lock context canceled, garbage collection already running error="context canceled" component=libvuln/updates/Manager.Run
```
GC lock working with multiple matchers running

hdonnay (approved):
SGTM

hdonnay (approved):

Issue #513, opened 2021-11-10
Use Red Hat Security Data API to scan Red Hat images


jasinner:
As of Nov 8th 2021 the accuracy of container image vulnerabilities in the Red Hat Security Data has [dramatically increased](https://access.redhat.com/articles/5554431) and we can now use this data to match Red Hat images released via registry.redhat.io to vulnerabilities in the data.

Take for example Red Hat's distribution of the Clair image, supported as part of Red Hat Quay. When a vulnerable version of the Clair image such as [v3.5.6-4](https://catalog.redhat.com/software/containers/quay/clair-rhel8/5efcb9ae7875a2070c71573d?tag=v3.5.6-4&push_date=1629392386000) is scanned it should be found vulnerable to [CVE-2021-3762](https://access.redhat.com/security/cve/cve-2021-3762).

We'll have to look for both the container "name", and "com.redhat.component' in the Red Hat Security Data as the container can be referred to by those names in the Red Hat Security Data. Both of those things are Labels in the Dockerfile:

```
$ skopeo inspect docker://registry.redhat.io/quay/quay-rhel8:v3.5.6-4 | jq '.Labels."com.redhat.component"'
"quay-registry-container"

$ skopeo inspect docker://registry.redhat.io/quay/quay-rhel8:v3.5.6-4 | jq '.Labels."name"'
"quay/quay-rhel8"
```

Get the version also from the "version" label:

```
$ skopeo inspect docker://registry.redhat.io/quay/quay-rhel8:v3.5.6-4 | jq '.Labels.version'
v3.5.6
```
In order to make sure we aware of the full name version revision, we should get the that also from the Labels:

```
$ skopeo inspect docker://registry.redhat.io/quay/quay-rhel8:v3.5.6-4 | jq '.Labels.release'
4
```

We can search the Red Hat Security Data using the [cvemap.xml](https://access.redhat.com/security/data/metrics/cvemap.xml), for example:

```
$ grep "quay/clair-rhel8:" -B 33 -A 3 cvemap.xml
...
<Vulnerability name="CVE-2021-3762">
...
    <AffectedRelease cpe="cpe:/a:redhat:quay:3" impact="important">
        <ProductName>Red Hat Quay 3</ProductName>
        <ReleaseDate>2021-09-28T00:00:00</ReleaseDate>
        <Advisory type="RHSA" url="https://access.redhat.com/errata/RHSA-2021:3665">RHSA-2021:3665</Advisory>
[...]

PR #518, opened 2021-11-14
add rhel container scanner

jasinner:
Add scanner for https://github.com/quay/claircore/issues/513

This adds both source, and binary package for the same container, but using the build name and container name respectively.

### DIFF
--- a/rhel/rhcc/updater.go
+++ b/rhel/rhcc/updater.go
@@ -1,0 +1,251 @@
+package rhcc
+
+import (
+	"context"
+	"encoding/xml"
+	"fmt"
+	"io"
+	"net/http"
+	"sort"
+
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+	"github.com/quay/claircore/pkg/cpe"
+	"github.com/quay/claircore/pkg/rhctag"
+	"github.com/quay/claircore/pkg/tmp"
+	"github.com/quay/claircore/rhel"
+)
+
+const (
+	dbURL = "https://access.redhat.com/security/data/metrics/cvemap.xml"
+)
+
+var (
+	_ driver.Updater      = (*updater)(nil)
+	_ driver.Configurable = (*updater)(nil)
+)
+
+// updater fetches and parses cvemap.xml
+type updater struct {
+	client *http.Client
+	url    string
+}
+
+type UpdaterConfig struct {
+	URL string `json:"url" yaml:"url"`
+}
+
+const updaterName = "rhel-container-updater"
+
+func (*updater) Name() string {
+	return updaterName
+}
+
+func UpdaterSet(_ context.Context) (driver.UpdaterSet, error) {
+	us := driver.NewUpdaterSet()
+	if err := us.Add(&updater{}); err != nil {
+		return us, err
+	}
+	return us, nil
+}
+
+func (u *updater) Configure(ctx context.Context, f driver.ConfigUnmarshaler, c *http.Client) error {
+	u.url = dbURL
+	u.client = c
+	var cfg UpdaterConfig
+	if err := f(&cfg); err != nil {
+		return err
+	}
+	if cfg.URL != "" {
+		u.url = cfg.URL
+	}
+	return nil
+}
+
+func (u *updater) Fetch(ctx context.Context, hint driver.Fingerprint) (io.ReadCloser, driver.Fingerprint, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "rhel/rhcc/Updater.Fetch")
+
+	zlog.Info(ctx).Str("database", u.url).Msg("starting fetch")
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.url, nil)
+	if err != nil {
+		return nil, hint, fmt.Errorf("rhcc: unable to construct request: %w", err)
+	}
+
+	if hint != "" {
+		zlog.Debug(ctx).
+			Str("hint", string(hint)).
+			Msg("using hint")
+		req.Header.Set("if-none-match", string(hint))
+	}
+
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, hint, fmt.Errorf("rhcc: error making request: %w", err)
+	}
+	defer res.Body.Close()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		if t := string(hint); t == "" || t != res.Header.Get("etag") {
+			break
+		}
+		fallthrough
+	case http.StatusNotModified:
+		zlog.Info(ctx).Msg("database unchanged since last fetch")
+		return nil, hint, driver.Unchanged
+	default:
+		return nil, hint, fmt.Errorf("rhcc: http response error: %s %d", res.Status, res.StatusCode)
+	}
+	zlog.Debug(ctx).Msg("successfully requested database")
+
+	tf, err := tmp.NewFile("", updaterName+".")
+	if err != nil {
+		return nil, hint, fmt.Errorf("rhcc: unable to open tempfile: %w", err)
+	}
+	zlog.Debug(ctx).
+		Str("name", tf.Name()).
+		Msg("created tempfile")
+
+	var r io.Reader = res.Body
+	if _, err := io.Copy(tf, r); err != nil {
+		tf.Close()
+		return nil, hint, fmt.Errorf("rhcc: unable to copy resp body to tempfile: %w", err)
+	}
+	if n, err := tf.Seek(0, io.SeekStart); err != nil || n != 0 {
+		tf.Close()
+		return nil, hint, fmt.Errorf("rhcc: unable to seek database to start: %w", err)
+	}
+	zlog.Debug(ctx).Msg("decompressed and buffered database")
+
+	hint = driver.Fingerprint(res.Header.Get("etag"))
+	zlog.Debug(ctx).
+		Str("hint", string(hint)).
+		Msg("using new hint")
+
+	return tf, hint, nil
+}
+
+func (u *updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vulnerability, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "rhel/rhcc/Updater.Parse")
+	zlog.Info(ctx).Msg("parse start")
+	defer r.Close()
+	defer zlog.Info(ctx).Msg("parse done")
+
+	var cvemap cveMap
+	if err := xml.NewDecoder(r).Decode(&cvemap); err != nil {
+		return nil, fmt.Errorf("rhel: unable to decode cvemap: %w", err)
+	}
+	zlog.Debug(ctx).Msg("xml decoded")
+
+	zlog.Debug(ctx).
+		Int("count", len(cvemap.RedHatVulnerabilities)).Msg("found raw entries")
+
+	vs := []*claircore.Vulnerability{}
+	for _, vuln := range cvemap.RedHatVulnerabilities {
+		description := getDescription(vuln.Details)
+		versionsByContainer := make(map[string]map[rhctag.Version]*consolidatedRelease)
+		for _, release := range vuln.AffectedReleases {
+			match, packageName, version := parseContainerPackage(release.Package)
+			if !match {
+				continue
+			}
+			// parse version
+			v, err := rhctag.Parse(version)
+			if err != nil {
+				zlog.Debug(ctx).
+					Str("package", packageName).
+					Str("version", version).
+					Err(err).
+					Msgf("tag parse error")
+				continue
+			}
+			// parse severity
+			var severity string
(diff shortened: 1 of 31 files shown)


jasinner:
@hdonnay This is getting close to being complete. Do you think you could give it look over and make sure it's on the right track?

rhel/containers/matcher.go
```diff
@@ -0,0 +1,61 @@
+package containers
+
+import (
+	"context"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+	"golang.org/x/mod/semver"
```
hdonnay, line 8: Do Red Hat containers actually use semver and not RPM versions?

hdonnay (commented):
Overall, I think this should just be in the `rhel` package and wired up as and additional Updater & Matcher.

rhel/containers/parser.go
```diff
@@ -0,0 +1,160 @@
...
+				Description:        description,
+				Issued:             release.ReleaseDate.time,
+				Severity:           severity,
+				NormalizedSeverity: rhel.NormalizeSeverity(severity),
+				Package:            pkg,
+				FixedInVersion:     version,
+				Dist: &claircore.Distribution{
+					DID:  ID,
+					Name: Name,
+					CPE:  cpe,
```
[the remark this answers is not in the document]
jasinner: The CPE is not being used at the moment in the matcher because we don't have the CPE in the container Dockerfile or similar. Eventually it should be available via Pyxis, so I decided it was better to record it somewhere now.

What we have to filter in the matcher is Distribution ID and Name because all the data relates to Red Hat containers which are distributed on Red Hat containers, with very few exceptions, if any. Therefore I put CPE with Distribution for now. 

When we are ready to actually use CPE in the matcher should I added it as a repository instead? I think we'd only have one repository, we could call it something like registry.access.redhat.com, but each product or "application" would have a unique CPE. However, I didn't see any libvuln.driver.matcher.MatchConstraint for RepositoryCpe.

jasinner:
I originally tried that approach, however I ran into issues. I think the Ecosystem objects was causing issue with overlapping the with rhel (rpm) code. Maybe we can move that instead into a subclass of rhel called "rpm"?

rhel/containers/updater.go
```diff
@@ -0,0 +1,50 @@
...
+const (
+	dbURL = "https://access.redhat.com/security/data/metrics/cvemap.xml"
+)
+
+var (
+	_ driver.Updater = (*Updater)(nil)
+)
+
+// Updater fetches and parses cvemap.xml
+type Updater struct {
```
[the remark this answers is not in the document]
jasinner: Ok, I'll add this, thanks.

rhel/containers/parser.go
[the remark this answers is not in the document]
jasinner: Actually, I think a better approach might be to use the container repository 'repository', as the repository. For example for the container image `quay/clair-rhel8:v3.5.7-8`, the repository would be `quay`, and the container name becomes `clair-rhel8`. This solves a corner case using CPEs where multiple minor versions all ship a container in the same container 'repository', but have their fixes in different versions. For example in CVE-2021-44228 section of the cvemap.xml we have:

```
    <AffectedRelease cpe="cpe:/a:redhat:logging:5" impact="moderate">
        <ProductName>OpenShift Logging 5.0</ProductName>
        <ReleaseDate>2021-12-14T00:00:00</ReleaseDate>
        <Advisory type="RHSA" url="https://access.redhat.com/errata/RHSA-2021:5137">RHSA-2021:5137</Advisory>
        <Package name="openshift-logging/elasticsearch6-rhel8">openshift-logging/elasticsearch6-rhel8:v5.0.10-1</Package>
    </AffectedRelease>
    <AffectedRelease cpe="cpe:/a:redhat:logging:5.1::el8">
        <ProductName>OpenShift Logging 5.1</ProductName>
        <ReleaseDate>2021-12-14T00:00:00</ReleaseDate>
        <Advisory type="RHSA" url="https://access.redhat.com/errata/RHSA-2021:5128">RHSA-2021:5128</Advisory>
        <Package name="openshift-logging/elasticsearch6-rhel8">openshift-logging/elasticsearch6-rhel8:v6.8.1-67</Package>
    </AffectedRelease>
    <AffectedRelease cpe="cpe:/a:redhat:logging:5.2::el8">
        <ProductName>OpenShift Logging 5.2</ProductName>
        <ReleaseDate>2021-12-14T00:00:00</ReleaseDate>
        <Advisory type="RHSA" url="https://access.redhat.com/errata/RHSA-2021:5127">RHSA-2021:5127</Advisory>
        <Package name="openshift-logging/elasticsearch6-rhel8">openshift-logging/elasticsearch6-rhel8:v6.8.1-66</Package>
    </AffectedRelease>
    <AffectedRelease cpe="cpe:/a:redhat:logging:5.3::el8">
        <ProductName>OpenShift Logging 5.3</ProductName>
        <ReleaseDate>2021-12-14T00:00:00</ReleaseDate>
        <Advisory type="RHSA" url="https://access.redhat.com/errata/RHSA-2021:5129">RHSA-2021:5129</Advisory>
        <Package name="openshift-logging/elasticsearch6-rhel8">openshift-logging/elasticsearch6-rhel8:v6.8.1-65</Package>
    </AffectedRelease>
```

We can parse this as 2 different vulnerabilities with unique ranges:

- v5.0.0 - v5.0.10-1
- v6.8.0 - v6.8.1-65

Whereas, if we use CPE (which are not available in Dockerfile yet) we'd end up with 4 vulnerabilities, 3 of which would match against a container such as `openshift-logging/elasticsearch6-rhel8:v6.8.1-30`.

rhel/containers/matcher.go
jasinner: Well, each tag begins with a "v", and all tags I've looked at comply with Semantic Version rules. However I haven't actually read anywhere that Red Hat containers, or container registry tags in general do use strict Semantic Versioning.

I'm fleshing this out further in a following commit, which creates vulnerabilities with affects ranges for each minor version. I'm creating a custom type for handing them, but using x/mod/semver still for parsing the tags.

jasinner: I just noticed that the UBI container doesn't use semantic versioning. Luckily that container doesn't have any container-first content, as it's comprised of RPMs only.

I'm going to stick with rhctag.Version for now, and just skip containers tag which we can't parse as Semantic Versions. We can revisit this if it becomes a problem.

jasinner:
> Overall, I think this should just be in the `rhel` package and wired up as and additional Updater & Matcher.

Why do you think they should be in the rhel package @hdonnay ?

If we did that, you wouldn't be able to disable the rhel/container update or matcher independently of the rhel/rpm or original one. I think that will be necessary, especially for early adopters.

rhel/containers/matcher.go
jasinner: BTW, I removed x/mod/semver import from this package. It's only in the pkg/rhctag package now.

jasinner:
I just noticed a bug matching vulnerabilities in containers such as 'openshift4/ose-metering-hive'. The Dockerfile label for those images set's the name as 'openshift/ose-metering-hive' (without the 4). Therefore the package names don't match the vulnerabilities in that container repo. Moving this back to draft until it's resolved.

I think there are a few ways we could fix it:

1. Use Pyxis and the container NVR to lookup the container external repo name in Claircore
2. Include the NVR in the cvemap.xml for containers.

I'm leaning towards the first option because it doesn't involve changes to the application producing cvemap.xml.

jasinner:
> I just noticed a bug matching vulnerabilities in containers such as 'openshift4/ose-metering-hive'. The Dockerfile label for those images set's the name as 'openshift/ose-metering-hive' (without the 4). Therefore the package names don't match the vulnerabilities in that container repo. Moving this back to draft until it's resolved.
> 
> I think there are a few ways we could fix it:
> 
>     1. Use Pyxis and the container NVR to lookup the container external repo name in Claircore
> 
>     2. Include the NVR in the cvemap.xml for containers.
> 
> 
> I'm leaning towards the first option because it doesn't involve changes to the application producing cvemap.xml.

Solved it using the first option.

rhel/rhcc/parser.go
```diff
@@ -0,0 +1,224 @@
...
+				Issued:   release.ReleaseDate.time,
+				Severity: severity,
+			}
+			// initialize and update the fixed in versions slice
+			if versionsByContainer[packageName][minorKey].FixedInVersions == nil {
+				vs := make(rhctag.Versions, 0)
+				versionsByContainer[packageName][minorKey].FixedInVersions = &vs
+			}
+			newVersions := versionsByContainer[packageName][minorKey].FixedInVersions.Append(v)
+			versionsByContainer[packageName][minorKey].FixedInVersions = &newVersions
```
hdonnay, line 166: This slice can only ever contain a single value, is that right?

rhel/rhcc/parser.go
```diff
@@ -0,0 +1,224 @@
...
+					NormalizedSeverity: rhel.NormalizeSeverity(releasesByMinor[minor].Severity),
+					Package:            p,
+					Dist: &claircore.Distribution{
+						DID:  ID,
+						Name: Name,
+						// TODO for containers such as openshift-logging/elasticearch6-rhel8
+						// This matches the CPE of the last entry. However we would probably
+						// need a vulnerability with a unique CPE for each once we start
+						// filtering by CPE from container scanner.
+						CPE: releasesByMinor[minor].Cpe,
```
hdonnay, line 212: This is in the wrong spot, this is the application's CPE.

jasinner: I took a look at the security data, and I found we only publish multiple container image tags with multiple CPEs in 0.5% of cases. Therefore I removed CPE from here.

As I mentioned [ealier](https://github.com/quay/claircore/pull/518#discussion_r783625460), I wasn't sure where I should store the CPE to be used in a matcher, so I put it in distribution. If we need to start using CPE in future, we'll have to figure out where to store it then.

rhel/rhcc/parser.go
jasinner: No, it can contain multiple fixed-in versions for the same minorKey. Using this example from the [testdata](https://github.com/quay/claircore/blob/4a612eb71afc054953b1c6b4fd363d4fa8024a9a/rhel/rhcc/testdata/cve-2021-44228-openshift-logging.xml#L75-L87) the slice will contain all of:

- v6.8.1-65
- v6.8.1-66
- v6.8.1-67

The [loop after this](https://github.com/quay/claircore/blob/4a612eb71afc054953b1c6b4fd363d4fa8024a9a/rhel/rhcc/parser.go#L192) use the lowest value as the value for fixed-in' for that minorKey.

This is one case where a container "openshift-logging/elasticsearch6-rhel8" is shipping multiple CPEs with the same minor version tags. You can see in the [testdata](https://github.com/quay/claircore/blob/4a612eb71afc054953b1c6b4fd363d4fa8024a9a/rhel/rhcc/testdata/cve-2021-44228-openshift-logging.xml#L75-L87) there are CPEs for all of these applications all pointing to the 6.8.1 version of this container.

- cpe:/a:redhat:logging:5.1::el8
- cpe:/a:redhat:logging:5.2::el8
- cpe:/a:redhat:logging:5.3::el8

We take the lowest version rather than the highest to avoid false positives for the 65, and 67 revisions.

jasinner: Also in the second loop, we are able to determine the start value for all the fixedIn versions by sorting them. At the start value we always use a claircore.Version = 0 (Type: "rhctag") as the minor version in the Range. That allows us to leverage a VersionFilter in the matcher.

jasinner:
I made some improvement to this branch. We can now parse every affected container image release version from cvempa.xml with this new code. It no longer uses the semver library, but instead does some basic checks of the leading characters in the version string in order to create minor version ranges. Those ranges are used to filter issues using a versioned matcher.

Finally once packages are matches to the correct minor version range we use the existing rpm version comparison library we already use to compare packages in the rpm library to determine if the container image package in the range is vulnerable or not.

rhel/rhcc/updater.go
```diff
@@ -0,0 +1,248 @@
...
+
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, hint, fmt.Errorf("rhcc: error making request: %w", err)
+	}
+	defer res.Body.Close()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		// break
```
crozzy, line 91: We changed these all recently to still check on 200, should we also change this or are we confident that https://access.redhat.com/security/data/metrics/cvemap.xml is going to return correctly?

crozzy (approved):

hdonnay:
@jasinner please give this a once-over, I did quite a bit of code-touching.

crozzy (dismissed):

crozzy (approved):
Makes sense to me

jasinner:
> @jasinner please give this a once-over, I did quite a bit of code-touching.

Looks good to me! Thanks

PR #520, opened 2021-11-17
jar: handle archive without filesystem semantics

hdonnay:
Ivan found a jar in the wild that doesn't have filesystem semantics.
Instead of entries `a/` and `a/b`, it contains `a` and `a/b`. This
tricks the zip package into synthesizing a file entry that cannot be
opened.

A long-term solution would be to ditch the use of the fs package inside
this package, but this does make testing a bit more complicated.

### DIFF
--- a/java/jar/jar.go
+++ b/java/jar/jar.go
@@ -37,10 +37,10 @@ import (
 	"io"
 	"io/fs"
 	"net/mail"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
-	"sync"
 
 	"github.com/quay/zlog"
 	"go.opentelemetry.io/otel/baggage"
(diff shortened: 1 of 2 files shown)


hdonnay:
@crozzy @ibazulic This implements as fix by detecting and doing best-effort coping with these archives. There's another way to fix this (outlined in the BUG comment) that would be more involved, but more correct.

crozzy:
So is the thought that these faux-directory files might in fact be legitimate, and therefore, deserve to be parsed as if they are in a correct directory structure?

hdonnay:
It's more so that zip files _technically_ don't follow filesystem semantics -- they're more like "S3 in a can" -- so _technically_ anything that wants to handle an arbitrary zip file needs to cope with that. A valid jar is any valid zip that the jvm can make sense of and there's nothing that *requires* filesystem semantics; the fact that basically every other user of the format treats it like a compressed filesystem be damned.

I used the fs.FS interface inside this package to make things quicker to write and test, but it's now looking like that a proper solution would deal with the zip files concretely in order to cope with these corner cases where the jvm accepts a zip that everything else thinks is bonkers.

hdonnay:
Just rewrote this to not use the FS interface, actually not that bad.

crozzy (approved):

PR #521, opened 2021-11-18
indexer: manifest deletion

hdonnay:
These changes modify the database to add `CASCADE` constraints on some foreign keys and exposes + implements API surface to remove manifests from the underlying indexer data store. This is needed to prevent the indexer's database from growing unbounded.

Along the way there were some refactors, clean ups, and helpers.


libindex/migrations/04-foreign-key-cascades.sql
```diff
@@ -0,0 +1,165 @@
+-- Add CASADEs on all foreign-key relations.
```
crozzy, line 1: Do we want to lock the tables? I'm trying to think of someone is cycling a bunch of indexer pods with new versions while the old ones are still working.

libindex/migrations/migrations.go
```diff
@@ -2,34 +2,44 @@ package migrations
...
 	"github.com/remind101/migrate"
 )
 
-const (
-	MigrationTable = "libindex_migrations"
-)
+//go:embed *.sql
+var fs embed.FS
+
+func runFile(n string) func(*sql.Tx) error {
```
crozzy, line 13: Much nicer

crozzy (commented):
A couple of comments, but overall LGTM

libindex/migrations/04-foreign-key-cascades.sql
hdonnay: according to the docs, the PostgresQL migrator takes an advisory lock

crozzy (approved):

PR #527, opened 2021-11-30
dockerfile: correctly parse env declarations that span multiple lines

### LINKED ISSUE #526: RHEL dockerfile parser can panic parsing env vars
We've been seeing a lot of panics from the RHEL dockerfile parser which was recently replaced.

For example:
```
panic: runtime error: slice bounds out of range [:-1]

goroutine 86265 [running]:
github.com/quay/claircore/rhel/dockerfile.(*labelParser).Run(0xc00e8efef0, 0x1108f20, 0xc00ecbcab0)
	/build/vendor/github.com/quay/claircore/rhel/dockerfile/dockerfile.go:99 +0xe6a
github.com/quay/claircore/rhel/dockerfile.GetLabels(0x1121130, 0xc00e273140, 0x1108f20, 0xc00ecbcab0, 0x0, 0x0, 0x0)
	/build/vendor/github.com/quay/claircore/rhel/dockerfile/dockerfile.go:26 +0x93
github.com/quay/claircore/rhel.extractBuildNVR(0x1121130, 0xc00e273140, 0xc004c94c60, 0x2d, 0xc00ecbcab0, 0x0, 0x0, 0x0, 0x0, 0xc00e273150, ...)
	/build/vendor/github.com/quay/claircore/rhel/repositoryscanner.go:315 +0x66
github.com/quay/claircore/rhel.(*RepositoryScanner).getCPEsUsingContainerAPI(0xc003600ba0, 0x1121130, 0xc00e273140, 0xc0036008a0, 0xc00e273140, 0xc00804a6d0, 0x0, 0x0, 0x0)
	/build/vendor/github.com/quay/claircore/rhel/repositoryscanner.go:248 +0x14f
github.com/quay/claircore/rhel.(*RepositoryScanner).Scan(0xc003600ba0, 0x1121168, 0xc00e781c20, 0xc0036008a0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/build/vendor/github.com/quay/claircore/rhel/repositoryscanner.go:187 +0xafb
github.com/quay/claircore/internal/indexer/layerscanner.(*result).Do(0xc009767e58, 0x1121168, 0xc0084e6570, 0x7f2db434ef98, 0xc003600ba0, 0xc0036008a0, 0x20, 0x20)
	/build/vendor/github.com/quay/claircore/internal/indexer/layerscanner/layerscanner.go:241 +0x299
github.com/quay/claircore/internal/indexer/layerscanner.(*layerScanner).scanLayer(0xc003457020, 0x11210c0, 0xc0084e6570, 0xc0036008a0, 0x7f2db434ef98, 0xc003600ba0, 0x0, 0x0)
	/build/vendor/github.com/quay/claircore/internal/indexer/layerscanner/layerscanner.go:212 +0x85c
github.com/quay/claircore/internal/indexer/layerscanner.(*layerScanner).Scan.func1.1(0x0, 0x0)
	/build/vendor/github.com/quay/claircore/internal/indexer/layerscanner/layerscanner.go:173 +0x147
golang.org/x/sync/errgroup.(*Group).Go.func1(0xc000c46a80, 0xc0011d5240)
	/build/vendor/golang.org/x/sync/errgroup/errgroup.go:57 +0x59
created by golang.org/x/sync/errgroup.(*Group).Go
	/build/vendor/golang.org/x/sync/errgroup/errgroup.go:54 +0x66
```

I haven't managed to track down an example of a dockerfile that is causing this, which might help us solve this properly, but will share one if I manage to.  
When having a play with the code I seem to be able to trip it up when using a multi line env declaration, so I assume it is something along those lines, e.g. 
```
ENV MY_NAME="John Doe" \
    MY_CAT=fluffy
```
This seems to end up split into 3 'pairs': `MY_NAME="John Doe"`, `\`, and `MY_CAT=fluffy`
Then the parser would panic because the 2nd pair doesn’t contain an equals character. 

I’m working on a fix and submit a PR. I also think it would be a good to add a check for "-1" response before using it as an index anyway, and return an error, so I’ll include that in the PR. That will avoid any panics and hopefully provide a more helpful error in the case where we’ve missed something else.

paulaldridge:
the dockerfile parser would treat new line escape as an extra key
value pair, then panic when it did not contain an equals character

Fixes #526

Note: I've made it so that the new line escape and following whitespace is trimmed as that generates the pair as expected - using the first test I added as an example, it finds the 2nd pair as `"k=v"` rather than `"\\\n\t\t\t\tk=v"`

### DIFF
--- a/rhel/dockerfile/dockerfile.go
+++ b/rhel/dockerfile/dockerfile.go
@@ -96,6 +96,9 @@ func (p *labelParser) Run() error {
 				}
 				for _, kv := range pairs {
 					i := strings.IndexByte(kv, '=')
+					if i == -1 {
+						return fmt.Errorf("unrecognised format of env key value pair in dockerfile: %s", kv)
+					}
 					k, _, err := transform.String(p.unquote, kv[:i])
 					if err != nil {
 						return err
@@ -160,13 +163,23 @@ func (p *labelParser) Run() error {
 // Label and Env instructions allow for key-value pairs with this syntax.
 func splitKV(escchar rune, in string) ([]string, error) {
 	var ret []string
-	var esc, quote bool
+	var esc, quote, escapedNewLine bool
 	var quotechar rune
 	start := 0
 	for cur, r := range in {
 		switch {
 		case esc:
 			esc = false
+			if r == '\n' {
+				escapedNewLine = true
+			}
+		case escapedNewLine && unicode.IsSpace(r): // advance
+		case escapedNewLine:
+			escapedNewLine = false
+			start = start + 1
+			if !esc && r == escchar {
+				esc = true
+			}
 		case !esc && r == escchar:
 			esc = true
 		case !esc && !quote && (r == '"' || r == '\''):
@@ -179,7 +192,7 @@ func splitKV(escchar rune, in string) ([]string, error) {
 			runlen := cur - start
 			switch {
 			case runlen > 1:
-				ret = append(ret, in[start:cur])
+				ret = append(ret, strings.TrimSpace(in[start:cur]))
 				fallthrough
 			case runlen == 1:
 				start = cur + 1
@@ -189,7 +202,7 @@ func splitKV(escchar rune, in string) ([]string, error) {
 		}
 	}
 	if rem := in[start:]; len(rem) > 0 {
-		ret = append(ret, rem)
+		ret = append(ret, strings.TrimSpace(rem))
 	}
 	return ret, nil
 }
--- a/rhel/dockerfile/dockerfile_test.go
+++ b/rhel/dockerfile/dockerfile_test.go
@@ -88,6 +88,22 @@ func TestSplit(t *testing.T) {
 			In: "k=' v '	\v k=\"   \"",
 			Want: []string{`k=' v '`, `k="   "`},
 		},
+		{
+			In: `k="v" \
+				k=v`,
+			Want: []string{`k="v"`, `k=v`},
+		},
+		{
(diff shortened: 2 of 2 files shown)


hdonnay:
This should be handled by the lexer's `collectLine` method already. There are dockerfiles in the testdata directory that have newlines between and inside ENV/ARG instructions, how are they being missed?

paulaldridge:
Ah yes I see sorry, thanks for explaining! Those dockerfile tests do show the multi line envs aren’t causing the panic. Because they weren’t checking for env values in the test output I thought they weren’t covering it. I must have misunderstood the collectLine function, it looked like it was collecting the whole env line including any escaped new lines, and therefore end up with the splitKV function having to process a line like this test: https://github.com/paulaldridge/claircore/blob/dockerfile-parser-panics/rhel/dockerfile/dockerfile_test.go#L92, which didn’t seem to work. But it must be removing those new lines before that point, but I’m still not sure I can see how that is done (any chance you could help me understand how that happens for my understanding/curiosity?).  

Do you agree it’s still worth having the check for `-1` output from `IndexByte()`, as that is an expected response from that function, before using the output in the slice index? i.e. https://github.com/quay/claircore/pull/527/files#diff-791df6b1ddb182a2348e119184cddfca9d696bc9e551e206be560f53f67ff4b7L98
With a valid dockerfile I don't think we'd ever get a -1, but it could catch any other potential parser bugs.

hdonnay:
Closing in favor of #536

PR #530, opened 2021-12-06
pep440: log the invalid version string

crozzy:
When debugging it would be nice to see the strings that are
choking during the version parsing to be able to add them to
regressions.


crozzy:
quay.io/crozzy/test-py:bad-version2 Test image

pkg/pep440/version.go
```diff
@@ -131,7 +130,7 @@ func (a *Version) Compare(b *Version) int {
 // Parse attempts to extract a PEP-440 version string from the provided string.
 func Parse(s string) (v Version, err error) {
 	if !pattern.MatchString(s) {
-		return v, errors.New("no version found")
+		return v, fmt.Errorf("invalid pep440 version %s", s)
```
hdonnay, line 133:
```suggestion
		return v, fmt.Errorf("invalid pep440 version: %q", s)
```

hdonnay (dismissed):
LGTM

I'd reformat the error string, but that's not a blocker.

hdonnay (approved):

PR #531, opened 2021-12-06
bug: xml decoding

hdonnay:
Debian recently added an `encoding` attribute to the XML text element in their OVAL feeds, which causes the stdlib xml decoder to call the `CharsetReader` member. We should populate this for robustness' sake.


crozzy (dismissed):
LGTM, just to be clear, this wouldn't fix the debian issue until they add valid encoding string(say "us-ascii") right?

crozzy (approved):

PR #532, opened 2021-12-06
rhel: fix error on old containers

hdonnay:
This change causes the repository scanner to fall back to the Container
API when a Dockerfile is present but missing expected labels, not just
when a Dockerfile is missing.

Closes: PROJQUAY-2868

### DIFF
--- a/rhel/repositoryscanner.go
+++ b/rhel/repositoryscanner.go
@@ -246,11 +246,15 @@ func (r *RepositoryScanner) getCPEsUsingContainerAPI(ctx context.Context, l *cla
 	}
 
 	nvr, arch, err := extractBuildNVR(ctx, path, buf)
-	if err != nil {
-		return nil, err
-	}
-	if nvr == "" || arch == "" {
+	switch {
+	case errors.Is(err, nil):
+	case errors.Is(err, errBadDockerfile):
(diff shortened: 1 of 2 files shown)


rhel/repositoryscanner.go
```diff
@@ -318,16 +322,31 @@ func extractBuildNVR(ctx context.Context, dockerfilePath string, buf *bytes.Buff
...
-		return "", "", fmt.Errorf("dockerfile missing expected label %q", arch)
+		return "", "", missingLabel(arch)
 	}
 	v, r := parseVersionRelease(filepath.Base(dockerfilePath))
 	return fmt.Sprintf("%s-%s-%s", n, v, r), a, nil
 }
 
+var errBadDockerfile = errors.New("bad dockerfile")
+
+type missingLabel string
```
crozzy, line 337: As this fulfills the `Error` interface, should it be prefixed `err` for easy recognition?

crozzy (approved):
LGTM, non-blocking comment

rhel/repositoryscanner.go
hdonnay: As it's an unexported type, I don't think it matter enough to have you hit the button again.

PR #536, opened 2021-12-07
dockerfile: handle whitespace runs correctly

hdonnay:
This change adds explicit tracking for whitespace runs in the key-value
splitting routine. This returning incorrect results violated assumptions
elsewhere and resulted in panics.

Closes: #526


crozzy (approved):
LGTM

crozzy (approved):

PR #548, opened 2022-01-12
Checking airgap flag before Ecosystsem default value check.

bluefriday:
Checking airgap flag before Ecosystsem default value check.


hdonnay:
This seems wrong, as it would change the defaults if the `Airgap` option is set.

PR #550, opened 2022-01-14
debian: split OVAL feed into binary packages 

crozzy:
The Debian OVAL feed reports vulnerabilities by the source, this
change splits those source vulnerabilities into binary packages so they
can be matched easily with the packages the scanner finds in the distro.

### Doubts
- [x] How will this affect users upgrading and having both the source and the binary vulns in the DB
- [x] What to do in airgap?


crozzy:
[Pre-Change Vulnerability Report](https://gist.github.com/crozzy/25cc2f792e1d048b53b7f191682209b6)
[Post-Change Vulnerability Report](https://gist.github.com/crozzy/b47ea173b110697c2628d8214ee8335d)

debian/parser.go
```diff
@@ -28,6 +28,16 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
 		return nil, fmt.Errorf("debian: unable to decode OVAL document: %w", err)
 	}
 	zlog.Debug(ctx).Msg("xml decoded")
+
+	err := u.sm.Update(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("could not update source to binary map")
+	}
```
hdonnay, line 35: This should run in the fetch step, if it's talking to the network.

debian/sourcemapper.go
```diff
@@ -0,0 +1,170 @@
...
+	return bins
+}
+
+// Update pulls the Sources.gz files for the different repos and saves
+// the resulting source to binary relationships.
+func (m *SourcesMap) Update(ctx context.Context) error {
+	return m.updateSources(ctx)
+}
+
+func (m *SourcesMap) updateSources(ctx context.Context) error {
```
hdonnay, line 96: Two methods is needless, here.

debian/sourcemapper.go
```diff
@@ -0,0 +1,170 @@
...
+		// There are no Wheezy records we assume the source->binary relationship of Jessie.
+		m.release = Jessie
+	}
+	for _, r := range repos {
+		url := fmt.Sprintf(m.sourcesURL, m.release, r)
+		err := m.fetchSources(ctx, url)
+		if err != nil {
+			return fmt.Errorf("unable to fetch sources: %w", err)
+		}
+	}
```
hdonnay, line 107: Do this in goroutines if all the locking is happening, anyway.

debian/sourcemapper.go
```diff
@@ -0,0 +1,170 @@
...
+	newEtag := resp.Header.Get("etag")
+	if etag, ok := m.etagMap[url]; ok && etag == newEtag {
+		zlog.Debug(ctx).Str("etag", etag).Msg("already seen this etag, skipping update")
+		return nil
+	}
+	m.etagMap[url] = newEtag
+
+	var reader io.ReadCloser
+	switch resp.Header.Get("Content-Type") {
+	case "application/x-gzip":
```
hdonnay, line 141:
```suggestion
	case "application/gzip", "application/x-gzip":
```

Should have the IANA type and the pre-IANA type, just for future proofing.

debian/sourcemapper.go
```diff
@@ -0,0 +1,170 @@
...
+			return err
+		}
+		defer reader.Close()
+	default:
+		return fmt.Errorf("received bad content-type %s querying mapping url %s", resp.Header.Get("Content-Type"), url)
+	}
+
+	tp := textproto.NewReader(bufio.NewReader(reader))
+	hdr, err := tp.ReadMIMEHeader()
+	for ; err == nil && len(hdr) > 0; hdr, err = tp.ReadMIMEHeader() {
```
hdonnay, line 153: `Err` should be checked after the loop to see if it's `EOF` or not.

pkg/ovalutil/rpm.go
```diff
@@ -36,6 +36,11 @@ func init() {
 // defining how a vulnerability should be parsed
 type ProtoVulnsFunc func(def oval.Definition) ([]*claircore.Vulnerability, error)
 
+// PackageExpansionFunc allows a caller to expand the inserted vulns. For example
+// when the OVAL DB reports vulnerabilities from the source package only (Debian). Or
+// the name field has a var_ref indicating a variable lookup is needed (Ubuntu).
+type PackageExpansionFunc func(def oval.Definition, name *oval.DpkgName) []string
+
```
hdonnay, line 43: `Rpm.go` is a weird place for a function only used in `dpkg.go`

debian/sourcemapper.go
```diff
@@ -0,0 +1,170 @@
...
+	case http.StatusOK:
+	default:
+		return fmt.Errorf("received status code %d querying mapping url %s", resp.StatusCode, url)
+	}
+
+	newEtag := resp.Header.Get("etag")
+	if etag, ok := m.etagMap[url]; ok && etag == newEtag {
+		zlog.Debug(ctx).Str("etag", etag).Msg("already seen this etag, skipping update")
+		return nil
+	}
```
hdonnay, line 136: This should make a conditional request (`If-None-Match`) to avoid network traffic.

debian/sourcemapper.go
```diff
@@ -0,0 +1,170 @@
...
+// inserts into the map. That needs to be done explitly by calling
+// the Update method.
+func NewSourcesMap(release Release, opts ...option) *SourcesMap {
+	d := &SourcesMap{
+		release:    release,
+		sourcesURL: sourcesURL,
+		mu:         &sync.RWMutex{},
+		sourceMap:  make(map[string]map[string]struct{}),
+		etagMap:    make(map[string]string),
+		client:     http.DefaultClient,
```
hdonnay, line 51: I'm of the opinion that new code should not be using `DefaultClient` -- if an `http.Client` is needed to work properly it should be an argument to the constructor.

hdonnay (changes requested):

libvuln/migrations/06-delete-debian-update_operation.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM update_operation WHERE updater IN (
+	'debian-bullseye-updater', 'debian-buster-updater',
+	'debian-jessie-updater', 'debian-stretch-updater',
+	'debian-wheezy-updater');
```
crozzy, line 4: Happy to do this by changing the fingerprints like with `pyupio` but this way the GC should delete all the "old style" debian vulns on the initial update (after upgrade).

debian/sourcemapper.go
```diff
@@ -0,0 +1,161 @@
...
+		defer reader.Close()
+	default:
+		return fmt.Errorf("received bad content-type %s querying mapping url %s", resp.Header.Get("Content-Type"), url)
+	}
+
+	tp := textproto.NewReader(bufio.NewReader(reader))
+	hdr, err := tp.ReadMIMEHeader()
+	for ; err == nil && len(hdr) > 0; hdr, err = tp.ReadMIMEHeader() {
+		source := hdr.Get("Package")
+		if source == "linux" {
```
hdonnay, line 141: Does this ever have the version attached? I know it sometimes happens in other apt databases.

hdonnay (changes requested):

debian/sourcemapper.go
```diff
@@ -0,0 +1,161 @@
...
+	if m.sourceMap[source] == nil {
+		return []string{}
+	}
+	bins := []string{}
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	for bin := range m.sourceMap[source] {
+		bins = append(bins, bin)
+	}
+	return bins
```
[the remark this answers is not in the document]
crozzy: Yes on the mutex :facepalm:

As for the copy: I'm using the golang "naive" set (`map[string]struct[]`) to dedupe entries and hence the bloated for loop when unwinding the values.

debian/sourcemapper.go
crozzy: There are some binaries with the version attached but it seems like they tend to be for the packages that also have the package attatched i.e.: 

```
Package: tcl8.6
Binary: tcl8.6, libtcl8.6, tcl8.6-dev, tcl8.6-doc
Version: 8.6.11+dfsg-1
Maintainer: Debian Tcl/Tk Packagers <pkg-tcltk-devel@lists.alioth.debian.org>
```

debian/sourcemapper.go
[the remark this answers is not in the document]
hdonnay: ah, yeah I misread the type

debian/sourcemapper.go
hdonnay: Yeah, poking around, it seems to only be in the `Packages` file, for when multiple builds were needed. E.g.

```Packages
Package: acmetool
Source: acmetool (0.2.1-4)
Version: 0.2.1-4+b5
```
and
```Sources
Package: acmetool
Binary: acmetool
Version: 0.2.1-4
```

debian/sourcemapper_test.go
```diff
@@ -0,0 +1,30 @@
...
+	opensshBinaries := mapper.Get("openssh")
+	if len(opensshBinaries) != 8 {
+		t.Fatalf("expected 8 binaries related to openssh found %d found %v", len(opensshBinaries), opensshBinaries)
+	}
+
+	tarBinaries := mapper.Get("tar")
+	if len(tarBinaries) != 2 {
+		t.Fatalf("expected 2 binaries related to openssh found %d found %v", len(tarBinaries), tarBinaries)
+	}
+}
```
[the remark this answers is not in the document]
crozzy: Let me know what you think of the unit-test. Seemed eventually to me simpler to manipulate the client rather than create a test-server due to no having access to the generated URLs.

hdonnay (approved):
lgtm

hdonnay (approved):
also LGTM

PR #555, opened 2022-01-21
Add Prometheus Collector to pgx pools

### LINKED ISSUE #554: It is hard to work out how many connections are being used
We had a problem last year when upgrading Clair whereby our configuration did not allow enough connections to Postgres for Clair to run.  This prooved to be tricky to track down because it was hard to see how many connections in a given pool were being used.

To help with this problem I have written a [Prometheus Collector for pgxpools](https://github.com/IBM/pgxpoolprometheus) and integrating this into Clair would help to fix this problem.

iainduncani:
Fixes #554 

Signed-off-by: Iain Duncan <iain.duncan@uk.ibm.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/knqyf263/go-deb-version v0.0.0-20190517075300-09fca494f03d
 	github.com/knqyf263/go-rpm-version v0.0.0-20170716094938-74609b86c936
 	github.com/prometheus/client_golang v1.9.0
+	github.com/prometheus/client_model v0.2.0
 	github.com/quay/alas v1.0.1
 	github.com/quay/goval-parser v0.8.6
 	github.com/quay/zlog v0.0.0-20210113185248-ce16eed1dcec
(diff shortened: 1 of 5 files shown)


iainduncani:
Hi @crozzy as discussed in the [pgxpoolprometheus PR](https://github.com/IBM/pgxpoolprometheus/pull/3#issuecomment-1022430601) I have updated this to have the collector definined within this library rather than pulling in a new dependency.  I noticed you don't use testify's testing libraries in here so I had to change the tests slightly to not use that (seemed to defeat the point of the copy otherwise!) but the collector.go file is the same as discussed over there.

I've also tried to fix up the integration test problem that the PR tests found before, although I couldn't get the integration tests to work locally (with or without my change!) so hoping that I have done the right thing...

iainduncani:
@crozzy in order to get the integration tests working I've needed to take out the `t.Parallel()`.  I'm assuming this was there to speed up the testing rather than because there is an expectation that a single Go application would run multiple different libindex instances?

crozzy (approved):

hdonnay:
I took the liberty of updating this. Changes I made:
- [x] renamed the package
- [x] replaced the static label, as applications will have multiple pools
- [x] don't panic on registration

iainduncani:
@hdonnay thanks for taking a look, the changes look good to me and nice you were able to add the `t.Parallel()` back in.

PR #562, opened 2022-02-03
libindex: change ordering

hdonnay:
This changes the ordering around the refcount, singleflight, and file on
disk. This should prevent some file open/remove races.

An alternative fix would perhaps be to return an *os.File instead of the
file name and have the proxy dup it.

### DIFF
--- a/libindex/fetcher.go
+++ b/libindex/fetcher.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
+	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


hdonnay:
This chunk of code is using multiple concurrency primitives (:rotating_light: not best practices :rotating_light:), so this change simplifies the usage by not having it strewn across two objects, then makes sure to check the state of the "optimistic" one (the singleflight) inside the critical section of the "correct" one.

crozzy (approved):

PR #566, opened 2022-02-11
Command removal

hdonnay:
This is an attempt to slim down our dependency tree and remove unused commands.

It ends up doing a whole host of things:

- Adapts the `test/bisect` tool to use the go testing framework, so that it can re-use the automatic database harness we have.
- Adds API to the database test harness to add a `Persistent` variant that doesn't use random tables/roles and doesn't drop them when done.
  - This is done for the benefit of the `bisect` tool, but may be useful in other debugging.
- Removes all the commands in `cmd` and replaces them with stubs that print and exit.
- Moves a `go generate` use of `cctool` to use `bisect`.
- Removes make targets that use the commands.
- Updates documentation.
- Adds a documentation preprocessor to validate make targets.


crozzy (dismissed):

crozzy (dismissed):

hdonnay:
clean rebase, and tests, so merging with superpowers

PR #572, opened 2022-02-22
jar: use textproto in place of mail

hdonnay:
This replaces the mail use with textproto, to keep it in line with the
other use for a similar purpose in the dpkg package.

It also rewrites the mainSectionReader to avoid double-buffering reads.


crozzy (approved):
nice!

PR #574, opened 2022-02-28
dpkg: correct double epoch

crozzy:
This change mitigates against a bug in the debian
OVAL generation code that always adds an epoch to the version even
if one already exists. This code should be reverted once that bug is
fixed.

TODO
- [x] Write tests
- [x] Verify formatting of Ubuntu versions
- [ ] DB migration deleting existing fingerprint?

### DIFF
--- a/pkg/ovalutil/dpkg.go
+++ b/pkg/ovalutil/dpkg.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
 
 	"github.com/quay/goval-parser/oval"
 	"github.com/quay/zlog"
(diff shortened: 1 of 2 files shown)


crozzy:
Related to https://github.com/quay/clair/issues/1270

cathyyoung:
FYI, there is an open merge request on the Debian OVAL generation code to fix the bug:
https://salsa.debian.org/webmaster-team/webwml/-/merge_requests/790

hdonnay:
@crozzy is this needed anymore, since that MR for the OVAL feed landed upstream?

crozzy:
nope

PR #576, opened 2022-03-03
Diagnostics groundwork

hdonnay:
As a part of PROJQUAY-3147, this PR lays groundwork for having all the modular components emit event logs.

### DIFF
--- a/toolkit/events/events.go
+++ b/toolkit/events/events.go
@@ -1,0 +1,164 @@
+// Package events is a small event logging system.
+package events
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"runtime"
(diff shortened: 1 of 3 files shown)


toolkit/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/quay/claircore/toolkit
+
+go 1.16
```
crozzy, line 3: Bump the go version?

Also, I thought with 1.17 lazy loading of packages was standard, is it required to have this be it'd own module?

crozzy (commented):
LGTM just a few question

crozzy (dismissed):

PR #577, opened 2022-03-03
Check validator on 200 response

### LINKED ISSUE #559: rhel updater set request doesn't return 304 even when Etag matches
I noticed that for the rhel updater set, when the request is made to retrieve vulnerability data, even when the request headers Etag matches the Etag in the response, the response does not return 304 (not modified). This just means we process the retrieve the full response and process the updaters rather than skipping them as we know the source data hasn't changed. 

As far as I understand this is due to cache settings at the server side and/or them not checking for the Etag header and so not returning the 304. 

It won't be causing a huge problem I just wanted to raise that the behaviour won't be as expected.

Code point for reference: https://github.com/quay/claircore/blob/main/rhel/updaterset.go#L138

To prove this I added some debugging lines to claircore, and also wrote a script to repeatedy query the source URL directly (with the Etag header added to request):

Debugging added to claircore (excuse the typos and random highlighting):
Screenshot 2022-01-25 at 15 06 28

Script output:
```
Response status code 200, manifestEtag W/"d9b124333616db388279769b8dd86009"
Response status code 200, manifestEtag W/"d9b124333616db388279769b8dd86009"
Response status code 200, manifestEtag W/"d9b124333616db388279769b8dd86009"
```

hdonnay:
This makes all the uses of an `if-none-match` header also check the response's Etag. Some servers send weak validators with non-identity content encodings, and don't validate them when sent back. This prevents processing unchanged resources, but doesn't prevent all the network traffic.

Closes #559

### DIFF
--- a/alpine/fetcher.go
+++ b/alpine/fetcher.go
@@ -36,7 +36,10 @@ func (u *Updater) Fetch(ctx context.Context, hint driver.Fingerprint) (io.ReadCl
 
 	switch res.StatusCode {
 	case http.StatusOK:
-		// break
+		if t := string(hint); t == "" || t != res.Header.Get("etag") {
+			break
+		}
+		fallthrough
 	case http.StatusNotModified:
 		zlog.Info(ctx).Msg("database unchanged since last fetch")
 		return nil, hint, driver.Unchanged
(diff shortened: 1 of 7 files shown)


crozzy (approved):
LGTM

PR #580, opened 2022-03-15
alpine: Ignore any vulns that have 0 as the secfixes version

crozzy:
Alpine apparently uses 0 to denote that the distro was
"never affected", hence it shouldn't show up in any
vulnerability reports. This change does no save any vulns
to the DB that have a 0 version.


crozzy:
https://github.com/quay/claircore/pull/580 is seems simpler

PR #581, opened 2022-03-15
alpine: Ignore any vulns that have 0 as the fixed_in_version

crozzy:
Alpine apparently uses 0 to denote that the distro was
"never affected", hence it shouldn't show up in any
vulnerability reports. This change checks the fixed_in_version
and indicates not vulnerable if it's 0.


crozzy:
fixes https://github.com/quay/clair/issues/1524

hdonnay (approved):
LGTM

PR #583, opened 2022-03-18
crda: Stop using the Client Clair instanciates

crozzy:
The remote matcher is currently getting passed the client
used for all updating that includes more specific settings
for that workload (like ratelimiting API calls). This change
moves the CRDA matcher to use its own client with a specific
timeout that matches the context deadline.

### DIFF
--- a/crda/matcherfactory.go
+++ b/crda/matcherfactory.go
@@ -108,7 +108,6 @@ func (f *Factory) Configure(ctx context.Context, cfg driver.MatcherConfigUnmarsh
 			Msg("configured API key")
 	}
 
-	f.client = c
 	f.ecosystems = fc.Ecosystems
 	return nil
 }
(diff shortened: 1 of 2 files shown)


crozzy:
The clair roundtripper client is where the `rate: Wait(n=1) would exceed context deadline` errors originate, these requests aren't making it out (hence no proxy logs when people are debugging I guess).

hdonnay (approved):
LGTM

We should make an issue to have another class of client in Clair that's configured in a central place with `user-agent` and the like, but meant for external requests at match-time.

crozzy:
https://github.com/quay/claircore/issues/585

> LGTM
> 
> We should make an issue to have another class of client in Clair that's configured in a central place with `user-agent` and the like, but meant for external requests at match-time.

https://github.com/quay/claircore/issues/585

hdonnay (approved):

PR #594, opened 2022-04-04
matcher: fail vuln report on matcher error

### LINKED ISSUE #593: Vuln report can be returned successfully but with no vulns if matchers fail
I’ve noticed some of our test images have been getting inconsistent results each time a vuln report is retrieved: sometimes showing the vulns as expected, and sometimes showing no vulns.  

These logs I was using for debugging show the same digest requested twice within seconds of each other, and the summary of the vuln report that was returned (I was restricted on log size so excuse not having the full report for each) (spot the difference tip: `vuln_report_package_vulns_count` and `vuln_report_vul_count`): 
  
```Mar 31 14:15:12 info {"level":"info","ts":"2022-03-31T13:15:12.117Z","image_digest":"sha256:9480a1752c6b10f08e8338617115cf2eca7d18f31e561a8cca445155eac84e97","vuln_report_distro_count":1,"vuln_report_package_count":97,"vuln_report_repo_count":0,"vuln_report_env_count":97,"vuln_report_enrichments_count":0,"vuln_report_package_vulns_count":44,"vuln_report_vuln_count":265}```
  
``` Mar 31 14:15:29 info {"level":"info","ts":"2022-03-31T13:15:29.916Z","image_digest":"sha256:9480a1752c6b10f08e8338617115cf2eca7d18f31e561a8cca445155eac84e97","vuln_report_distro_count":1,"vuln_report_package_count":97,"vuln_report_repo_count":0,"vuln_report_env_count":97,"vuln_report_enrichments_count":0,"vuln_report_package_vulns_count":0,"vuln_report_vuln_count":0}```

It seems like there is some `matcher error` logs in the clair logs at this time (though it is hard to be 100% these are related as there isn’t anything to tie the logs to a specific digest or request), which led me to this part of the code where I think we should be returning an error, rather than continuing, when there is an error from a matcher, otherwise we can return incomplete and misleading results:
https://github.com/quay/claircore/blob/main/internal/matcher/match.go#L118

### PR DESCRIPTION
**Pull request by paulaldridge**
vuln reports could be returned without vulns

Fixes #593

Error will be logged further up the stack, so removing log as well.

### DIFF
--- a/internal/matcher/match.go
+++ b/internal/matcher/match.go
@@ -3,6 +3,7 @@ package matcher
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"runtime"
 	"sync/atomic"
 
@@ -112,10 +113,7 @@ func EnrichedMatch(ctx context.Context, ir *claircore.IndexReport, ms []driver.M
 				}
 				vs, err := NewController(m, s).Match(mctx, records)
 				if err != nil {
-					zlog.Error(ctx).
-						Err(err).
-						Msg("matcher error")
-					continue
+					return fmt.Errorf("matcher error: %w", err)
 				}
 				vCh <- vs
 			}

### CODE REVIEW

**Comment by paulaldridge** [comment]
@hdonnay any chance you'd be able to have a look at this when you have time please? I'd be interested in your thoughts on it, thanks!

File: internal/matcher/match.go [hunk 1]
```diff
@@ -112,10 +113,7 @@ func EnrichedMatch(ctx context.Context, ir *claircore.IndexReport, ms []driver.M
 				}
 				vs, err := NewController(m, s).Match(mctx, records)
 				if err != nil {
-					zlog.Error(ctx).
-						Err(err).
-						Msg("matcher error")
-					continue
+					return fmt.Errorf("matcher error: %v", err)
```
hdonnay, line 116:
```suggestion
					return fmt.Errorf("matcher error: %w", err)
```

hdonnay (changes requested):
I think this should use the wrapping verb, and needs to be rebased.

I did this and then figured out I don't have permissions, I think there's a checkbox somewhere on your view of the PR that gives maintainers permissions on the branch.

internal/matcher/match.go
paulaldridge: Yep you're right it should be %w, thanks! Will update and rebase as well.

Interestingly the checkbox was already ticked when I found it :/ 
Screenshot 2022-05-17 at 12 57 42

PR #598, opened 2022-04-07
periodic: add package for periodic tests

hdonnay:
This also sets up a CI to trigger them weekly.


crozzy (approved):
SGTM

PR #613, opened 2022-05-16
Bug/connstring handling

hdonnay:
I mentioned this in #587; here's the commits for fixing the connection string handling.


crozzy:
Thanks, I'll port the changes over

PR #617, opened 2022-05-20
postgres: run creates before creating transaction

### LINKED ISSUE #616: Matcher can hang when it runs out of connections
# Description of Problem

To recreate create an matcher with this configuration:

```yaml
http_listen_addr: 0.0.0.0:6060
introspection_addr: 0.0.0.0:6061
log_level: debug
matcher:
    connstring: postgres://<redacted>@pgpool:9999/ibmclouddb
    indexer_addr: http://iclair-scan-results:6060
    max_conn_pool: 2
    migrations: true
    update_retention: -1
metrics:
    name: prometheus
    prometheus:
        endpoint: /metrics
updaters:
    sets:
    - alpine
    - debian
    - rhel
    - ubuntu
```

Note the limited number of `max_conn_pool`.  When this is deployed to a fresh database the vulnerabilities and update_operation tables are not populated.

The logs from the clair container are not particularly interesting but show that it's started doing stuff and then just hangs:

```
{"level":"debug","updater":"alpine-community-v3.4-updater","component":"alpine/Updater.Fetch","time":"2022-05-20T07:53:32Z","message":"successfully requested database"}
{"level":"debug","updater":"alpine-community-v3.4-updater","component":"alpine/Updater.Fetch","name":"/tmp/alpine-community-v3.4-updater.87419023","time":"2022-05-20T07:53:32Z","message":"created tempfile"}
{"level":"debug","updater":"alpine-community-v3.4-updater","component":"alpine/Updater.Fetch","time":"2022-05-20T07:53:32Z","message":"decompressed and buffered database"}
{"level":"debug","updater":"alpine-community-v3.4-updater","component":"alpine/Updater.Fetch","hint":"\"6089cb76-23f\"","time":"2022-05-20T07:53:32Z","message":"using new hint"}
{"level":"info","updater":"alpine-community-v3.4-updater","component":"apline/Updater.Parse","time":"2022-05-20T07:53:32Z","message":"starting parse"}
```

Running `SELECT * FROM pg_stat_activity` shows that the connection for the container is stuck in `Idle in transaction`:

```
 datid |  datname   |   pid   | usesysid |   usename  |    application_name       |  client_addr   | client_hostname | client_port |         backend_start         |          xact_start           |          query_start          |         state_change          | wait_event_type |     wait_event      |        state        | backend_xid | backend_xmin | query | backend_type 
-------+------------+---------+----------+------------+---------------------------+----------------+-----------------+-------------+-------------------------------+-------------------------------+-------------------------------+-------------------------------+-----------------+---------------------+---------------------+-------------+--------------+-------+---------------
 16473 | ibmclouddb | 1249213 |    16558 | <redacted> | iclair-sn-c6bd769f7-q6svb | 172.30.153.194 |                 |       23612 | 2022-05-20 07:53:31.498946+00 | 2022-05-20 07:53:32.394792+00 | 2022-05-20 07:53:32.394792+00 | 2022-05-20 07:53:32.394867+00 | Client          | ClientRead          | idle in transaction |             |            . | begin | client backend
```

Somewhere the transaction has begun and then claircore is trying to create a new connection and can't because it is already used up for the transaction that is sat doing nothing.  There are two places this can occur:

https://github.com/quay/claircore/blob/main/internal/vulnstore/postgres/updatevulnerabilities.go#L84-L97
https://github.com/quay/claircore/blob/main/internal/vulnstore/postgres/enrichment.go#L112-L125

# Background

I have been doing some performance testing on the query load of Clair and was seeing a similar problem doing queries where under heavy load clair would freeze and all of the queries to a specific pod would timeout.  Digging into that problem I found that it was caused by a similar create-transaction-call-queries problem in the enrichment:

https://github.com/quay/claircore/blob/v1.3.2/internal/vulnstore/postgres/enrichment.go#L195-L202

In that case the transaction was never used and as it happens @hdonnay [tidied it up this week!](https://github.com/quay/claircore/commit/46558830d91c05cd4c06c281513b0364d55727b1), a fix that I've tested on our setup and see it stops this hanging of queries.

Having seen that problem though it reminded me of our original setup where we had issues trying to keep the number of connections to the minimum possible but not being able to run a bare minimum for the vulnerability loading part of our setup and therefore wondered if the problem above existed.

I also saw that @hdonnay has removed the [hard-coding of 30 connections for `libindex` instances](https://github.com/quay/claircore/commit/44aeb933233e4e7aed732f4099b3f76bb74c4235) this week which will further help us out (although I'm a little freaked out that Hank appears to know what I'll want a week before I do!)

iainduncani:
Signed-off-by: Iain Duncan <iain.duncan@uk.ibm.com>

This fixes #616 and has been tested in our local setup.

I wasn't sure if using the transaction was the right approach, if the `QueryRow` commands were moved above the `pool.Begin` commands then that would also fix the problem without having to put these commands in the transaction.

### DIFF
--- a/internal/vulnstore/postgres/enrichment.go
+++ b/internal/vulnstore/postgres/enrichment.go
@@ -109,12 +109,6 @@ DO
 	)
 	ctx = zlog.ContextWithValues(ctx, "component", "internal/vulnstore/postgres/UpdateEnrichments")
 
-	tx, err := s.pool.Begin(ctx)
-	if err != nil {
-		return uuid.Nil, fmt.Errorf("unable to start transaction: %w", err)
-	}
-	defer tx.Rollback(ctx)
-
 	var id uint64
 	var ref uuid.UUID
 
@@ -126,6 +120,12 @@ DO
 
 	updateEnrichmentsCounter.WithLabelValues("create").Add(1)
 	updateEnrichmentsDuration.WithLabelValues("create").Observe(time.Since(start).Seconds())
+
+	tx, err := s.pool.Begin(ctx)
+	if err != nil {
+		return uuid.Nil, fmt.Errorf("unable to start transaction: %w", err)
+	}
+	defer tx.Rollback(ctx)
 
 	zlog.Debug(ctx).
 		Str("ref", ref.String()).
--- a/internal/vulnstore/postgres/updatevulnerabilities.go
+++ b/internal/vulnstore/postgres/updatevulnerabilities.go
@@ -81,12 +81,6 @@ func updateVulnerabilites(ctx context.Context, pool *pgxpool.Pool, updater strin
 	)
 	ctx = zlog.ContextWithValues(ctx, "component", "internal/vulnstore/postgres/updateVulnerabilities")
 
-	tx, err := pool.Begin(ctx)
-	if err != nil {
-		return uuid.Nil, fmt.Errorf("unable to start transaction: %w", err)
-	}
-	defer tx.Rollback(ctx)
-
 	var id uint64
 	var ref uuid.UUID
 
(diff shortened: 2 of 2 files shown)


crozzy:
Thanks for the PR! I _think_ this makes sense to house all the related mutating queries in the same tx, I wonder if there was a legacy reason for separating the `update_operation` insert and the `vuln` inserts.

hdonnay:
I think the original idea was to create the update op such that the system could see it immediately and another process running updaters wouldn't start processing the "same" update, then just get blocked in its transaction until the first committed.

I think a more correct approach may be to pull a connection out of the pool, create the op with it, use it for the transaction and cleanup if the transaction fails.

iainduncani:
@hdonnay does it need to use the same connection rather than just moving the create step before the tx is created?

crozzy (approved):

hdonnay (approved):

PR #626, opened 2022-06-03
rhel: Include cve defs when parsing through rhel oval feeds

crozzy:
There is condition that excludes cve definition types when converting
OVAL definitions to vulns, currently for rhel8 unpatched feeds
the definitions are either oval:com.redhat.cve:def:... or
oval:com.redhat.unaffected:def:...

### DIFF
--- a/rhel/parse_test.go
+++ b/rhel/parse_test.go
@@ -9,7 +9,65 @@ import (
 
 	"github.com/quay/goval-parser/oval"
 	"github.com/quay/zlog"
+
+	"github.com/quay/claircore/libvuln/driver"
 )
+
+func TestCVEDefFromUnpatched(t *testing.T) {
+	ctx, done := context.WithCancel(context.Background())
+	defer done()
+	var table = []struct {
+		name              string
+		fileName          string
+		configFunc        driver.ConfigUnmarshaler
+		expectedVulnCount int
+	}{
+		{
+			name:              "default path",
+			fileName:          "testdata/rhel-8-rpm-unpatched.xml",
+			configFunc:        func(_ interface{}) error { return nil },
+			expectedVulnCount: 192,
+		},
+		{
+			name:     "ignore unpatched path",
+			fileName: "testdata/rhel-8-rpm-unpatched.xml",
+			configFunc: func(c interface{}) error {
+				cfg, ok := c.(*UpdaterConfig)
+				if ok {
+					cfg.IgnoreUnpatched = true
+				}
+				return nil
+			},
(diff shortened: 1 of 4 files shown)


crozzy:
This came up when asked about why https://access.redhat.com/security/cve/cve-2021-35937 wasn't surfaced when reporting on rhel8 images. For this one definition it would add 192 (one hundred and ninety two) new vulns to the DB so we should probably talk about storage. @hdonnay do you have the context about why these vulns were initially excluded?

hdonnay:
The message [here](https://github.com/quay/claircore/commit/8cc282379a39c822b125a7c424b753b4fe9586eb) is as much context as I recall.

crozzy:
This would add ~400 vulnerabilities to an average RHEL8 manifest

rhel/parser.go
```diff
@@ -89,6 +89,5 @@ func isSkippableDefinitionType(defType ovalutil.DefinitionType) bool {
 	// TODO: Delete CVEDefinition of the condition when all work related
```
RTann, line 89: nit: you can remove this comment

rhel/parse_test.go
RTann, line 1: unrelated: should this be renamed to `parser_test.go`?

RTann:
This is essentially what ACS Scanner does today. When it comes to updating our vulnerability data in Postgres, we do the following:

* Upon startup, ACS Scanner contains the latest vulnerability data as of creating the release being used (this can get old quick...)
* ACS Scanner retrieves the a diff off the latest data vs what was originally embedded in the database
* This diff special cases Red Hat a bit: when creating the diff we try to find all the CVEs linked by Red Hat advisories, and mark them for deletion
* Upon updating the database in a live ACS Scanner, we delete all entries for the CVEs found based on the CVE name, package name, package module, and the CPEs listed in the related advisory entry.

I think that's the only thing special we had to do. I do not think ClairCore needs to worry about doing anything ACS had to, which is nice. ClairCore does not perform any kind of diffing like ACS does, so I believe what is done in this PR is all that is required.

RTann (approved):
LGTM, though I'm sure @hdonnay may be able to find any potential issues with this better than I can.

Don't forget to rebase :)

crozzy:
@RTann We discussed this a little and I added a config option to revert the behaviour

RTann (approved):

crozzy:
I think we should hold on this until after v4.7 just cause of the amount of updates, but would dearly love a review (hang the CI expense).

PR #640, opened 2022-06-14
wip: alpine: dynamic distributions

hdonnay:
These changes move the `alpine` package over to not relying on a static list of releases.


alpine/updater.go
```diff
@@ -1,91 +1,243 @@
...
+	defer res.Body.Close()
+	switch res.StatusCode {
+	case http.StatusNotModified:
+		zlog.Debug(ctx).
+			Stringer("url", u).
+			Msg("not modified")
+		return f.cur, nil
+	case http.StatusOK:
+	default:
+		return s, fmt.Errorf("alpine: unexpected status requesting %q: %v", "last-update", res.Status)
```
crozzy, line 99: Does the `last-update` string need to be formatted into the error message?

alpine/updater.go
```diff
@@ -1,91 +1,243 @@
...
+	}
+	var b bytes.Buffer
+	if _, err := b.ReadFrom(res.Body); err != nil {
+		return s, fmt.Errorf("alpine: error requesting %q: %w", "last-update", err)
```
crozzy, line 103: same here

crozzy (changes requested):
Will be a great improvement for sure, just a couple of comments

crozzy (approved):

PR #641, opened 2022-06-22
ubuntu: dynamic distributions

hdonnay:
These changes update Ubuntu to a new scheme for discovering available distributions at run-time instead of having a hard-coded list.

In addition, it makes some fixups in the `ovalutil` package.


ubuntu/updaterset.go
```diff
@@ -2,125 +2,307 @@ package ubuntu
...
 }
+
+type seriesResponse struct {
+	TotalSize        int            `json:"total_size"`
+	Start            int            `json:"start"`
+	Entries          []distroSeries `json:"entries"`
+	ResourceTypeLink string         `json:"resource_type_link"`
+}
+
+type distroSeries struct {
```
crozzy, line 241: This is kind of a nit, but is there a reason to define all these fields and not just `Active`, `Name` and `Version` (there might be others needed that I've missed)

crozzy (approved):
This looks good to me, just one comment

ubuntu/updaterset.go
hdonnay: No, these are just auto-generated from a live response. I'll bust them down.

crozzy (approved):

PR #643, opened 2022-06-29
debian: dynamic distributions

hdonnay:
These changes bring the `debian` package in-line with the new auto discovering scheme.


hdonnay:
forgot I need to improve a grip of log lines

updater/defaults/defaults.go
```diff
@@ -56,9 +56,13 @@ func inner(ctx context.Context) error {
 		return err
 	}
 	updater.Register("ubuntu", uf)
+	df, err := debian.NewFactory(ctx)
+	if err != nil {
+		return err
+	}
+	updater.Register("debian", df)
```
crozzy, line 63: considering the constructor is doing:
```go
func NewFactory(ctx context.Context) (*Factory, error) {
	f := &Factory{}
	return f, nil
}
```
do we need to return the error (and subsequently instantiate it in this different way?

debian/updater.go
```diff
@@ -1,98 +1,357 @@
...
 
 var (
-	_ driver.Updater      = (*Updater)(nil)
-	_ driver.Configurable = (*Updater)(nil)
+	// LinkRegexp is a bad regexp to extract link targets.
+	// This will break if Debian's codenames include a double-quote in the future.
+	linkRegexp = regexp.MustCompile(`href="([^"]+)"`)
+	// SkipList is a list of strings that, experimentally, indicate the string
+	// is not a codename.
+	skipList = []string{
```
crozzy, line 128: Could possibly add `updates` and `backports` here to avoid more requests

crozzy (commented):

updater/defaults/defaults.go
hdonnay: I'm just making it look like the new, future proofed constructor functions.

crozzy (approved):

hdonnay:
clean rebase, so I'm merging with superpowers.

PR #645, opened 2022-06-30
URL documentation helper

hdonnay:
This adds and then uses a helper to extract URLs from go source directly into documentation.

### DIFF
--- a/crda/remotematcher.go
+++ b/crda/remotematcher.go
@@ -25,7 +25,8 @@ var (
 )
 
 const (
-	batchSize     = 10
+	batchSize = 10
+	//doc:url matcher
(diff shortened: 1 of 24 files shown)


hdonnay:
Here's a the rendered documentation:

image

crda/remotematcher.go
```diff
@@ -25,7 +25,8 @@ var (
 )
 
 const (
-	batchSize     = 10
+	batchSize = 10
+	//doc:url indexer
```
crozzy, line 29: I would have thought this warranted it's own category

crozzy (commented):
Nice idea,  just one question

crozzy (approved):

hdonnay:
Merging with green CI

PR #647, opened 2022-07-05
updates: Use the tx's methods to record updater's status

crozzy:
This change makes sure all calls to modify the updater_status
table go through the transaction's connection instead of attempting
to acquire another one.


hdonnay (approved):

crozzy:
cc @paulaldridge this was causing prod resource contention

PR #651, opened 2022-07-08
tarfs: format error

hdonnay:
This change adds a dedicated error for when an invalid tarball is detected, and then adds a validation step to the fetcher.


crozzy (approved):
This lgtm, we should probably create an issue for Clair to start checking for this error and returning something in the 4XX range

hdonnay:
made the suggested change, merging with superpowers

PR #671, opened 2022-08-29
rpm: add support for NDB backend

crozzy:
RHEL has move to sqlite but NDB is also getting
some traction. This adds support for NDB which is
pretty close to BDB so shares a lot of code.

TODO
- [ ] Add support for registry.suse.com to the test Fetcher


hdonnay:
Thoughts on this vs #680?

crozzy:
Let's use #680 to avoid using the RPM dependency

PR #675, opened 2022-09-21
tarfs: Symlink changes

crozzy:
Make files with symlinks in their paths accessible both through the symlink path and the original path. This also takes into account when tars have been concatenated and symlinked dirs have been previously defined as normal dirs.


hdonnay (approved):
SGTM, but I rewrote large parts of it

pkg/tarfs/tarfs_test.go
```diff
@@ -379,3 +379,88 @@ func TestKnownLayers(t *testing.T) {
...
+			if err != nil {
+				return err
+			}
+			t.Logf("path: %q", path)
+			seen, ok := test.expectedFS[path]
+			if !ok {
+				t.Fatalf("didn't expect path %s", path)
+			}
+			if seen {
+				t.Logf("we already saw this path: %s", path)
```
crozzy, line 452: I think, as the "exactly once" problem is solved with your update, this should be now be fatal (so we can catch any regressions).

crozzy (commented):
Also LGTM, just a comment

pkg/tarfs/tarfs_test.go
crozzy: I should just push this change

hdonnay (approved):

PR #677, opened 2022-09-23
notifier: delete manifest_index_manifest_id_package_id_dist_id_repo_id_idx

crozzy:
Currently, we have two indexes (manifest_index_manifest_id_package_id_dist_id_repo_id_idx and manifest_index_unique) that cover the same data. As manifest_index_unique is chosen by the query planner to service affected manifest queries *and* keeps the unique constraint, and manifest_index_manifest_id_package_id_dist_id_repo_id_idx takes up a large amount of storage, it is surplus to requirements.


crozzy:
```sql
EXPLAIN SELECT
	manifest.hash
FROM
	manifest_index
	JOIN manifest ON
			manifest_index.manifest_id = manifest.id
WHERE
	package_id = 61
	AND (
			CASE
			WHEN 3::INT8 IS NULL THEN dist_id IS NULL
			ELSE dist_id = 3
			END
		)
	AND (
			CASE
			WHEN null::INT8 IS NULL THEN repo_id IS NULL
			ELSE repo_id = 2
			END
		);
```
```
Nested Loop  (cost=4.35..21.86 rows=1 width=32)
  ->  Bitmap Heap Scan on manifest_index  (cost=4.20..13.68 rows=1 width=8)
        Recheck Cond: (package_id = 61)
        Filter: ((repo_id IS NULL) AND (dist_id = 3))
        ->  Bitmap Index Scan on manifest_index_unique  (cost=0.00..4.20 rows=6 width=0)
              Index Cond: (package_id = 61)
  ->  Index Scan using manifest_pkey on manifest  (cost=0.15..8.17 rows=1 width=40)
        Index Cond: (id = manifest_index.manifest_id)
```

hdonnay (dismissed):
LGTM

PR #680, opened 2022-09-27
RPM ndb support

hdonnay:
This adds native RPM ndb support --- "native" in the sense that it's implemented in-process.


rpm/packagescanner.go
```diff
@@ -201,26 +204,56 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
...
 				if err != nil {
-					return err
+					return nil, fmt.Errorf("rpm: error reading ndb db: %w", err)
 				}
-				pkgs = append(pkgs, ps...)
-				return nil
-			}(); err != nil {
-				return nil, fmt.Errorf("rpm: error reading sqlite db: %w", err)
+				defer f.Close()
+				r, ok := f.(io.ReaderAt)
```
crozzy, line 224: I'm a little confused here, will we ever get a `ReaderAt` from tarfs.Open() or is this in case tarfs.File fulfills more interfaces in the future?

rpm/ndb/package.go
```diff
@@ -0,0 +1,279 @@
...
+func (db *PackageDB) Parse(r io.ReaderAt) error {
+	const (
+		headerSz = 4 * 8
+		pageSz   = 4096
+	)
+
+	// Read and verify the header.
+	b := make([]byte, headerSz)
+	if _, err := r.ReadAt(b, 0); err != nil {
+		return fmt.Errorf("pkg: unable to read header: %w", err)
```
crozzy, line 33: This is a nit but maybe the error messages can be a little more descriptive as to where they're being raised as `index` and `pkg` are pretty generic in clair?

crozzy (commented):
First pass LGTM, just a couple of comment

rpm/packagescanner.go
hdonnay: The latter -- no point in spooling out to a temp file if/when my skunkworks changes that would allow for implementing io.ReaderAt land.

rpm/ndb/package.go
hdonnay: will do.

crozzy (approved):

PR #694, opened 2022-10-13
chore: update go dependencies

hdonnay:
Closes: #688 #689 #690 #691 #692 #693


hdonnay:
The dependabot suggested upgrade in #693 breaks zlog because of an API change, so we'd need to update that first.

hdonnay:
updated with new zlog and zerolog

crozzy (approved):

PR #701, opened 2022-10-18
gobin: enable matching machinery

hdonnay:
This enables `gobin` discovered Packages to appear in index reports and get matched.


crozzy (approved):
LGTM

crozzy (approved):

PR #711, opened 2022-10-24
fetcher: Add layer hash to error message

crozzy:
In order to better debug errors it helps to have the layer hash to be able to analyse the offending layer.


libindex/fetcher.go
```diff
@@ -312,7 +312,7 @@ func (a *RemoteFetchArena) realizeLayer(ctx context.Context, l *claircore.Layer)
 	switch _, err := tarfs.New(fd); {
 	case errors.Is(err, nil):
 	case errors.Is(err, tarfs.ErrFormat):
-		fallthrough
+		return "", fmt.Errorf("tarfs error with layer %s: %w", l.Hash.String(), err)
```
hdonnay, line 315: I think this should happen at the caller, so all returned errors get prefixed with the layer information

hdonnay (changes requested):

hdonnay (approved):

PR #720, opened 2022-10-27
sqlite: fs.FS support via vfs

hdonnay:


hdonnay:
After some debugging, I _think_ this is because the underlying `fs.File` implementations don't implement `io.Seeker` (see [here](https://gitlab.com/cznic/sqlite/-/blob/v1.20.0/vfs/patches.go#L94)).

PR #730, opened 2022-11-07
jar: Close META-INF

crozzy:
We open META-INF dir in the jar file, this closes it.


crozzy:
Something that came out of the OOM debugging, not actually sure if this makes any difference.

hdonnay (changes requested):
SGTM, but I think the defer is unneeded

hdonnay (approved):

PR #731, opened 2022-11-07
periodic: attempt to follow API-created links

hdonnay:
A few constructed APIs started 404/403-ing. This change attempts to only use links returned by the API. This is all reverse-engineered, so I can only be so annoyed at having to do some additional reversing.

### DIFF
--- a/test/periodic/rpm_test.go
+++ b/test/periodic/rpm_test.go
@@ -1,8 +1,10 @@
 package periodic
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


hdonnay:
This passes when actually run:
```
% go test -run RPM -enable
PASS
ok  	github.com/quay/claircore/test/periodic	13.274s
```
Ran without `-v` because it dumps all the API responses. That'll make debugging next time quicker, but does make the output largely un-paste-able.

crozzy (approved):

PR #733, opened 2022-11-09
ndb: handle noncontiguous package slots

hdonnay:
This should handle the case where the RPM DB has unpopulated slots. My previous reads of the RPM C code had me thinking that the DB was recompacted after transactions, but this seems to no be the case. This implements a sanity check that exists in the C slot reading code.

Closes: quay/clair#1645


crozzy (approved):
LGTM

Issue #735, opened 2022-11-10
rpm: failed to scan all layer contents: mkdir /tmp/rpmscanner.2708622941/var/lib: file exists


crozzy:
Strange error in prod from the RPM scanner, happens with all different files:

```failed to scan all layer contents: mkdir /tmp/rpmscanner.1694209508/var/lib/yum/yumdb/o: file exists```

PR #743, opened 2022-11-21
tarfs: lookup path bug

### LINKED ISSUE #736: rpm: failed to scan all layer contents: rpm: error walking fs: readdir Filter/Util/Call: file does not exist
Problematic layer added

hdonnay:
This fixes the `tarfs` code to use the full path where appropriate, instead of just the file name.

Closes #736


hdonnay:
This makes #735 worse, by letting some layers make more progress.

crozzy (approved):

PR #746, opened 2022-11-23
new changelog idea

hdonnay:
This is a prototype for generating a changelog from `git-notes` in the `changelog` ref.


hdonnay:
Here's an example:
```
% ./changelog-render -f v1.4.12
```
* * *
<a name="v1.4.13"></a>
## [v1.4.13] - 2022-11-23
[v1.4.13]: https://github.com/quay/claircore/compare/v1.4.12...v1.4.13

#### `Tarfs` robustness fixes
<details>
There are changes regarding internal behavior when needing to create
implied directories (previously, the code mistakenly used the file name
instead of the file path) and hard links (the code did not handle hard
links to directories and did not traverse them).
</details>

#### `Rpm` indexer robustness fixes
<details>
Some container build systems do not seem to follow OCI image best
practices and produce layers that are out-of-order. This change handles
that better when a layer needs to be extracted for the `rpm` indexer.
</details>

#### The `rhcc` indexer now tolerates misnamed Dockerfiles.
<details>
Prior to this change, the indexer assumed that the name was well-formed
in such a way that could cause a `panic`; this case is now checked for
and noted in the log.
</details>

hdonnay:
could go either way on the the detail toggles

hdonnay:
I'm also of the opinion that if we do this, the changelog file should get nuked and we can just use the release notes

hdonnay:
here's how to play with these scripts locally:

```sh
.github/scripts/changelog-edit -s # sync-only to get the changelog notes
.github/scripts/changelog-render -f v1.4.12 # generate a changelog from v1.4.12 to v1.4.13
.github/scripts/changelog-update -f v1.4.12 # update the changelog file
git checkout -- CHANGELOG.md # reset the changelog
```

crozzy (approved):
LGTM let's see how it goes

PR #750, opened 2022-11-30
Whiteout support

crozzy:


indexer/options.go
```diff
@@ -4,16 +4,17 @@ import (
 	"net/http"
 )
```
hdonnay, line 6:
```suggestion

// Opts is a compatibility alias.
//
// Deprecated: Calling code should use the new name.
type Opts = Options
```

datastore/postgres/filesbylayer.go
```diff
@@ -0,0 +1,114 @@
...
+		query = `
+		SELECT file.id,
+			   file.path
+		FROM file_scanartifact
+				 LEFT JOIN file ON file_scanartifact.file_id = file.id
+				 JOIN layer ON layer.hash = $1
+		WHERE file_scanartifact.layer_id = layer.id
+		  AND file_scanartifact.scanner_id = ANY($2);
+		`
+	)
```
hdonnay, line 57: I keep waffling on whether we should keep this style of sql statement. It's this or `embed`.

datastore/postgres/filesbylayer.go
```diff
@@ -0,0 +1,114 @@
...
+			Scan(&scannerIDs[i])
+		done()
+		if err != nil {
+			return nil, fmt.Errorf("store:FilesByLayer failed to retrieve file ids for scanner %q: %w", scnr, err)
+		}
+		filesByLayerCounter.WithLabelValues("selectScanner").Add(1)
+		filesByLayerDuration.WithLabelValues("selectScanner").Observe(time.Since(start).Seconds())
+	}
+
+	ctx, done := context.WithTimeout(ctx, 30*time.Second)
```
hdonnay, line 78: Unsure about adding more hard timeouts

datastore/postgres/indexfiles.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+	"github.com/quay/claircore/pkg/microbatch"
```
hdonnay, line 12: I don't know if we should keep using `microbatch` -- I guess it keeps this inline with the other indexer stuff.

datastore/postgres/indexfiles.go
```diff
@@ -0,0 +1,130 @@
...
+		ON CONFLICT (path, kind) DO NOTHING;
+		`
+
+		insertWith = `
+		INSERT
+		INTO file_scanartifact (layer_id, file_id, scanner_id)
+		VALUES (
+			(SELECT id FROM layer WHERE layer.hash = $1),
+			(SELECT id FROM file WHERE file.path = $2 AND file.kind = $3),
+			(SELECT id FROM scanner WHERE scanner.name = $4 AND scanner.version = $5 AND scanner.kind = $6)
```
hdonnay, line 53: I dunno if having 3 subselects on every insert is good for perf

datastore/postgres/migrations/indexer/06-file-artifacts.sql
```diff
@@ -0,0 +1,18 @@
+--- file
```
hdonnay, line 1: why the extra `-`?

file.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+type FileKind string
+
+const (
+	FileKindWhiteout = FileKind("whiteout")
+)
+
+type File struct {
+	ID   string
```
hdonnay, line 14: what's this used for?

file.go
```diff
@@ -0,0 +1,18 @@
...
+	Path string
+	Kind FileKind
+	Info fs.FileInfo
```
hdonnay, line 17: What's this used for?

indexer/filescanner.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+
+	"github.com/quay/claircore"
+)
+
+// FileScanner reports the Distributions found in a given layer.
+type FileScanner interface {
+	VersionedScanner
+	Scan(context.Context, *claircore.Layer) ([]claircore.File, error)
```
hdonnay, line 12: Use a different method name

indexreport.go
```diff
@@ -33,6 +33,8 @@ type IndexReport struct {
 	Success bool `json:"success"`
 	// an error string in the case the index did not succeed
 	Err string `json:"err"`
+	// Files doesn't end up in the json report but needs to be available at post-coalesce
+	Files map[string]File
```
hdonnay, line 37: Needs an annotation to keep it out of the marshal output

package.go
```diff
@@ -17,8 +17,12 @@ type Package struct {
 	Source *Package `json:"source,omitempty"`
 	// the file system path or prefix where this package resides
 	PackageDB string `json:"-"`
+	// File
+	File string `json:"-"`
 	// a hint on which repository this package was downloaded from
 	RepositoryHint string `json:"-"`
+	// Path is where the package was found (possible paths?)
+	Path string
```
hdonnay, line 25: why "file" _and_ "path"?

whiteout/coalescer.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"context"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+)
+
+type Coalescer struct {
+	// the IndexReport this Coalescer is working on
+	ir *claircore.IndexReport
```
hdonnay, line 12: don't hold a pointer in here, it should just be local to the `Coalesce` method.

hdonnay (commented):
I think this is definitely on the right track.

I'm a little jumpy about adding the `File` API -- I'll have to think on it more.

libindex/libindex_integration_test.go
```diff
@@ -202,8 +202,8 @@ func checkEqual(ctx context.Context, t *testing.T, tc testcase, lib *Libindex, i
 		cmp.FilterPath(func(p cmp.Path) bool {
 			s := p.Index(-3)
 			m := p.Last().String()
-			return s.Type() == reflect.TypeOf((*claircore.Package)(nil)) &&
-				(m == ".RepositoryHint" || m == ".PackageDB")
+			return m == ".Files" || (s.Type() == reflect.TypeOf((*claircore.Package)(nil)) &&
```
crozzy, line 205: This is a weird side-effect of excluding `Files` from the json output, `Index()` will return:
```
Files: map[string]claircore.File{}
```
And `IndexReport()` will return:
```
Files: nil
```
As it's been unmarshaled from the DB blob

datastore/postgres/indexfiles.go
crozzy: Updated to look up the scanner ID and layer ID first

whiteout/resolver.go
```diff
@@ -0,0 +1,92 @@
...
+	var checkFile string
+	switch {
+	case filepath.Base(whiteoutPath) == ".wh..wh..opq":
+		// Special opaque case, "indicating that all siblings are hidden in the lower layer"
+		checkFile = filepath.Dir(whiteoutPath)
+	default:
+		checkFile = strings.ReplaceAll(whiteoutPath, ".wh.", "")
+	}
+
+	// There has to be a better way
```
hdonnay, line 66: Perhaps take a look a the `filepath.IsRel` implementation and adapt it? This is probably not too far off

hdonnay (changes requested):

datastore/postgres/filesbylayer.go
crozzy: As discussed this will be in a separate PR

datastore/postgres/migrations/indexer/06-file-artifacts.sql
```diff
@@ -0,0 +1,18 @@
...
+-- FileScanArtifact
+-- A relation linking discovered file to a layer
+CREATE TABLE IF NOT EXISTS file_scanartifact (
+	  file_id bigint REFERENCES file(id) ON DELETE CASCADE,
+	  scanner_id bigint REFERENCES scanner(id) ON DELETE CASCADE,
+	  layer_id bigint REFERENCES layer(id) ON DELETE CASCADE,
+	  PRIMARY KEY(layer_id, scanner_id, file_id)
+);
+
+ALTER TABLE package_scanartifact ADD COLUMN file text;
```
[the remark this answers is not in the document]
crozzy: The `file` in this case is more like metadata associated to the package akin to `package_db` and `repository_hint`. The file where we find (metadata about) a certain python package for example, won't be persisted in the database but used as context when reasoning with the wo files. Maybe this is a naming issue and it should be `file_hint` or `location` or something.

datastore/postgres/filesbylayer.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, fmt.Errorf("store:FilesByLayer no file found for hash %v and scanners %v", hash, scnrs)
+	default:
+		return nil, fmt.Errorf("store:FilesByLayer failed to retrieve file rows for hash %v and scanners %v: %w", hash, scnrs, err)
+	}
+	filesByLayerCounter.WithLabelValues("query").Add(1)
+	filesByLayerDuration.WithLabelValues("query").Observe(time.Since(start).Seconds())
+	defer rows.Close()
+
+	res := []claircore.File{}
+	for rows.Next() {
```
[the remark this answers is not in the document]
crozzy: I think this is addressed

whiteout/resolver.go
crozzy: I took a look at  this today and I think it's the least worst option considering:

1. We need path prefix not just string prefix
2. We're lucky enough that the paths are not relative (so the code can be simpler),
3. We could use filepath.Match() but the fact that we don't know how deep the match will be i.e `.wo.root` and `root/algo/something/interesting.txt`. This makes the pattern generation non-trivial / makes it less readable IMO

file.go
RTann, line 1: Should we add some docs here?

indexer/filescanner.go
```diff
@@ -0,0 +1,13 @@
+package indexer
+
+import (
+	"context"
+
+	"github.com/quay/claircore"
+)
+
+// FileScanner reports the Distributions found in a given layer.
```
RTann, line 9: I don't fully follow this comment

indexer/options.go
```diff
@@ -4,15 +4,16 @@ import (
...
 
-// Opts are options to instantiate a indexer
-type Opts struct {
+// Options are options to instantiate a indexer
```
RTann, line 7: should we consider doing this in another PR to reduce the blast readius?

indexer/versionedscanner.go
```diff
@@ -110,8 +110,28 @@ func (vs VersionedScanners) VStoRS() []RepositoryScanner {
 	return out
 }
 
+func (vs *VersionedScanners) FStoVS(scnrs []FileScanner) {
```
RTann, line 113: is this meant to overwrite the previous contents?

indexer/versionedscanner.go
```diff
@@ -110,8 +110,28 @@ func (vs VersionedScanners) VStoRS() []RepositoryScanner {
...
+	}
+	for i := 0; i < n; i++ {
+		(*vs)[i] = scnrs[i]
+	}
+}
+
+func (vs VersionedScanners) VStoFS() []FileScanner {
+	out := make([]FileScanner, len(vs))
+	for _, vscnr := range vs {
+		out = append(out, vscnr.(FileScanner))
```
RTann, line 128: guessing this is only called when this is guaranteed to be true?

libindex/libindex.go
```diff
@@ -105,6 +106,12 @@ func New(ctx context.Context, opts *Options, cl *http.Client) (*Libindex, error)
 			ruby.NewEcosystem(ctx),
 		}
 	}
+	// Add whiteout objects
+	// Always add the white out ecosystem
```
RTann, line 110: nit: whiteout

libindex/options.go
```diff
@@ -11,6 +11,11 @@ const (
 	DefaultLayerScanConcurrency = 10
 )
 
+// Opts is a compatibility alias.
+//
+// Deprecated: Calling code should use the new name.
+type Opts = Options
```
RTann, line 17: is this needed? isn't it already `Options`?

whiteout/scanner.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"context"
+	"fmt"
+	"io/fs"
+	"strings"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+	"github.com/quay/claircore/pkg/tarfs"
+	"github.com/quay/zlog"
```
RTann, line 12: nit: I think this is usually in a previous block

whiteout/scanner_test.go
RTann, line 1: Can we document somewhere the contents of the layers? As in, what is the difference between the basic and advanced? Just the number of files?

whiteout/coalescer.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+)
+
+type Coalescer struct {
+	// the IndexReport this Coalescer is working on
+	ir claircore.IndexReport
```
RTann, line 12: curious, why not just make this a pointer?

python/packagescanner.go
```diff
@@ -114,6 +114,7 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
 			Name:              strings.ToLower(hdr.Get("Name")),
 			Version:           v.String(),
 			PackageDB:         "python:" + pkgDB,
+			File:              n,
```
RTann, line 117: Do we only need this for Python? I imagine other language, like Java for sure. Probably not necessary for non-languages, right?

RTann (commented):
Curious, is it possible to add the whiteout package scanner to all ecosystems and then ignore any files in lower layers which eventually have the .wh. prefix (maybe coalesce in reverse layer order)? This would remove the need for the Resolve and FileScanner I think (assuming whiteout is a package scanner)

datastore/postgres/filesbylayer.go
```diff
@@ -0,0 +1,106 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v4"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
```
hdonnay, line 13: Fix grouping

datastore/postgres/filesbylayer.go
hdonnay, line 1: Don't use the weird old error formatting.

datastore/postgres/filesbylayer.go
```diff
@@ -0,0 +1,106 @@
...
+	scannerIDs := make([]int64, len(scnrs))
+	for i, scnr := range scnrs {
+		start := time.Now()
+		err := s.pool.QueryRow(ctx, selectScanner, scnr.Name(), scnr.Version(), scnr.Kind()).
+			Scan(&scannerIDs[i])
+		if err != nil {
+			return nil, fmt.Errorf("store:FilesByLayer failed to retrieve file ids for scanner %q: %w", scnr, err)
+		}
+		filesByLayerCounter.WithLabelValues("selectScanner").Add(1)
+		filesByLayerDuration.WithLabelValues("selectScanner").Observe(time.Since(start).Seconds())
```
hdonnay, line 71: The metrics only capture successful calls

datastore/postgres/indexfiles.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+	"github.com/quay/claircore/pkg/microbatch"
```
hdonnay, line 12: Fix grouping

datastore/postgres/indexfiles.go
hdonnay, line 1: Don't use the weird old error formatting.

datastore/postgres/migrations/indexer/06-file-artifacts.sql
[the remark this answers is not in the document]
hdonnay: Isn't it the same thing as `file.path`?

If it not related to the `file` type, then it's just confusing to have it in this PR

file.go
```diff
@@ -0,0 +1,12 @@
+package claircore
+
+type FileKind string
+
+const (
+	FileKindWhiteout = FileKind("whiteout")
+)
+
+type File struct {
```
hdonnay, line 9: ✍

file.go
```diff
@@ -0,0 +1,12 @@
+package claircore
+
+type FileKind string
```
hdonnay, line 3: ✍

indexer/filescanner.go
hdonnay, line 9: Copypaste error

indexer/options.go
hdonnay: I think it's fine to fix while we're in here.

indexer/resolver.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"context"
+
+	"github.com/quay/claircore"
+)
+
+// Resolver only needs to fulfill one method. It can be used for any
+// reasoning that needs to be done with all the layers in context.
+// Resolvers are called at the end of the coalesce step when reports
+// from separate scanners are merged.
```
hdonnay, line 12:
```suggestion
// Resolver is used for any reasoning that needs to be done with all the layers in context.
//
// Resolvers are called at the end of the coalesce step when reports
// from separate scanners are merged.
```

libindex/options.go
hdonnay: Yeah, this doesn't do anything here. I think this is confusing `indexer.Opts` and `libindex.Options`.

whiteout/coalescer.go
hdonnay, line 12: This shouldn't have a member, the struct can be local to the method.

whiteout/resolver.go
```diff
@@ -0,0 +1,90 @@
+package whiteout
+
+import (
+	"context"
+	"path/filepath"
+	"strings"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+	"github.com/quay/zlog"
```
hdonnay, line 10: Fix grouping

whiteout/resolver.go
```diff
@@ -0,0 +1,90 @@
...
+			finalEnvironments[pkgID] = ir.Environments[pkgID]
+
+		}
+	}
+	ir.Packages = finalPackages
+	ir.Environments = finalEnvironments
+	return ir
+}
+
+// fileIsDeleted returns whether or not the filepath(fp) has been deleted
```
hdonnay, line 52:
```suggestion
// FileIsDeleted returns whether or not the filepath(fp) has been deleted
```

whiteout/scanner.go
hdonnay, line 12: Fix grouping

whiteout/resolver.go
```diff
@@ -0,0 +1,90 @@
...
+// by the corresponding whiteoutPath. It follows the OCI spec for whiteouts:
+// https://github.com/opencontainers/image-spec/blob/main/layer.md#whiteouts
+func fileIsDeleted(fp, whiteoutPath string) bool {
+	var checkFile string
+	switch {
+	case filepath.Base(whiteoutPath) == ".wh..wh..opq":
+		// Special opaque case, "indicating that all siblings are hidden in the lower layer"
+		checkFile = filepath.Dir(whiteoutPath)
+	default:
+		checkFile = strings.ReplaceAll(whiteoutPath, ".wh.", "")
```
hdonnay, line 62: This needs to be handled component-wise: consider `a/.wh.file.wh.txt`

whiteout/scanner_test.go
hdonnay: Second

hdonnay (changes requested):

indexer/versionedscanner.go
crozzy: It is, probably worth looking at this and how it's used in the controller.Coalescer again in a different PR, adding missing docs.

indexer/versionedscanner.go
crozzy: Hmmm, even better, I don't think the inverse methods are ever called, I'm going to delete this method (and will delete the others in a different PR once confirmed not used)

libindex/options.go
crozzy: :+1:

python/packagescanner.go
crozzy: Yes we'll probably want to add it for all languages, I was between adding the python part and just doing to individual language support in separate PRs, I chose to add it as a blueprint.

crozzy:
> Isn't it the same thing as file.path?
>
> If it not related to the file type, then it's just confusing to have it in this PR

@hdonnay The `package_scanartifact.file` is metadata about where the package came from, it is not a first class `File` (perhaps `filepath` is a better name for this). e.g. when the python indexer finds a package it adds the `file(path)` metadata about where it found the package, it doesn't create and persist a `File` object. When the (whiteout) resolving step happens it uses the `package_scanartifact.file` and any whiteout `File`s to determine if the package still exists in the final image. We need these two pieces of context. 

With this PR the only `File`s created are whiteout file with the idea being that other file types could be added going forward.

whiteout/scanner_test.go
crozzy: :+1:

datastore/postgres/migrations/indexer/06-file-artifacts.sql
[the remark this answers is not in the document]
crozzy: Going to rename this `filepath`

whiteout/resolver.go
crozzy: Added the test

datastore/postgres/indexfiles.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create statement: %w", err)
+	}
+	insertFileScanArtifactWithStmt, err := tx.Prepare(ctx, "insertFileScanArtifactWithStmt", insertWith)
+	if err != nil {
+		return fmt.Errorf("failed to create statement: %w", err)
+	}
+
+	start = time.Now()
+	mBatcher := microbatch.NewInsert(tx, 500, time.Minute)
```
RTann, line 106: just curious, why do we have microbatch instead of using https://pkg.go.dev/github.com/jackc/pgx/v4#Batch directly?

indexer/resolver.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"context"
+
+	"github.com/quay/claircore"
+)
+
+// Resolver is used for any reasoning that needs to be done with all the layers in context.
+//
+// Resolvers are called at the end of the coalesce step when reports
```
RTann, line 11: Is this the indexing equivalent to Enricher?

libindex/libindex.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/quay/claircore/rhel/rhcc"
 	"github.com/quay/claircore/rpm"
 	"github.com/quay/claircore/ruby"
+	"github.com/quay/claircore/whiteout"
 )
 
 const versionMagic = "libindex number: 2\n"
```
RTann, line 34: unrelated: just curious what this is for and why it's more detailed than just a number

package.go
```diff
@@ -17,6 +17,8 @@ type Package struct {
 	Source *Package `json:"source,omitempty"`
 	// the file system path or prefix where this package resides
 	PackageDB string `json:"-"`
+	// a location in the layer where the package is located, this is useful for language packages.
```
RTann, line 20: what is the difference between this and `PackageDB`?

python/packagescanner_test.go
RTann, line 1: these changes seem like they must have been painful 😞

whiteout/coalescer.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+)
+
+type Coalescer struct{}
+
+// NewCoalescer is a constructor for a Coalescer
+func NewCoalescer() *Coalescer {
+	return &Coalescer{}
```
RTann, line 14: For Java, I think this was switched to start using `nil` instead of an empty struct. Is that a pattern we want to continue? @hdonnay

whiteout/resolver.go
```diff
@@ -0,0 +1,94 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+type layerSorter map[string]int
+
+func newLayerSorter(layers []*claircore.Layer) layerSorter {
+	ls := layerSorter{}
```
RTann, line 85: up to you since it's just a suggestion to the runtime and not guaranteed, but maybe initialize with the known length: `make(map[string]int, len(layers))`

whiteout/resolver.go
```diff
@@ -0,0 +1,94 @@
...
+func (r *Resolver) Resolve(ctx context.Context, ir *claircore.IndexReport, layers []*claircore.Layer) *claircore.IndexReport {
+	// Here we need to check if any of the packages
+	// found are moot due to whiteouts.
+	ls := newLayerSorter(layers)
+	finalPackages := map[string]*claircore.Package{}
+	finalEnvironments := map[string][]*claircore.Environment{}
+	for pkgID, pkg := range ir.Packages {
+		packageDeleted := false
+		// First env.IntroducedIn?
+		packageLayer := ir.Environments[pkgID][0].IntroducedIn.String()
```
RTann, line 29: this is guaranteed to be sorted by layer?

whiteout/resolver.go
```diff
@@ -0,0 +1,94 @@
...
+
+func newLayerSorter(layers []*claircore.Layer) layerSorter {
+	ls := layerSorter{}
+	for i, l := range layers {
+		ls[l.Hash.String()] = i
+	}
+	return ls
+}
+
+func (ls layerSorter) isChildOf(whiteoutLayer, packageLayer string) bool {
```
RTann, line 92: can you add a comment to make it clear this checks if `b < a` given `ls.isChildOf(a, b)`?

whiteout/resolver.go
```diff
@@ -0,0 +1,94 @@
...
+	ls := newLayerSorter(layers)
+	finalPackages := map[string]*claircore.Package{}
+	finalEnvironments := map[string][]*claircore.Environment{}
+	for pkgID, pkg := range ir.Packages {
+		packageDeleted := false
+		// First env.IntroducedIn?
+		packageLayer := ir.Environments[pkgID][0].IntroducedIn.String()
+		for fileLayer, f := range ir.Files {
+			// Check if it's a whiteout file, if it applies to the package's
+			// file and if the layer the whiteout file came from came after.
```
RTann, line 32: can we add s comment on why we don't look at the same layer (maybe even just say see the spec for why we don't bother)?

whiteout/resolver.go
```diff
@@ -0,0 +1,94 @@
...
+		// Special opaque case, "indicating that all siblings are hidden in the lower layer"
+		checkFile = filepath.Dir(whiteoutPath)
+	case strings.HasPrefix(filepath.Base(whiteoutPath), ".wh."):
+		origFileName := filepath.Base(whiteoutPath)[4:]
+		checkFile = filepath.Join(filepath.Dir(whiteoutPath), origFileName)
+	default:
+		return false
+	}
+
+	checkFileParts := strings.Split(checkFile, "/")
```
RTann, line 69: confirming the spec says only / is allowed vs Microsoft's favorite path separator: \ ?

whiteout/resolver_test.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+func TestIsDeleted(t *testing.T) {
+	type testcase struct {
+		Path     string
+		Whiteout string
+		Deleted  bool
+	}
+
+	for _, tc := range []testcase{
```
RTann, line 249: can you add
```
{
	Path:     "a",
	Whiteout: "a/.wh..wh..opq",
	Deleted:  false,
},
```

I believe this should be `false` but I think we'd get `true` based on the implementation? Let me know your thoughts on this

whiteout/resolver.go
crozzy: This is a good point, I'm updating this to be newest layer that has the package as it could feasibly have been installed, deleted, and re-installed.

package.go
crozzy: `PackageDB` is more used for storing the location of the DB/metadata (bdb, alpine `installed`, etc.), it's deletion doesn't mean that the packages are deleted, it just provides some context.

libindex/libindex.go
crozzy: No idea about this one

indexer/resolver.go
crozzy: Kind of, but with the distinction that a `Resolver` is (potentially) modifying the result once the layer indexing has finished. Enrichers really only add data.

datastore/postgres/filesbylayer.go
```diff
@@ -0,0 +1,107 @@
...
+	case errors.Is(err, nil):
+	case errors.Is(err, pgx.ErrNoRows):
+		return nil, fmt.Errorf("no file found for hash %v and scanners %v", hash, scnrs)
+	default:
+		return nil, fmt.Errorf("failed to retrieve file rows for hash %v and scanners %v: %w", hash, scnrs, err)
+	}
+	defer rows.Close()
+
+	res := []claircore.File{}
+	var i int
```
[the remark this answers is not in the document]
crozzy: https://github.com/quay/claircore/pull/750#discussion_r1126968970

datastore/postgres/indexfiles.go
crozzy: I believe it's just for more control over the batch i.e.  batch size, timeout

whiteout/resolver.go
crozzy: My understanding is that when the fs is converted to a `tarfs.FS` we will be basically dealing with a unix-like filesystem

libindex/libindex.go
hdonnay: It gets thrown in to the `state` hash when `libindex`'s logic changes, to ensure that consumers can see that they should re-index manifests.

whiteout/coalescer.go
hdonnay: Yeah, if it's effectively the zero struct, using nil is basically the same as using a pointer to a (the) literal

whiteout/resolver.go
hdonnay: tar is defined to have `slash` as the path separator

datastore/postgres/indexfiles.go
hdonnay: yeah, it does incremental submission. I'm not a huge fan, but I think the only way to get equivalent semantics would be to use the copy interface

whiteout/coalescer.go
crozzy: :+1:

python/packagescanner_test.go
```diff
@@ -70,12 +70,6 @@ var scanTable = []test.ScannerTestcase{
 		Hash:    "sha256:109a55eba749c02eb6a4533eba12d8aa02a68417ff96886d049798ed5653a156",
 		Scanner: &python.Scanner{},
 	},
-	{
```
RTann, line 73: how come this was dropped?

whiteout/resolver.go
```diff
@@ -0,0 +1,107 @@
...
+				packageLayer = ir.Environments[pkgID][i].IntroducedIn.String()
+			}
+		}
+		for fileLayer, f := range ir.Files {
+			// Check if it's a whiteout file, if it applies to the package's
+			// filepath and if the layer the whiteout file came from came after.
+			// The spec states: "Whiteout files MUST only apply to resources in
+			// lower/parent layers" hence why we don't check if they're in the same
+			// layer.
+			if f.Kind == claircore.FileKindWhiteout && ls.isChildOf(fileLayer, packageLayer) && fileIsDeleted(pkg.Filepath, f.Path) {
```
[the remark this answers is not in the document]
crozzy: Yes, you would still be able to see the packages, in order for whiteouts to work with package-management-origin-packages you would need to include a filepath when parsing the package denoting that the removal of the aforementioned filepath would mean the package is deleted. (The filepath to the package DB wouldn't make sense as deleting wouldn't mean deleting the package).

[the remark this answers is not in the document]
RTann: oh I see, we do not plan on adding the `file` attribute to things like rpm, dpkg, apk?

RTann (dismissed):

python/packagescanner_test.go
crozzy: Added back, good catch. It didn't exist when I went back to browse the old fixtures

whiteout/resolver.go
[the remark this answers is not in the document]
crozzy: Not on this first pass, maybe when we have a reliable `filepath` to key off.

hdonnay:
I think this looks good (pending rebase)

hdonnay (approved):

PR #756, opened 2022-12-07
pep440: unify on one package

hdonnay:
This moves all pep440 usage over to the local package instead of being split across two packages.


pkg/pep440/range.go
```diff
@@ -6,39 +6,43 @@ import (
 	"unicode"
 )
 
+//go:generate stringer -linecomment -type op
```
crozzy, line 9: https://github.com/quay/claircore/blob/b10fa7900054a953584c68777bff9c919406af00/pkg/pep440/range.go#L58 could maybe simplify this now that it's been stringered.

python/matcher.go
```diff
@@ -39,16 +39,20 @@ func (*Matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, v
 
 	v, err := pep440.Parse(record.Package.Version)
 	if err != nil {
+		zlog.Info(ctx).
```
crozzy, line 42: Thoughts on these being `warn`s?

crozzy (commented):
Looks good, couple of comments

python/matcher.go
hdonnay: Yeah, okay. If these are this far into the system without anything noticing the versions are weird, that's more of a problem.

crozzy (approved):

Issue #181, opened 2020-05-22
Unsupported scan results on Google Distroless images


eumel8:
# Description of Problem / Feature Request

When I push [Google Distroless Images](https://github.com/GoogleContainerTools/distroless) to my Quay registry, I've got Unsupported security scan results. I.e from [Debian Java Image](https://console.cloud.google.com/gcr/images/distroless/GLOBAL/java?gcrImageListsize=30)

### Expected Outcome

Security findings based on CVE for Debian.

### Actual Outcome

Unsupported

### Environment

- Clair version/image: clair-jwt:v3.2.2
- Clair client name/version: clair-jwt:v3.2.2
- Host OS: Ubuntu 18.04
- Kernel (e.g. `uname -a`):  4.15.0-76-generic 
- Kubernetes version (use `kubectl version`):  1.16.8
- Helm version (use `helm version`): 3.1.2
- Network/Firewall setup: canal

Additional infos: in 
https://github.com/quay/clair/blob/8cdd815ccdab27a2ded0e68740b27444efca8d1e/ext/featurefmt/dpkg/dpkg.go#L41 is a file regex for "var/lib/dpkg/status". It seems Google holds the  package information in one file per package in "var/lib/dpkg/status.d"

Maybe it's easy for a Go programmer to add a loop to catch the package information from there :-) Workaround would be an extra build to provide the required status file with information from the status directory.

ldelossa:
@hdonnay what do you think of this? 
It looks like google is packaging containers that utilize distributions we do support. 
At first glance it does seem like some simple modifications to the dpkg scanner would make this work. 

Running `cctool unpack`
```
/tmp/cctool-unpack-248681396
❯ fd os-release | xargs cat
PRETTY_NAME="Distroless"
NAME="Debian GNU/Linux"
ID="debian"
VERSION_ID="10"
VERSION="Debian GNU/Linux 10 (buster)"
HOME_URL="https://github.com/GoogleContainerTools/distroless"
SUPPORT_URL="https://github.com/GoogleContainerTools/distroless/blob/master/README.md"
BUG_REPORT_URL="https://github.com/GoogleContainerTools/distroless/issues/new"
PRETTY_NAME="Debian GNU/Linux 10 (buster)"

NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
[...]

PR #758, opened 2022-12-09
dpkg: Add distroless scanner

crozzy:
This change adds a distroless scanner which is in itself a different flavour of dpkg scanner. It reads the dpkg files found in the var/lib/dpkg/status.d/ directory.

### DIFF
--- a/dpkg/distroless_scanner.go
+++ b/dpkg/distroless_scanner.go
@@ -1,0 +1,148 @@
+package dpkg
+
+import (
+	"bufio"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"net/textproto"
+	"path/filepath"
+	"runtime/trace"
+
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+	"github.com/quay/claircore/pkg/tarfs"
+)
+
+const (
+	distrolessName    = "dpkg-distroless"
+	distrolessKind    = "package"
+	distrolessVersion = "1"
+)
+
+var (
+	_ indexer.VersionedScanner = (*Scanner)(nil)
+	_ indexer.PackageScanner   = (*Scanner)(nil)
+)
+
+// DistrolessScanner implements the scanner.PackageScanner interface.
+//
+// This looks for directories that look like dpkg databases and examines the
+// files it finds there.
+//
+// The zero value is ready to use.
+type DistrolessScanner struct{}
+
+// Name implements scanner.VersionedScanner.
+func (ps *DistrolessScanner) Name() string { return distrolessName }
+
+// Version implements scanner.VersionedScanner.
+func (ps *DistrolessScanner) Version() string { return distrolessVersion }
+
+// Kind implements scanner.VersionedScanner.
+func (ps *DistrolessScanner) Kind() string { return distrolessKind }
+
+// Scan attempts to find a dpkg database files in the layer and read all
+// of the installed packages it can find. These files are found in the
+// dpkg/status.d directory.
+//
+// It's expected to return (nil, nil) if there's no dpkg databases in the layer.
+//
+// It does not respect any dpkg configuration files.
+func (ps *DistrolessScanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircore.Package, error) {
+	defer trace.StartRegion(ctx, "Scanner.Scan").End()
+	trace.Log(ctx, "layer", layer.Hash.String())
+	ctx = zlog.ContextWithValues(ctx,
+		"component", "dpkg/DistrolessScanner.Scan",
+		"version", ps.Version(),
+		"layer", layer.Hash.String())
+	zlog.Debug(ctx).Msg("start")
+	defer zlog.Debug(ctx).Msg("done")
+
+	rd, err := layer.Reader()
+	if err != nil {
+		return nil, fmt.Errorf("opening layer failed: %w", err)
+	}
+	defer rd.Close()
+	sys, err := tarfs.New(rd)
+	if err != nil {
+		return nil, fmt.Errorf("opening layer failed: %w", err)
+	}
+
+	var pkgs []*claircore.Package
+	walk := func(p string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.Name() == "status.d" && d.IsDir() {
+			zlog.Debug(ctx).Str("path", p).Msg("found potential distroless dpkg db directory")
+			dbFiles, err := sys.ReadDir(p)
+			if err != nil {
+				return fmt.Errorf("error reading DB directory: %w", err)
+			}
+			for _, f := range dbFiles {
+				pkgCt := 0
+				fn := filepath.Join(p, f.Name())
+				ctx = zlog.ContextWithValues(ctx, "database-file", fn)
+				zlog.Debug(ctx).Msg("examining package database")
+				db, err := sys.Open(fn)
+				if err != nil {
+					return fmt.Errorf("reading database files from layer failed: %w", err)
+				}
+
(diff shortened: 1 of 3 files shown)


crozzy:
closes https://github.com/quay/clair/issues/1598

crozzy:
related to https://github.com/quay/claircore/issues/181

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil, fmt.Errorf("opening layer failed: %w", err)
+	}
+
+	var pkgs []*claircore.Package
+	walk := func(p string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.Name() == "status.d" {
+			zlog.Debug(ctx).Msg(p)
```
hdonnay, line 83: Dynamic message, for no reason apparently?

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+			for _, f := range dbFiles {
+				pkgCt := 0
+				fn := filepath.Join(p, f.Name())
+				ctx = zlog.ContextWithValues(ctx, "database-file", fn)
+				zlog.Debug(ctx).Msg("examining package database")
+				db, err := sys.Open(fn)
+				switch {
+				case errors.Is(err, nil):
+				case errors.Is(err, fs.ErrNotExist):
```
hdonnay, line 98: This looks like a porting artifact and should never happen because we're using the results of the `ReadDir`.

hdonnay (changes requested):

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,146 @@
...
+// dpkg/status.d directory.
+//
+// It's expected to return (nil, nil) if there's no dpkg databases in the layer.
+//
+// It does not respect any dpkg configuration files.
+func (ps *DistrolessScanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircore.Package, error) {
+	defer trace.StartRegion(ctx, "Scanner.Scan").End()
+	trace.Log(ctx, "layer", layer.Hash.String())
+	ctx = zlog.ContextWithValues(ctx,
+		"component", "distroless-dpkg/Scanner.Scan",
```
hdonnay, line 61:
```suggestion
		"component", "dpkg/DistrolessScanner.Scan",
```

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+	"github.com/quay/claircore/pkg/tarfs"
+)
+
+const (
+	distrolessName    = "distroless-dpkg"
```
hdonnay, line 23:
```suggestion
	distrolessName    = "dpkg-distroless"
```

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,146 @@
+// Package dpkg implements a package indexer for dpkg packages.
```
hdonnay, line 1:
```suggestion
```

Goofs up the documentation.

hdonnay (changes requested):

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,145 @@
...
+
+var (
+	_ indexer.VersionedScanner = (*Scanner)(nil)
+	_ indexer.PackageScanner   = (*Scanner)(nil)
+)
+
+// Scanner implements the scanner.PackageScanner interface.
+//
+// This looks for directories that look like dpkg databases and examines the
+// "base" and "netbase" files it finds there.
```
RTann, line 35: Only these two? `gcr.io/distroless/python3-debian11` has `openssl`, `libssl1`, and `libc6`

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,145 @@
...
+	distrolessKind    = "package"
+	distrolessVersion = "1"
+)
+
+var (
+	_ indexer.VersionedScanner = (*Scanner)(nil)
+	_ indexer.PackageScanner   = (*Scanner)(nil)
+)
+
+// Scanner implements the scanner.PackageScanner interface.
```
RTann, line 32: nit: `DistrolessScanner`

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,145 @@
...
+						}
+					}
+					pkgCt++
+					pkgs = append(pkgs, p)
+				}
+				switch {
+				case errors.Is(err, io.EOF):
+				default:
+					zlog.Warn(ctx).Err(err).Msg("unable to read entry")
+					goto Restart
```
RTann, line 131: can this potentially infinitely loop?

dpkg/distroless_scanner.go
crozzy: Bad comment, updating

dpkg/distroless_scanner.go
crozzy: I don't believe so, we're going to out when we get to the EOF, you might be seeing something I'm not

RTann: I just wonder if it's possible the underlying reader returns non-EOF errors without reading anything, so we keep trying over and over again due to some underlying filesystem issues or something. Maybe I'm overthinking it?

RTann: Something like this: https://go.dev/play/p/2NKKCBv3PUp

crozzy: I guess that is possible, could do something like this:
```go
switch {
case errors.Is(err, io.EOF):
default:
    if _, ok := err.(textproto.ProtocolError); ok {
        goto Restart
    }
    zlog.Warn(ctx).Err(err).Msg("unable to read entry")
}
```
so we only carry on

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,148 @@
...
+				pkgCt := 0
+				fn := filepath.Join(p, f.Name())
+				ctx = zlog.ContextWithValues(ctx, "database-file", fn)
+				zlog.Debug(ctx).Msg("examining package database")
+				db, err := sys.Open(fn)
+				if err != nil {
+					return fmt.Errorf("reading database files from layer failed: %w", err)
+				}
+
+				// The database is actually an RFC822-like message with "\n\n"
```
RTann, line 97: Sorry I forgot about this one. This may not be true. I'm looking at how StackRox/ACS handles it, and it looks like we append an extra `\n` to the contents of the file: https://github.com/stackrox/scanner/blob/master/ext/featurefmt/dpkg/dpkg.go#L241 (though we did not use `net/textproto`)

RTann: Actually I think I essentially recreated this, and you'd be able to read the contents and you'd get an `io.EOF` returned at the same time, which is ok from what I can tell. Though, if `err == nil`, do we still log `error reading DB file`?

RTann: oh I misread it, and I thought the `switch` was inside the `for` loop. We are expecting this to end with `io.EOF` in the happy-path, so it makes sense

RTann (approved):

dpkg/distroless_scanner.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+	"github.com/quay/claircore/pkg/tarfs"
+)
+
+const (
+	distrolessName    = "dpkg-distroless"
+	distrolessKind    = "package"
```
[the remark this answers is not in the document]
crozzy: Yeah, might also have a "file" type soon

hdonnay (dismissed):
LGTM, pending rebase

PR #764, opened 2022-12-19
gobin: Don't error out on EOF

crozzy:
This change catches EOF errors as empty files will throw them.


hdonnay (approved):
I was going to suggest a size check, since we stat the file anyway, but the size differences between formats would get weird fast. This seems like a good approach.

PR #762, opened 2022-12-15
rhel: Check that after casting to mappingFile we have a usable mapper

crozzy:
Currently it is possible that if the repo2cpe_mapping_url or the repo2cpe_mapping_file (or indeed if the endpoint is down) that we will panic as the mappingFile will cast to a nil. This will check for a nil mapper before it gets accessed and error out.

Signed-off-by: crozzy <joseph.crosland@gmail.com>

PR #769, opened 2023-01-03
airgapped: Remove airgapped logic from the core in lieu of the client being passed being airgap aware

crozzy:
This change removes the code that disables RPC scanners when airgap
is true allowing hybrid implementations to be configured with local
resources. It assumes that the client passed will be airgap aware.

TODO

- [x] Check for the error in the RHEL and the RHCC scanners once https://github.com/quay/claircore/pull/762 is merged
- [x] Add error check in the layerscanner to see if the error comes from the client being airgapped

### DIFF
--- a/indexer/layerscanner/layerscanner.go
+++ b/indexer/layerscanner/layerscanner.go
@@ -2,7 +2,9 @@ package layerscanner
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net"
 	"runtime"
 
 	"github.com/quay/zlog"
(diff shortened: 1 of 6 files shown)


indexer/layerscanner/layerscanner.go
```diff
@@ -235,9 +237,16 @@ func (r *result) Do(ctx context.Context, s indexer.VersionedScanner, l *claircor
 	default:
 		panic(fmt.Sprintf("programmer error: unknown type %T used as scanner", s))
 	}
-	if err != nil {
+	addrErr := &net.AddrError{}
+	switch {
+	case errors.Is(err, nil):
+	case errors.As(err, &addrErr):
```
crozzy, line 243: Could get more specific here if needed

hdonnay (dismissed):
LGTM

indexer/layerscanner/layerscanner.go
hdonnay: I don't think that's needed immediately.

PR #788, opened 2023-01-16
gobin: Add more logging

crozzy:
Adding logging so we can see more context in and around any errors that surface when opening and reading interesting files.


hdonnay (changes requested):
I guess this would be fine with the addition of some `TODO` so we come back and remove the logs after the error reporting chain gets fixed up.

hdonnay (dismissed):

PR #791, opened 2023-01-18
java: maven http request timeout added

kripanshdevtron:
maven package search http request timeout made configurable.

### DIFF
--- a/java/packagescanner.go
+++ b/java/packagescanner.go
@@ -34,12 +34,14 @@ var (
 )
 
 const DefaultSearchAPI = `https://search.maven.org/solrsearch/select`
+const DefaultRequestTimeout = 2
 
 // ScannerConfig is the struct used to configure a Scanner.
 type ScannerConfig struct {
 	// API is a URL endpoint to a maven-like REST API.
 	// The default is DefaultSearchAPI.
-	API string `yaml:"api" json:"api"`
+	API               string `yaml:"api" json:"api"`
+	APIRequestTimeout int    `yaml:"api_request_timeout" json:"api_request_timeout"`
 }
 
 // Scanner implements the scanner.PackageScanner interface.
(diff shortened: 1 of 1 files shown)


java/packagescanner.go
```diff
@@ -34,12 +34,14 @@ var (
...
 const DefaultSearchAPI = `https://search.maven.org/solrsearch/select`
+const DefaultRequestTimeout = 10
 
 // ScannerConfig is the struct used to configure a Scanner.
 type ScannerConfig struct {
 	// API is a URL endpoint to a maven-like REST API.
 	// The default is DefaultSearchAPI.
-	API string `yaml:"api" json:"api"`
+	API               string `yaml:"api" json:"api"`
+	APIRequestTimeout int    `yaml:"api_request_timeout" json:"api_request_timeout"`
```
hdonnay, line 44:
```suggestion
	APIRequestTimeout time.Duration `yaml:"api_request_timeout" json:"api_request_timeout"`
```

The timeout should just be a duration.

java/packagescanner.go
```diff
@@ -34,12 +34,14 @@ var (
 )
 
 const DefaultSearchAPI = `https://search.maven.org/solrsearch/select`
+const DefaultRequestTimeout = 10
```
hdonnay, line 37: If this is configurable, I'd like the default timeout to be much smaller, like 2 seconds.

hdonnay (changes requested):
Please squash the commits and have them follow the project conventions.

java/packagescanner.go
kripanshdevtron: done this to prevent extra memory, time.Duration contains extra params, which are not required.

kripanshdevtron: I have done this to prevent add extra memory to struct.

hdonnay: That's not a concern for a low-usage struct like this, compared to the documentation and language-level benefit. A time.Duration is an int64 under the hood, so there's no memory saved by making this an int.

hdonnay:
The code changes look good -- please squash the commits down to one and add a signoff.

PR #803, opened 2023-01-24
fetcher: Add a maximum layer size

crozzy:
This change adds a limit when downloading the layer files. The layer size is computed from the content-length reported from a HEAD request, this means it'll often be the size of the compressed layer and it requires the storage to accurately report the content-length.

### DIFF
--- a/libindex/fetcher.go
+++ b/libindex/fetcher.go
@@ -12,6 +12,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


libindex/fetcher.go
```diff
@@ -206,14 +209,32 @@ func (a *RemoteFetchArena) realizeLayer(ctx context.Context, l *claircore.Layer)
...
+		case err == nil:
+			if a.maxBytes > 0 && size > a.maxBytes {
+				return "", fmt.Errorf("fetcher: layer size (%d) larger than maximum layer size (%d)", size, a.maxBytes)
+			}
+		default:
+			zlog.Warn(ctx).Msg("unable to convert content-length header to bytes")
+		}
+	}
+
+	req.Method = http.MethodGet
```
hdonnay, line 233: This won't work with signed URLs, they usually have only a set of allowed methods.

crozzy: I changed it to just make the GET request which means we should avoid any latency implication as long as the length check comes before the body reading.

crozzy: need to update the commit msg

hdonnay:
This isn't the maximum layer size, though. It's a transfer limit, as we currently need to decompress it.

As implemented, it would make a configuration hard to reason about: if an operator allows 10 layers with a "maximum layer size" of of 500 MiB then allocates 5000MiB of disk, it's still possible (and expected behavior) to use more disk than that.

crozzy:
My thinking is that if you have a registry that allows layer files big enough that the request to Clair times out before it can even download all the compressed bytes then we should have a knob to configure a reasonable upper bound. The naming might be a bit off but essentially it's not about the storage but about the transfer time. The caller checking the size is also an option.

hdonnay:
If it's about a timeout, then the knob should be expressed in time.

crozzy:
Right, the point was to get out early before doing work, but I think we're on the same page that perhaps this would be more trouble to explain than being useful.

PR #807, opened 2023-01-30
rpm: filter layer contents when extracting

hdonnay:
This should reduce processing time and space requirements when handling rpm-containing layers.


crozzy (approved):
If I'm reading this right, we're going to be constructing a very sparse layer FS on disk and passing it to rpm as opposed to all the layer contents.

hdonnay:
Added a changelog item for this.

PR #810, opened 2023-02-03
alpine: update vulnerability link

RTann:
https://cve.mitre.org is deprecated in favor of https://www.cve.org, so Alpine vulnerabilities should point to the new website.

### DIFF
--- a/alpine/parser.go
+++ b/alpine/parser.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	nvdURLPrefix = "https://cve.mitre.org/cgi-bin/cvename.cgi?name=%s"
+	cveURLPrefix = "https://www.cve.org/CVERecord?id=%s"
 )
 
 var _ driver.Parser = (*updater)(nil)
@@ -59,7 +59,7 @@ func unpackSecFixes(partial claircore.Vulnerability, secFixes map[string][]strin
 			v := partial
 			v.Name = id
 			v.FixedInVersion = fixedIn
-			v.Links = fmt.Sprintf(nvdURLPrefix, id)
+			v.Links = fmt.Sprintf(cveURLPrefix, id)
 			out = append(out, &v)
 		}
 	}
--- a/alpine/parser_test.go
+++ b/alpine/parser_test.go
@@ -18,7 +18,7 @@ var dist310 = release{3, 10}.Distribution()
 var v3_10CommunityTruncatedVulns = []*claircore.Vulnerability{
 	{
 		Name:               "CVE-2018-20187",
-		Links:              "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-20187",
+		Links:              "https://www.cve.org/CVERecord?id=CVE-2018-20187",
 		Updater:            "alpine-community-v3.10-updater",
 		FixedInVersion:     "2.9.0-r0",
 		NormalizedSeverity: claircore.Unknown,
@@ -30,7 +30,7 @@ var v3_10CommunityTruncatedVulns = []*claircore.Vulnerability{
 	},
 	{
 		Name:               "CVE-2018-12435",
-		Links:              "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-12435",
+		Links:              "https://www.cve.org/CVERecord?id=CVE-2018-12435",
 		Updater:            "alpine-community-v3.10-updater",
 		FixedInVersion:     "2.7.0-r0",
 		NormalizedSeverity: claircore.Unknown,
@@ -42,7 +42,7 @@ var v3_10CommunityTruncatedVulns = []*claircore.Vulnerability{
 	},
 	{
 		Name:               "CVE-2018-9860",
-		Links:              "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-9860",
+		Links:              "https://www.cve.org/CVERecord?id=CVE-2018-9860",
 		Updater:            "alpine-community-v3.10-updater",
 		FixedInVersion:     "2.6.0-r0",
 		NormalizedSeverity: claircore.Unknown,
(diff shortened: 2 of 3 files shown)


RTann:
Not sure if something like this merits a DB migration. I'm also not sure how we'd migrate this using pure SQL anyway.

crozzy:
It's possible this will lead to a problem in this scenario:
* We merge and deploy this PR

crozzy:
I think this will lead to a problem in this scenario:
* We merge and deploy this PR
* An update is made to one of the alpine feeds and the update_operation fingerprint changes
* We save *all* the vulnerabilities in that feed (as the hash of the vulnerabilities don't match what is already in the DB)
* Vulnerabilities are duplicated in vulnerability reports

I tried to mimic this locally:
Pre:
```json
    "3943101": {
      "id": "3943101",
      "updater": "alpine-main-v3.16-updater",
      "name": "CVE-2022-37434",
      "description": "",
      "issued": "0001-01-01T00:00:00Z",
      "links": "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-37434",
      "severity": "",
      "normalized_severity": "Unknown",
      "package": {
        "id": "",
        "name": "zlib",
        "version": "",
        "kind": "source"
      },
      "distribution": {
        "id": "",
        "did": "alpine",
        "name": "Alpine Linux",
        "version": "",
        "version_code_name": "",
        "version_id": "3.16",
        "arch": "",
        "cpe": "",
        "pretty_name": "Alpine Linux v3.16"
      },
      "repository": {},
      "fixed_in_version": "1.2.12-r2"
    },

```

Post
```json
    "7001689": {
      "id": "7001689",
      "updater": "alpine-main-v3.16-updater",
      "name": "CVE-2022-37434",
      "description": "",
      "issued": "0001-01-01T00:00:00Z",
      "links": "https://www.cve.org/CVERecord?id=CVE-2022-37434",
      "severity": "",
      "normalized_severity": "Unknown",
      "package": {
        "id": "",
        "name": "zlib",
        "version": "",
        "kind": "source"
      },
      "distribution": {
        "id": "",
        "did": "alpine",
        "name": "Alpine Linux",
        "version": "",
        "version_code_name": "",
        "version_id": "3.16",
        "arch": "",
        "cpe": "",
        "pretty_name": "Alpine Linux v3.16"
      },
      "repository": {},
      "fixed_in_version": "1.2.12-r2"
    },
    "6072838": {
      "id": "6072838",
      "updater": "alpine-main-v3.16-updater",
      "name": "CVE-2022-37434",
      "description": "",
      "issued": "0001-01-01T00:00:00Z",
      "links": "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-37434",
      "severity": "",
      "normalized_severity": "Unknown",
      "package": {
        "id": "",
        "name": "zlib",
        "version": "",
        "kind": "source"
      },
      "distribution": {
        "id": "",
        "did": "alpine",
        "name": "Alpine Linux",
        "version": "",
        "version_code_name": "",
        "version_id": "3.16",
        "arch": "",
        "cpe": "",
        "pretty_name": "Alpine Linux v3.16"
      },
      "repository": {},
      "fixed_in_version": "1.2.12-r2"
    }
```

IIRC the new vulnerabilities should be associated with a new update_operation and we should only be querying the latest update_operation's vuln's so not entirely sure how they all end up in the same report here. I need to investigate more.

crozzy:
Yeah, I think this merits a DB "migration" (deletion of the alpine update_operations), otherwise I think we'll surface duplicate vulns until the GC gets round to clearing them up.

hdonnay:
I don't understand why it's not just a new updateOp with a whole new set of vulnerabilites

crozzy:
This will require a migration to avoid intermediate duplicates (before the GC gets to deleting the older `update_operations`). If these URLs aren't going away anytime soon I say we hold off and try and fix the problem directly by only querying the latest `update_operation`'s `vuln`s

RTann:
We can hold off on this for now, but we should return to this at some point, as this site is migrating to cve.org.

I'm not very familiar with the `update_operation` table, but I do see a migration in which all references to Alpine were deleted. What was the reason for that? Would that help here, too?

crozzy:
> We can hold off on this for now, but we should return to this at some point, as this site is migrating to cve.org.
> 
> I'm not very familiar with the `update_operation` table, but I do see a migration in which all references to Alpine were deleted. What was the reason for that? Would that help here, too?

yeah that would be the same deal here. When you delete an update operation the next time the GC runs (at the end of the update process) it will delete all the vulns that were associated to it so you basically get a fresh slate with that particular updater. 

So in theory: the migration is run deleting the alpine update_operations, the alpine updater runs and inserts all the vulns again (with their corrected URLs), then the GC runs and removes all the old alpine vulns.

RTann:
Also, I'm curious what makes this different from something like https://github.com/quay/claircore/pull/835 which updates all the Amazon Linux vulns? Why did we not need to worry about the update_operations there, but we do here?

crozzy:
> Also, I'm curious what makes this different from something like #835 which updates all the Amazon Linux vulns? Why did we not need to worry about the update_operations there, but we do here?

It is the same kind of deal. This is how I verified what the code seems to suggest might happen in existing installations when there is an upgrade and then a change to an alpine vulnerability DB, maybe there is some error in that process that I haven't accounted for:

- Run `main` and wait for the vulns to be populated
- Stop the matcher(s)
- Modify the finger print of one of the alpine updaters in the `update_operations` table to simulate a change to alpine vuln DB
- Pull this branch
- Restart the matcher(s)

When the matcher(s) start again they will see that they need to re-pull the alpine DB and save all the new vulns.

crozzy:
https://github.com/quay/claircore/pull/947 this should address the concerns about duplicate/incorrect vulns being returned

hdonnay (approved):

PR #816, opened 2023-02-07
chore: update go versions

hdonnay:
Start building the 1.20 container -- this is needed to pick up a fix in the `debug/elf` package.


crozzy (dismissed):

hdonnay:
Expect this to fail the 1.20 test, as the container image doesn't exist yet.

PR #827, opened 2023-02-14
python: consider path for some packages

hdonnay:
This considers where a python package was discovered:

For Debian python, dpkg installs to a "dist-packages" directory. For Red Hat python, rpm installs to "usr/lib/python".


python/packagescanner.go
```diff
@@ -119,13 +119,37 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
...
+	}
+
 	return out, fs.WalkDir(sys, ".", func(p string, d fs.DirEntry, err error) error {
 		ev := zlog.Debug(ctx).
 			Str("file", p)
 		switch {
 		case err != nil:
 			ev.Discard().Send()
 			return err
+		case d.Name() == `dist-packages` && d.Type().IsDir():
```
crozzy, line 144: What happens in the situation when someone does:
```
# ubuntu:latest
apt-get update && apt-get install -y python3-pip && pip install urllib3
ls -l /usr/local/lib/python3.10/dist-packages/urllib3-1.26.14.dist-info/INSTALLER
```
Would this branch mean that we'd skip this package despite it not coming from an Ubuntu package?

hdonnay: ah, yeah I misremembered how this works -- I'll fix it

python/packagescanner.go
```diff
@@ -119,13 +119,55 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
...
+	}
+
 	return out, fs.WalkDir(sys, ".", func(p string, d fs.DirEntry, err error) error {
 		ev := zlog.Debug(ctx).
 			Str("file", p)
 		switch {
 		case err != nil:
 			ev.Discard().Send()
 			return err
+		case dpkg && d.Type().IsDir():
```
crozzy, line 149: Should this be removed now?

hdonnay: yes. slow today.

crozzy (approved):

PR #835, opened 2023-02-17
aws: include epoch in FixedInVersion

RTann:
We are missing the epoch for the Amazon Linux vulnerability `FixedInVersion`. The `rpm` package scanner accounts for it, so it's possible we have been missing a bunch of Amazon Linux vulnerabilities.


hdonnay (approved):
LGTM

PR #836, opened 2023-02-17
ruby: support gemspec

RTann:


RTann:
I can make more tests in a bit, but I wanted to open this up for review anyway

ruby/testdata/rails.tar
RTann, line 1: I understand this is pretty large, but I'm not sure a better way to do this. I think we should have a Quay repo we can use for test images

crozzy: We could try making use of the `fetch.Layer()` functionality as we do here: https://github.com/quay/claircore/blob/main/rpm/packagescanner_test.go#L1616. We already have https://quay.io/repository/projectquay/clair-fixtures I can skopeo cp as needed and/give you access to push (if you don't have it).

hdonnay (changes requested):
This looks pending a rewrite down to a single commit

RTann:
> This looks pending a rewrite down to a single commit

not sure I follow?

hdonnay:
oh, I suppose it would. It does skip all the commit lints though, so make sure to keep a `Signed-off-by` trailer and a properly formatted subject line.

hdonnay (approved):

PR #839, opened 2023-02-21
chore: use string instead of []byte in pom analysis

RTann:
I feel like this looks cleaner than using `bytes`

### DIFF
--- a/java/jar/jar.go
+++ b/java/jar/jar.go
@@ -559,18 +559,18 @@ func (i *Info) parseProperties(ctx context.Context, r io.Reader) error {
 	var group, artifact, version string
 	s := bufio.NewScanner(r)
 	for s.Scan() && (group == "" || artifact == "" || version == "") {
-		b := bytes.TrimSpace(s.Bytes())
-		ls := bytes.SplitN(b, []byte("="), 2)
-		if len(ls) != 2 {
+		line := strings.TrimSpace(s.Text())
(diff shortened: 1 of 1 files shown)


java/jar/jar.go
```diff
@@ -559,18 +559,18 @@ func (i *Info) parseProperties(ctx context.Context, r io.Reader) error {
 	var group, artifact, version string
 	s := bufio.NewScanner(r)
 	for s.Scan() && (group == "" || artifact == "" || version == "") {
-		b := bytes.TrimSpace(s.Bytes())
-		ls := bytes.SplitN(b, []byte("="), 2)
-		if len(ls) != 2 {
+		line := strings.TrimSpace(s.Text())
+		key, value, found := strings.Cut(line, "=")
```
RTann, line 563: I wonder if we can rely on this always being `key=value` or if we should handle spaces?

hdonnay: I think _technically_ spaces could be there: https://en.wikipedia.org/wiki/.properties
But I think that in practice, maven never emits spaces there.

hdonnay (approved):

java/jar/jar.go
RTann: https://maven.apache.org/shared/maven-archiver/#pom-properties-content tells me it does not use spaces, so that's nice assurance

PR #840, opened 2023-02-21
java: support Jenkins plugin archives

RTann:
Each example plugin https://plugins.jenkins.io/ I tried had a `pom.properties`, but I also added updates to the `MANIFEST.MF` support just in case


java/jar/jar.go
```diff
@@ -418,10 +417,17 @@ func (i *Info) parseManifest(ctx context.Context, r io.Reader) error {
...
+		name = hdr.Get("Extension-Name")
+	case hdr.Get("Short-Name") != "":
+		// Attribute found in Jenkins plugins.
+		name = hdr.Get("Short-Name")
 	}
 	for _, key := range []string{
 		"Bundle-Version",
 		"Implementation-Version",
+		"Plugin-Version",
 		"Specification-Version",
```
RTann, line 431: Found an example plugin https://plugins.jenkins.io/ssh-credentials/ where `Specification-Version: 0.0` and `Plugin-Version: 305.v8f4381501156`, and the `Plugin-Version` is the correct one

java/jar/jar.go
```diff
@@ -418,10 +417,17 @@ func (i *Info) parseManifest(ctx context.Context, r io.Reader) error {
 		// This attribute is marked as "Deprecated," but there's nothing that
 		// provides the same information.
 		name = hdr.Get("Implementation-Vendor-Id")
+	case hdr.Get("Extension-Name") != "":
```
RTann, line 420: not 100% sure how reliable this is

java/packagescanner_test.go
```diff
@@ -683,14 +683,14 @@ var scanTable = []test.ScannerTestcase{
 				Name:           "commons-jelly-tags-fmt",
 				Version:        "1.0",
 				Kind:           "binary",
-				PackageDB:      "file:usr/share/jenkins/jenkins.war",
+				PackageDB:      "jar:usr/share/jenkins/jenkins.war",
```
RTann, line 686: now that we read "Extension-Name", we can parse this data from the JAR

java/jar/jar.go
```diff
@@ -161,8 +161,7 @@ func extractManifest(ctx context.Context, name string, z *zip.Reader) (Info, err
 	return i, nil
 }
 
-// ExtractProperties pulls pom.properties files out of the META-INF directory
-// of the provided zip.
+// ExtractProperties pulls pom.properties files out of the provided zip.
 func extractProperties(ctx context.Context, name string, z *zip.Reader) ([]Info, error) {
 	const filename = "pom.properties"
 	mf, err := z.Open(`META-INF`)
```
RTann, line 167: Removed the META-INF directory reference from the godoc because I don't see this variable actually used anywhere other than ensuring it exists? It doesn't look like we enforce pom.properties must exist within META-INF

hdonnay (commented):
I'd like a test for an hpi, perhaps grabbing the ssh-credentials plugin along the way.

This looks right (file renamed to `ssh-credentials.par`):
```
    jar_test.go:226: org.jenkins-ci.plugins:ssh-credentials/305.v8f4381501156 [META-INF/maven/org.jenkins-ci.plugins/ssh-credentials/pom.properties]
    jar_test.go:226: org.jenkins-ci.plugins:ssh-credentials/305.v8f4381501156(sha1:4f4e3bf2666c0857ae39e5e270adfb107742c86a) [ssh-credentials.par:META-INF/maven/org.jenkins-ci.plugins/ssh-credentials/pom.properties]
```

but it'd be good for the java package to check for regressions as the manifest/properties heuristics change.

java/jar/jar.go
hdonnay: Originally, the package always used filesystem semantics (see c37acbc11d39b0ab08e9e489c9974c2c6b936d09), but this isn't quite right, so I expect this is a holdover comment.

crozzy:
This looks good to me, the only question I have is whether `checkExt()` should be updated, it looks at the minute like it's only used in tests and when finding archives in archives, I assume for jenkins plugins any embedded archives won't be `.hpi`. Apart from that the test to check we're triggering on `hpi`s sounds like a good idea.

java/jar/ext.go
```diff
@@ -0,0 +1,17 @@
+package jar
+
+import "path"
+
+// ValidExt determines if the file name extension
+// is a valid JAR file extension.
+func ValidExt(name string) bool {
+	ext := path.Ext(name)
```
RTann, line 8: I have seen a few instances of `path` used and other instances of `filepath` being used. How do you usually determine when to use which? I tend to exclusively use `path` for URL related stuff. I believe it also makes sense for `zip`. However, the original `checkExt` used `filepath` and another check used `path`, and I can't determine why

RTann:
> This looks good to me, the only question I have is whether `checkExt()` should be updated, it looks at the minute like it's only used in tests and when finding archives in archives, I assume for jenkins plugins any embedded archives won't be `.hpi`. Apart from that the test to check we're triggering on `hpi`s sounds like a good idea.

I don't think you'd find `hpi` nor `jpi` embedded in a JAR, but I guess it doesn't hurt to check for it. I updated the PR to include that

RTann:
> > This looks good to me, the only question I have is whether `checkExt()` should be updated, it looks at the minute like it's only used in tests and when finding archives in archives, I assume for jenkins plugins any embedded archives won't be `.hpi`. Apart from that the test to check we're triggering on `hpi`s sounds like a good idea.
> 
> I don't think you'd find `hpi` nor `jpi` embedded in a JAR, but I guess it doesn't hurt to check for it. I updated the PR to include that

I'm wrong JARs may contain `.hpi` files. For example, the `jenkins.war` file

java/jar/ext.go
crozzy: IFAIK filepath should be used when dealing with files at the `os` level as it has the ability to make determinations based on the os filesystem and path should be used when the dealing with generic related paths.

Looks like for `Ext()` specifically the functions are the same save for some os compatibility guarding in `filepath.Ext`. I'd say you're right in using path as the os (whatever Clair is running on) has little to do with what is being inspected.

crozzy (approved):

RTann:
Rebased. Will merge with CI

PR #843, opened 2023-02-23
chore: skip 'none' rhel vulnerability definition type

RTann:
The 'none' type appears in "empty" OVAL v2 files (look for any 567 byte file like this one: https://access.redhat.com/security/data/oval/v2/RHEL8/openshift-service-mesh-2.3-including-unpatched.oval.xml.bz2).

We already ignore these because `def.Advisory.AffectedCPEList` is empty/`nil`, but this PR makes the code more defensive.

### DIFF
--- a/pkg/ovalutil/rpm.go
+++ b/pkg/ovalutil/rpm.go
@@ -12,12 +12,15 @@ import (
 	"github.com/quay/claircore"
 )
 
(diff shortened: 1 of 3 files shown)


hdonnay (approved):
LGTM

RTann:
rebased. will merge with CI

PR #844, opened 2023-02-23
suse: change security tracker URL

astebenkova:
The current link lacks relevant CVE statistics compared to the SUSE upstream.
Moreover, it doesn't include some of the SUSE distributions (e.g. Opensuse Leap 15.3, Opensuse Leap 15.4, etc.).

### DIFF
--- a/suse/distributionscanner.go
+++ b/suse/distributionscanner.go
@@ -49,6 +49,18 @@ var suseRegexes = []suseRegex{
 		regexp:  regexp.MustCompile(`(?i)SUSE Linux Enterprise Server 11`),
 	},
 	{
+		release: Leap154,
+		regexp:  regexp.MustCompile(`(?i)openSUSE Leap 15.4`),
+	},
+	{
+		release: Leap153,
+		regexp:  regexp.MustCompile(`(?i)openSUSE Leap 15.3`),
+	},
+	{
+		release: Leap152,
+		regexp:  regexp.MustCompile(`(?i)openSUSE Leap 15.2`),
+	},
+	{
 		release: Leap151,
 		regexp:  regexp.MustCompile(`(?i)openSUSE Leap 15.1`),
 	},
--- a/suse/distributionscanner_test.go
+++ b/suse/distributionscanner_test.go
@@ -8,13 +8,13 @@ import (
 )
 
 var enterpriseServer15OSRelease []byte = []byte(`NAME="SLES"
-VERSION="15-SP1"
+VERSION="15-SP4"
 VERSION_ID="15.1"
-PRETTY_NAME="SUSE Linux Enterprise Server 15 SP1"
+PRETTY_NAME="SUSE Linux Enterprise Server 15 SP4"
 ID="sles"
 ID_LIKE="suse"
 ANSI_COLOR="0;32"
-CPE_NAME="cpe:/o:suse:sles:15:sp1"`)
+CPE_NAME="cpe:/o:suse:sles:15:sp4"`)
 
 var enterpriseServer12OSRelase []byte = []byte(`NAME="SLES"
 VERSION="12-SP5"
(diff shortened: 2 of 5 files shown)


crozzy:
Hi, thanks for the PR, this looks like a better source, a few things:

* What do you mean by "CVE statistics"? The suse bugzilla links?
* This change won't automatically pull in CVE data for new releases, that will have to be done manually or we need to add dynamic discovery (e.g debian).
* The commit must begin with <scope>: <subject>

astebenkova:
Hi Joseph! Thanks for your reply.
> * What do you mean by "CVE statistics"? The suse bugzilla links?

Basically, I'm talking about the number of CVEs found by Clair. We've been using it to scan our images for a long time now and recently we started scanning `opensuse`-based images - Clair returns empty reports. I tested some other open-source tools for the same images and got a bunch of vulnerabilities. They use the upstream tracker URL (for the record, it was used by [claircore](https://github.com/quay/claircore/commit/3d53113b1f893dc0cb2de7eb58f61600c928e091) as well, but was changed due to security matter).

BTW, I tried to play with different options in Clair v4.6.0 configuration (according to the [docs](https://quay.github.io/clair/reference/config.html#updatersconfig)), but none succeeded... Could you please tell me if there is a way to replace the URL through the configuration?

> * This change won't automatically pull in CVE data for new releases, that will have to be done manually or we need to add dynamic discovery (e.g debian).

So, I would need to add new updaters manually? Or recreate the database? Could you please specify what steps would be required?
> * The commit must begin with :

Fixed.

crozzy:
> Hi Joseph! Thanks for your reply.
> 
> > * What do you mean by "CVE statistics"? The suse bugzilla links?
> 
> Basically, I'm talking about the number of CVEs found by Clair. We've been using it to scan our images for a long time now and recently we started scanning `opensuse`-based images - Clair returns empty reports. I tested some other open-source tools for the same images and got a bunch of vulnerabilities. They use the upstream tracker URL (for the record, it was used by [claircore](https://github.com/quay/claircore/commit/3d53113b1f893dc0cb2de7eb58f61600c928e091) as well, but was changed due to security matter).

Got it
> 
> BTW, I tried to play with different options in Clair v4.6.0 configuration (according to the [docs](https://quay.github.io/clair/reference/config.html#updatersconfig)), but none succeeded... Could you please tell me if there is a way to replace the URL through the configuration? 
```yaml
updaters:
  config:
    suse-updater-opensuse.leap.15.0:
      url: "https://ftp.suse.com/pub/projects/security/oval/opensuse.leap.15.0.xml"
```
>
> > * This change won't automatically pull in CVE data for new releases, that will have to be done manually or we need to add dynamic discovery (e.g debian).
> 
> So, I would need to add new updaters manually? Or recreate the database? Could you please specify what steps would be required?
> 
You would need to add the releases you need here: https://github.com/quay/claircore/blob/main/suse/releases.go or write something to dynamically discover releases as exists for debian, ubuntu, alpine etc.
> > * The commit must begin with :
> 
> Fixed.

astebenkova:
> You would need to add the releases you need here: https://github.com/quay/claircore/blob/main/suse/releases.go or write something to dynamically discover releases as exists for debian, ubuntu, alpine etc.

Ahh... I got it. Didn't know they were hardcoded.
I've just made some amendments based on logic (not familiar with `go` whatsoever). Could you please, take a look just to make sure it's ok?

> updaters:
>   config:
>     suse-updater-opensuse.leap.15.0:
>     url: "https://ftp.suse.com/pub/projects/security/oval/opensuse.leap.15.0.xml"

Yep, tried this one as well, but for `suse-updater-opensuse.leap.15.4`, and now I clearly see why it didn't work out.

hdonnay:
Is there any way to get a list of current SuSE releases?

astebenkova:
> Is there any way you know of to get a list of current SuSE releases? I'll look around a little

Hi Joseph!

Obviously, the first thought is to use the upstream link (`https://ftp.suse.com/pub/projects/security/oval/`) and somehow parse it. They already provide releases in the file names. However, I saw the way you did it for other Linux OS's and it seems this method wouldn't suffice...

I'll dig into it as well.

hdonnay:
I suppose we could also ask the suse security team if we can consider the index page API, but I'd consider that a method of last resort because parsing a limited subset of HTML is still annoying and prevents changes that are human-equivalent and the "out of support" would just be a heuristic based on updated time (e.g. assume any supported product has an update within the last two months).

astebenkova:
> I suppose we could also ask the suse security team if we can consider the index page API, but I'd consider that a method of last resort because parsing a limited subset of HTML is still annoying and prevents changes that are human-equivalent and the "out of support" would just be a heuristic based on updated time (e.g. assume any supported product has an update within the last two months).

I've looked through a bunch of information and it seems it would be difficult (rather impossible) to find a single source for both SUSE and OpenSUSE releases:
> SUSE Linux is available under two brands, openSUSE and SUSE Linux Enterprise. openSUSE is a free, community distribution driven by the openSUSE Project. It includes some of the latest "bleeding edge" Linux technologies and is designed for home users and enthusiasts. SUSE Linux Enterprise is Suse's tested and certified open-source solution for major enterprises.

Thus they manage them separately, including websites, registries, repositories etc.

On the other hand, they say in their [README](https://ftp.suse.com/pub/projects/security/oval/README):
> This directory provides OVAL data for various SUSE Linux products and the openSUSE distributions. 
The OVAL data is generated once a day.
Note that only the current products receive OVAL updates.

So, at least I reckon we can rely on the file timestamps as a method to determine current releases. However, you are right - it doesn't look trustworthy...

crozzy:
superseded by https://github.com/quay/claircore/pull/1432

PR #851, opened 2023-02-28
all: update generate directives with run commands

hdonnay:
This change uses `go run` in the generate directives where applicable. The indirect "tool" build tag is removed. This should track dependencies more accurately and has been supported for several go versions.

This means `go generate` should just work everywhere now, without needing any install setup.


crozzy (approved):

RTann:
I think I'd prefer keeping this in the `go.mod`, as version bumps only require updating a single file, and it can be updated by dependabot easily.

hdonnay:
> I think I'd prefer keeping this in the `go.mod`, as version bumps only require updating a single file, and it can be updated by dependabot easily.

That's a good point, I think just switching to the `go run` style has the same benefit.

hdonnay:
For more context, this came up because of quay/clair#1696, but I think @RTann is right that the same "version-in-command" solution makes less sense here.

hdonnay:
going to consider @crozzy's approval good enough

PR #854, opened 2023-03-02
datastore: Consolidate all IndexStore methods

crozzy:
An idea I'm experimenting with.

TODO

- [ ] Update metrics


datastore/postgres/store_metrics.go
```diff
@@ -0,0 +1,52 @@
...
+		Subsystem: "datastore_postgres",
+		Name:      "query_total",
+		Help:      "Database query count for noted query.",
+	}, metricLabels)
+)
+
+//go:embed queries/***/*.sql
+var queries embed.FS
+
+func getQuery(ctx context.Context, db string, name string, e *error) (string, func() *prometheus.Timer, func(*prometheus.Timer)) {
```
hdonnay, line 34: Could this return `(string, func() func())`, so that it can be used like:
```
query, timer := getQuery()
defer timer()()
pool.Exec(ctx, query)
```

hdonnay: I guess it closes over an error -- maybe this needs to be spilt so that this easy, one-off form is easier but the metrics could also be split out

crozzy: I was playing around with this:
```go
type query struct {
	SQL string

	labels prometheus.Labels
	timer  *prometheus.Timer
}

func (q *query) Start() {
	q.timer = prometheus.NewTimer(prometheus.ObserverFunc(func(v float64) {
		databaseTimer.With(q.labels).Observe(v)
	}))
}

func (q *query) End(err error) {
	q.labels["success"] = strconv.FormatBool(errors.Is(err, nil))
	databaseCounter.With(q.labels).Inc()
	if q.timer != nil {
		q.timer.ObserveDuration()
	}
}

func newQuery(ctx context.Context, db string, name string) query {
	b, err := fs.ReadFile(queries, path.Join("queries", db, name+".sql"))
	if err != nil {
		panic(err)
	}

	return query{
		SQL:    string(b),
		labels: prometheus.Labels{"query": name, "db": db},
	}
}
```
Which makes it fairly self explanatory for the caller:

```go
func (s *IndexerStore) LayerScanned(ctx context.Context, hash claircore.Digest, scnr indexer.VersionedScanner) (bool, error) {
	var scannerID int64
	selectScanner := newQuery(ctx, "indexer", "select_scanner")
	selectScanner.Start()
	err := s.pool.QueryRow(ctx, selectScanner.SQL, scnr.Name(), scnr.Version(), scnr.Kind()).
		Scan(&scannerID)
	selectScanner.End(err)
	switch {
	case errors.Is(err, nil):
	case errors.Is(err, pgx.ErrNoRows):
		return false, fmt.Errorf("scanner %s not found", scnr.Name())
	default:
		return false, err
	}

	var ok bool
	selectScanned := newQuery(ctx, "indexer", "select_layer_scanned")
	selectScanned.Start()
	err = s.pool.QueryRow(ctx, selectScanned.SQL, hash.String(), scannerID).
		Scan(&ok)
	if err != nil {
		return false, err
	}
	selectScanned.End(err)
	return ok, nil
}
```
Callers can also choose to never call `Start()` or `End()`

hdonnay: Start should  be something like `func(err *error) func()` so the idiomatic usage of `defer query.Start(&err)()` works. For example, the second call in the example leaks a timer

crozzy: I like this syntax (and have updated this PR), it doesn't fit into every use-case i.e where there are multiple separate queries in one function or batched queries. I've kind of gone back and forth on just timing the function but it feels like we'd miss out on something query specific metrics.

crozzy: Hmm, maybe for the batch stuff it makes sense to time the entire insert (up to `.Done()`) as how long it takes to queue the batch doesn't really tell us anything

crozzy:
Closing as out of data now

PR #856, opened 2023-03-02
jar: improve name heuristic

daynewlee:
Trying to have a better parsing result from jar Manifest so Clair can have a better CVE matching result for Java components


hdonnay:
So I think as written, this produces better results in some cases:
```
            - 		Name:    "com.google.inject",
            + 		Name:    "com.google.inject:guice",
```
but some lateral moves:
```
             + 		Name:           ":spring-aop",
```
and worse in others:
```
            + 		Name:           "Apache Software Foundation:org.apache.commons.jelly.tags.fmt",
```

* * *

Something we may not call out sufficiently in contributing docs: if you run the tests with the `integration` build tag once (e.g. `go test -tags integration github.com/quay/claircore/java/jar`) this with cache assets such that even normal runs will run the full suite of tests.

java/jar/jar.go
```diff
@@ -407,18 +407,48 @@ func (i *Info) parseManifest(ctx context.Context, r io.Reader) error {
...
+		artifactId = ""
+	}
+	if artifactId == groupId {
+		artifactId = ""
+	}
+	if len(groupId) > 0 && len(artifactId) > 0 {
+		name = groupId + ":" + artifactId
+	} else {
+		name = groupId + artifactId
+	}
```
hdonnay, line 450:
```suggestion
	// Trim to account for empty components.
	name = strings.Trim(groupId + ":" + artifactId, ":")
```

hdonnay (changes requested):
SGTM.
I think there's a better way to write that one bit, though.

The only thing that really needs to change is the commit message.
The commit should start with `jar: ...`. "Consolidation" means nothing when reading through the git log.

hdonnay:
Sorry to nitpick, but the subject shouldn't start capitalized. Running a command like `git log --oneline -5` should give nice, relatively uniform lines.

RTann:
@hdonnay does this merit a version bump of the package scanner?

daynewlee:
> Oh yeah, it totally does -- it'll change the reported names on some artifacts. Good catch

So I need to update this line `func (*Scanner) Version() string { return "3" }` to 4? cc: @RTann

RTann:
> > Oh yeah, it totally does -- it'll change the reported names on some artifacts. Good catch
> 
> So I need to update this line `func (*Scanner) Version() string { return "3" }` to 4 in the **packagescanner.go**? cc: @RTann

https://youtu.be/fvtQYsckLxk?t=65

hdonnay (changes requested):
Squash the second commit into the first and rebase :+1:

PR #861, opened 2023-03-07
all: audit `tmp.NewFile` use

hdonnay:
The fact that this idiom was applied unevenly means the interface is in need of retooling, IMO.

See-also: PROJQUAY-5165


crozzy (approved):
LGTM

PR #863, opened 2023-03-09
chore: delete airgap options

RTann:
At this point, these options are no longer used anywhere, so might as well remove them to ensure nobody uses them.


crozzy (dismissed):
My idea was to leave these badboys for a version or so for anyone wondering but I suppose the git history is evidence enough that these things existed.

RTann:
Just rebased. Will merge with CI

PR #867, opened 2023-03-13
libindex: Change location of layerScanner construction

crozzy:
The layerScanner should be constructed during libindex instantiation so the lifespan of the objects it creates are tied to the process as opposed to the request (as now).

- Instantiate layerScanner in the libindex.New function so it is scoped by process
- Remove LayerScanner interface that is unneeded
- Simplify ControllerFactory func
- As the layerScanner and indexer.Opts are instantiated in libvuln.New it no longer makes sense for the Realizer to be instantiated at the same time, just create the FetchArena as part of the controller and call the `Realizer()` method during indexing.


indexer/layerscanner.go
```diff
@@ -2,13 +2,244 @@ package indexer
...
+				Msg("unknown scanner kind")
+			continue
+		}
+
+		f, haveCfg := cfgMap[n]
+		if !haveCfg {
+			f = func(interface{}) error { return nil }
+		}
+		cs, csOK := interface{}(s).(ConfigurableScanner)
+		rs, rsOK := interface{}(s).(RPCScanner)
```
crozzy, line 84: `any` gives me typescript nightmares but is more concise

indexer/layerscanner.go
crozzy, line 1: This is more or less the old `indexer/layerscanner/layerscanner.go` with the generic function added.

hdonnay (dismissed):
This all looks right to me. I'm a little jumpy about touching it, but it does need to be done.

Worst case, we'll catch regressions against the staging Quayio corpus.

hdonnay (approved):

PR #868, opened 2023-03-13
etc: add image debugging script

hdonnay:
This takes a skopeo-formatted source and some tar patterns and extracts the layers in order. For example:

    etc/extract-layers docker://registry.access.redhat.com/ubi8/ubi:latest var/lib/dnf var/lib/rpm

### DIFF
--- a/etc/extract-layers
+++ b/etc/extract-layers
@@ -1,0 +1,34 @@
+#!/bin/sh
+if [ "$#" -lt 2 ]; then
+	printf 'missing "image name" and "tar pattern(s)" arguments\n' >&2
+	exit 99
+fi
+for cmd in jq skopeo; do
+	command -v "$cmd" >/dev/null 2>/dev/null || {
+		printf 'missing "%s"\n' "$cmd" >&2
+		exit 99
+	}
+done
+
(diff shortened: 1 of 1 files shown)


etc/extract-layers
```diff
@@ -0,0 +1,34 @@
...
+			sed 's,:,/,'
+	done |
+	while read -r l; do
+		d=$((n++))
+		mkdir -p "layer/$d"
+		echo ": $l"
+		tar --no-same-permissions -x -z -f "blobs/$l" -C "layer/$d" "$@" ||:
+	done
+)
+command -v tree >/dev/null 2>/dev/null && tree "${local}/layer"
```
crozzy, line 34:
```suggestion
command -v tree >/dev/null 2>/dev/null && tree -a "${local}/layer"
```
Add the hidden files?

crozzy (commented):
Great idea, I've been using `dive` but this is more CLI friendly

hdonnay:
yeah, I've managed to rewrite the oneliner(s) just about every time I've needed to do this, so it seems like I should just commit a script.

crozzy (approved):

Issue #809, opened 2023-02-02
False positives for packages installed from RH EUS streams 


bainsy88:
**Summary**

False positive is caused because Clair finds the security notice for RHEL 8 rather than for example RHEL 8.4 Extended support

**Example**

Openssl has a vulnerability [CVE-2022-0778 ](https://access.redhat.com/security/cve/CVE-2022-0778) which was fixed in the 8.4 extended support stream in [openssl-1.1.1g-16.el8_4.src.rpm](https://access.redhat.com/errata/RHSA-2022:1071). However Clair finds the RHEL 8 [security notice](https://access.redhat.com/errata/RHSA-2022:1065) which states the fix is in `openssl-1.1.1k-6.el8_5.src.rpm`.

As 1.1.1g-16 is deemed a lower version than openssl-1.1.1k-6 it is marked as vulnerble to the CVE. 

I think the core of issue is Clairs ability to find the correct security notice.

**Recreate** 

The following Dockerfile can be used to produce an image that contains many of these false positives 

```
FROM registry.access.redhat.com/ubi8/ubi:8.4

## Commented out as only needed if you building on a non RHEL box or one that doesn't have the correct subscriptions
#RUN --mount=type=secret,id=RHEL_SUBSCRIPTION_USER --mount=type=secret,id=RHEL_SUBSCRIPTION_PASSWORD \
    subscription-manager register --username=$(cat /run/secrets/RHEL_SUBSCRIPTION_USER) --password=$(cat /run/secrets/RHEL_SUBSCRIPTION_PASSWORD) --auto-attach

RUN dnf upgrade -y --repo=rhel-8-for-x86_64-baseos-eus-rpms --releasever=8.4
```

Interested in whether there is something we can do to help Clair determine the correct stream as a user or if this is a bug/missing feature on the Clair side

hdonnay:
I suspect this is because of claircore's (poor) heuristic for matching packages to repositories. I think claircore would need to interrogate dnf/yum databases to get this information.

iainduncani:
We'd also seen [this article](https://redhat-connect.gitbook.io/partner-guide-for-adopting-red-hat-oval-v2/determining-common-platform-enumeration-cpe) about determining the CPE for a package but installing packages from the EUS stream did not appear to alter the `/root/buildinfo/content_manifests`, if something was added to alter those files would Clair pick it up?

hdonnay:
[...]

PR #869, opened 2023-03-13
rpm: consult dnf database for repository information

hdonnay:
These changes make the `rhel` machinery consult both the `rpm` and `dnf` databases when possible.

See-also: #809


rpm/dnf.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+	}
+	if toOpen == "" {
+		// Nothing found.
+		return nil, nil
+	}
+	if fi, err := fs.Stat(sys, `root/buildinfo/content_manifests`); errors.Is(err, nil) && fi.IsDir() {
+		// This is a RHEL layer, skip.
+		return nil, nil
+	}
```
hdonnay, line 52: - [ ] Update this to check the files for a marker value and continue if found.

Should look for `from_dnf_hint` key in the yaml files.
See also: containerbuildsystem/atomic-reactor#2140

cgwalters:
I created https://github.com/rpm-software-management/rpm/discussions/3505 related to this

rpm/dnf.go
hdonnay: Punting on this unless we determine it's needed.

internal/rpm/find.go
```diff
@@ -0,0 +1,274 @@
...
+		dir, n := path.Split(p)
+		dir = path.Clean(dir)
+		db := FoundDB{
+			sys:  sys,
+			path: dir,
+		}
+		switch n {
+		case `Packages`:
+			db.kind = kindBDB
+			if ok, err := testpath(p, bdb.CheckMagic); !ok {
```
RTann, line 80: confirming we want to ignore `true, err != nil`?

internal/rpm/packages.go
```diff
@@ -0,0 +1,146 @@
...
+func PackagesFromDB(ctx context.Context, pkgdb string, db NativeDB) (iter.Seq[claircore.Package], func() error) {
+	defer trace.StartRegion(ctx, "PackagesFromDB").End()
+	var final error
+
+	seq := func(yield func(claircore.Package) bool) {
+		var err error
+		blobs, dbErr := db.All(ctx)
+		seq, parseErr := parseBlob(ctx, blobs)
+		defer func() {
+			final = errors.Join(err, parseErr(), dbErr(), db.Close())
```
RTann, line 62: just confirming we want to close the DB here?

internal/rpm/info.go
```diff
@@ -0,0 +1,220 @@
...
+	// TODO(hank) The blanket binary pattern is too broad and can miss things.
+	// Long-term, we should add pattern matching akin to [yara] or file(1) as a
+	// plugin mechanism that all indexers can use. That way, the Go indexer
+	// could register a pattern and use a shared filter over the
+	// [fs.WalkDirFunc] while this package (and dpkg, etc) can tell that another
+	// indexer will find those files relevant.
+	//
+	// [yara]: https://github.com/VirusTotal/yara
+	pat := []string{
+		`^.*/[^/]+\.jar$`, // Jar files
```
RTann, line 196: I know this is copied over, but I'm wondering if we should extend this to the other file extensions we support like WAR, EAR, etc?

RTann (commented):
made it through most of the files, but I'll need to come back to this another time. Would you mind making more commits and then squashing things at the end so it's easier to see the diffs after each round? Thanks

rhel/packagescanner.go
```diff
@@ -0,0 +1,94 @@
...
+
+	var final error
+	seq := func(yield func(*claircore.Package) bool) {
+		for _, db := range found {
+			ctx := zlog.ContextWithValues(ctx, "db", db.String())
+			zlog.Debug(ctx).Msg("examining database")
+
+			seq, checkPkgs := db.Packages(ctx)
+			seq, checkDNF := a.Wrap(ctx, seq)
+			wart.AsPointer(seq)(yield)
```
RTann, line 84: This was kind of a 🤯 for me, and I had to play with it in Go Playground to understand this: https://go.dev/play/p/j2-vMOOmrGw

rhel/packagescanner_test.go
```diff
@@ -0,0 +1,119 @@
...
+		got, err := s.Scan(ctx, &ls[0])
+		if err != nil {
+			t.Error(err)
+		}
+		t.Logf("found %d packages", len(got))
+		/*
+			for _, pkg := range got {
+				t.Logf("%s: %#q", pkg.Name, pkg.RepositoryHint)
+			}
+		*/
```
RTann, line 114: still desired or to delete?

rpm/packagescanner.go
```diff
@@ -30,21 +26,21 @@ var (
...
 // The zero value is ready to use.
 type Scanner struct{}
 
-// Name implements scanner.VersionedScanner.
+// Name implements [indexer.VersionedScanner].
 func (*Scanner) Name() string { return pkgName }
 
-// Version implements scanner.VersionedScanner.
+// Version implements [indexer.VersionedScanner].
 func (*Scanner) Version() string { return pkgVersion }
```
RTann, line 41: Just confirming: even though the code is very different, we aren't expecting different results, right?

rhel/repositoryscanner.go
```diff
@@ -204,24 +206,66 @@ func (r *RepositoryScanner) Scan(ctx context.Context, l *claircore.Layer) (repos
...
+			}
+		}
+
+		if r.apiFetcher != nil {
+			// Embedded content-sets are unavailable in very old images.
+			// For these, use fallback option and query Red Hat Container API.
+			ctx, done := context.WithTimeout(ctx, r.cfg.Timeout)
+			defer done()
+
+			// !!! This ends up being a weird nonlocal return. !!!
```
RTann, line 242: what does this mean?

rhel/repositoryscanner.go
```diff
@@ -204,24 +206,66 @@ func (r *RepositoryScanner) Scan(ctx context.Context, l *claircore.Layer) (repos
...
+				continue
+			}
+			for _, cpe := range cpes {
+				if !yield(repoid, cpe) {
+					return
+				}
+			}
+		}
+
+		if r.apiFetcher != nil {
```
RTann, line 236: let's say this is enabled. we seem to go here, no matter what, right? Before, we only got here if we were unable to read the content-sets files or the prod sec data is faulty.

internal/dnf/dnf.go
```diff
@@ -0,0 +1,330 @@
...
+					a.repo.Store(key, id)
+				case errors.Is(err, sql.ErrNoRows):
+					a.absent.Store(key, struct{}{})
+					if !yield(pkg) {
+						return
+					}
+					continue
+				default:
+					final = fmt.Errorf("internal/dnf: database error: %w", err)
+					return
```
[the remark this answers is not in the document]
crozzy: Yes the consumer will just see that at the end of the iteration (in this case there's also a consumable error so the reason for it should be known.

internal/dnf/dnf.go
```diff
@@ -0,0 +1,330 @@
...
+		return nil, nil, fmt.Errorf("internal/dnf: error reading sqlite db: %w", err)
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			zlog.Warn(ctx).Err(err).Msg("unable to close sqlite db file")
+		}
+	}()
+
+	zlog.Debug(ctx).Str("file", f.Name()).Msg("copying sqlite db out of tar")
+	if _, err := io.Copy(f, r); err != nil {
```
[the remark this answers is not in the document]
crozzy: I believe this copy needs to happen in order to make the file that gets passed to the `sql.Open` seekable (fulfill the `io.Seeker` interface). [This](https://github.com/quay/claircore/pull/720) is mentioned in the comment above.

internal/dnf/dnf.go
```diff
@@ -0,0 +1,330 @@
...
+//
+// Provided as a variable to allow callers to determine if the returned
+// [Annotator] will do anything.
+//
+// [Annotator.Close] is safe to call multiple times.
+var Identity Annotator
+
+func init() {
+	Identity = ident{}
+}
```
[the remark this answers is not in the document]
crozzy: I think the idea here is the variable can't be (accidentally or deliberately) reassigned

internal/dnf/dnf.go
```diff
@@ -0,0 +1,330 @@
...
+		removedEnum: enum,
+	}
+	a.db, a.spool, err = openDatabase(ctx, r)
+	if err != nil {
+		return nil, err
+	}
+
+	// Only way to get this is via the outer [NewAnnotator], so skip an extra
+	// frame.
+	_, file, line, _ := runtime.Caller(2)
```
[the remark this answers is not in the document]
crozzy: I believe so yes, it's skipping `newAnnotator` and `NewAnnotator` and reporting the caller or `NewAnnotator` which is what we're after.

rpm/packagescanner.go
```diff
@@ -69,214 +65,31 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
...
 	}
 
 	zlog.Debug(ctx).Int("count", len(found)).Msg("found possible databases")
 
-	var pkgs []*claircore.Package
-	done := map[string]struct{}{}
-	for _, db := range found {
-		ctx := zlog.ContextWithValues(ctx, "db", db.String())
-		zlog.Debug(ctx).Msg("examining database")
-		if _, ok := done[db.Path]; ok {
```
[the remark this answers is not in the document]
crozzy: That's what I read it as too, the `slices.SortFunc` is putting the `found` in order (so dbs with the same path are adjacent), then `slices.CompactFunc` is compacting them together, so by the time FindDBs returns the dupes should be taken care of

rhel/coalescer.go
```diff
@@ -116,10 +117,15 @@ func (*Coalescer) Coalesce(ctx context.Context, artifacts []*indexer.LayerArtifa
 					PackageDB:      pkg.PackageDB,
 					IntroducedIn:   layerArtifacts.Hash,
 					DistributionID: distID,
-					RepositoryIDs:  make([]string, len(layerArtifacts.Repos)),
 				}
-				for i := range layerArtifacts.Repos {
-					environment.RepositoryIDs[i] = layerArtifacts.Repos[i].ID
+				v, _ := url.ParseQuery(pkg.RepositoryHint)
+				if id := v.Get("repoid"); id != "" {
+					environment.RepositoryIDs = v["repoid"]
```
crozzy, line 123: This will leave inconsistencies in how the repo IDs are reported in index report. Packages with no specific repos will report their repo ids as numerical with the ability to be referenced back to `repository` objects (like current behaviour):
```
        "repository_ids": [
          "1",
          "2",
...
```
But packages that have a specific repoid specified by the DNF DB will have the repo slug that can not be directly referenced back to a `repository` object:
```
        "repository_ids": [
          "rhel-9-for-x86_64-appstream-rpms"
        ]
```
We should think of a way to make these consistent to keep the expectation that `"repository_ids"` will refer to keys in the `"repository"` section and not a mix of the `claircore.Repository.ID` and (what will be persisted as) the `claircore.Repository.Name`.

crozzy: Could do something like this:
```go
func lookupRepositoryID(repos []*claircore.Repository, slug string) string {
       for _, r := range repos {
               if r.Name == slug {
                       return r.ID
               }
       }
       return ""
}
```

and then
```go
if slug := v.Get("repoid"); slug != "" {
    if id := lookupRepositoryID(layerArtifacts.Repos, slug); id != "" {
         environment.RepositoryIDs = []string{id}
    }
```
This should always find a repo as repository scanner should be finding the same repos as the package scanner.

internal/rpm/nerva.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"strings"
+
+	"github.com/quay/claircore"
+)
+
+// NERVA returns an rpm NERVA for the package "p", assuming that the
+// [Package.Version] is an EVR string. This should hold true for all
+// [claircore.Package] instances returned by this package.
+func NERVA(p *claircore.Package) string {
```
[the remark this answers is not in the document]
crozzy: Yeah, I _think_ this is a spelling issue

internal/rpm/nerva.go
```diff
@@ -0,0 +1,20 @@
...
+	var b strings.Builder
+	b.WriteString(p.Name)
+	b.WriteByte('-')
+	b.WriteString(p.Version)
+	b.WriteByte('.')
+	b.WriteString(p.Arch)
+	return b.String()
```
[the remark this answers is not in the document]
crozzy: I think `strings.Builder` tends to be better allocations-wise with longer strings / many concatenations, I think in this case the advantages are probably not significant.

internal/rpm/info.go
crozzy: Yes, but maybe let's address that in https://github.com/quay/claircore/pull/1473

internal/wart/pointer_slice.go
```diff
@@ -0,0 +1,34 @@
...
+	return func(yield func(*T) bool) {
+		for v := range seq { // OK because of go1.22 loop changes.
+			if !yield(&v) {
+				return
+			}
+		}
+	}
+}
+
+// CollectPointer is the equivalent of "[slices.Collect]([AsPointer](seq))".
```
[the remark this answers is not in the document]
crozzy: I don't think this function is actually used

internal/rpm/find.go
crozzy: Yeah, `testpath` could return that if `fs.File.Close()` fails, I _think_ that case should be accounted for.

rhel/repositoryscanner.go
crozzy: I think it's saying that any error returned from the `mapContainerAPI` is not evaluated or returned until outside of this function body L#259

rhel/repositoryscanner.go
crozzy: Yes, it seems like `if len(repoids) == 0 && r.apiFetcher != nil` would be preferred, I can't see a situation where we'd want to reach out to the container API if we had content-sets either from DNF or the embedded file

rhel/coalescer.go
crozzy: This suggestion is incomplete in that we need to `append` all repositories who's `Name` == repoid, not just the first one as the example code does

internal/rpm/rpmdb/header.go
```diff
@@ -54,6 +54,7 @@ func ParseHeader(ctx context.Context, r io.ReaderAt) (*Header, error) {
 //
 // The ReaderAt must stay available throughout the lifetime of the Header.
 func (h *Header) Parse(ctx context.Context, r io.ReaderAt) error {
+	*h = Header{}
```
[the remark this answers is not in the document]
crozzy: I think this is essentially a `Reset`, not sure why it's needed now, I suspect it was just something noticed during porting it over.

internal/dnf/dnf.go
```diff
@@ -0,0 +1,330 @@
...
+		return nil, fmt.Errorf("internal/dnf: error reading rows: %w", err)
+	}
+
+	return ret, nil
+}
+
+func findDatabase(sys fs.FS) (toOpen string, rm int, err error) {
+	for i, p := range []string{
+		`usr/lib/sysimage/libdnf5/transaction_history.sqlite`, // dnf5 location
+		`var/lib/dnf/history.sqlite`,                          // dnf3/4 location
```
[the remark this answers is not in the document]
RTann: It was found UBI 9 deletes the dnf database, so I happened to choose the wrong image to test.

internal/dnf/dnf.go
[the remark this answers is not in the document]
hdonnay: The `sqlite` package (currently) needs a file on-disk that it can open by name.

internal/dnf/dnf.go
```diff
@@ -0,0 +1,330 @@
...
+// Close implements [Annotator].
+func (ident) Close() error {
+	return nil
+}
+
+// NewAnnotator holds book-keeping for producing multiple independent mapping
+// iterators for a given dnf history database.
+func newAnnotator(ctx context.Context, r fs.File, enum int) (*annotator, error) {
+	var err error
+	a := annotator{
```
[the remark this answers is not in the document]
hdonnay: I don't think so, I think it just grew that way.

internal/dnf/dnf.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+// Annotator holds the state for a mapping iterators that use a given dnf
+// database.
+type annotator struct {
+	spool *os.File
+	db    *sql.DB
+
+	// Concurrent maps for memoizing database lookups.
+	absent sync.Map
```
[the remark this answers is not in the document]
hdonnay: I think this was for clarity, yes. Meaning that the `repo` map doesn't store three states for a key (empty, "absent", present) but instead "absent" is denoted explicitly.

internal/dnf/dnf.go
```diff
@@ -0,0 +1,330 @@
...
+
+	"github.com/quay/claircore"
+)
+
+// BUG(hank) The dnf mapping is less useful than it could be because there's no
+// reliable way to turn the RepoID that it reports into something with meaning
+// outside of the Red Hat build system's builder's context. See [CLAIRDEV-45]
+// for more information.
+//
+// [CLAIRDEV-45]: https://issues.redhat.com/browse/CLAIRDEV-45
```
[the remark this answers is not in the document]
hdonnay: This is still true, AFAIK -- the "mapping" file is still needed.

internal/rpm/find.go
hdonnay: Failing to close the file should stop the process, IMO.

internal/rpm/find.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	slices.SortFunc(found, func(a, b FoundDB) int {
+		cmp := strings.Compare(a.path, b.path)
+		if cmp == 0 {
+			switch {
+			case a.kind < b.kind:
+				cmp = -1
+			case a.kind > b.kind:
+				cmp = +1
+			}
```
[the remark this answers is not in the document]
hdonnay: Yeah, this shouldn't happen; the extra compare just has a repeatable order if this case is hit. I'll replace it with a panic.

rpm/packagescanner.go
hdonnay: Right.

hdonnay:
Waiting on quay/claircore#1504 to land to (probably) fix CI.

internal/rpm/testdata/Info.Files.SQLite.txtar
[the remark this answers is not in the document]
hdonnay: yeah, I'll figure out where they need to be

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/quay/claircore
 
-go 1.23.0
+go 1.24.0
 
-toolchain go1.24.1
+toolchain go1.24.3
```
RTann, line 5: Not sure how required this is, but for the purposes of Clair and StackRox, this looks good to me. I believe this is the latest version of go1.24.3 available to us at this time

internal/rpm/bdb/bdb.go
```diff
@@ -168,87 +169,99 @@ type hashoffpage struct {
...
 			}
-			if k[0] != pagetypeHashOffIndex {
+			if h.Type != pagetypeHashUnsorted && h.Type != pagetypeHash {
 				continue
 			}
-			// Read the page header, now that we know it's correct.
-			var offpg hashoffpage
-			if err := binary.Read(view, db.ord, &offpg); err != nil {
-				return nil, fmt.Errorf("bdb: error reading hashoffpage: %w", err)
+			if h.Entries%2 != 0 {
```
RTann, line 188: I know this is just copied from the original implementation, but I am wondering if the compiler would optimize this or if it'd be faster if we explicitly did `h.Entries&1 != 0` instead

internal/rpm/bdb/bdb.go
```diff
@@ -168,87 +169,99 @@ type hashoffpage struct {
...
+				if !yield(nil, errors.New("bdb: odd number of entries")) {
+					return
+				}
+			}
+
+			ent := make([]hashentry, int(h.Entries)/2)
```
RTann, line 195: Going to make the same kind of comment, but perhaps it is better to keep it as-is and be explicit with a `/2` instead of `>1` operation

internal/rpm/bdb/bdb.go
```diff
@@ -168,87 +169,99 @@ type hashoffpage struct {
...
-			// First, check what kind of hash entry this is.
-			view := io.NewSectionReader(pg, off, hashoffpageSize)
-			if _, err := view.ReadAt(k, 0); err != nil {
-				return nil, fmt.Errorf("bdb: error peeking page type: %w", err)
+	return func(yield func(io.ReaderAt, error) bool) {
+		for n, lim := int64(0), int64(db.m.LastPageNo)+1; n < lim; n++ {
+			pg := io.NewSectionReader(db.r, n*pageSz, pageSz)
+			var h hashpage
+			if err := binary.Read(pg, db.ord, &h); err != nil {
+				if !yield(nil, fmt.Errorf("bdb: error reading hashpage: %w", err)) {
```
RTann, line 180: This is an example spot where I get a little bit confused about iterators. Let's say the caller opts to ignore the error here (ie. `yield() == true`): we just keep going even though the rest of this may fail or even panic?

internal/rpm/find.go
```diff
@@ -0,0 +1,271 @@
...
+			err = nil
+		}
+		return err
+	}
+
+	return seq, errFunc
+}
+
+func walk(ctx context.Context, out chan<- FoundDB, sys fs.FS) fs.WalkDirFunc {
+	handle := unique.Make(sys)
```
RTann, line 71: I've been meaning to read about this package, so thank you for encouraging me to do so

internal/rpm/find.go
```diff
@@ -0,0 +1,271 @@
...
+	return func(p string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			return nil
+		}
+
+		dir, n := path.Split(p)
```
RTann, line 80: I know this is copied over from before, but just curious why choose `path` instead of `filepath`?

internal/rpm/find.go
```diff
@@ -0,0 +1,271 @@
...
+				return err
+			}
+			ok := ndb.CheckMagic(ctx, f)
+			f.Close()
+			if !ok {
+				break
+			}
+			kind = kindNDB
+		}
+		if kind == 0 {
```
RTann, line 109: nit: since there is already the line there for `break` why not just remove this conditional here and turn those `break`s into `return nil`?

internal/rpm/info.go
```diff
@@ -0,0 +1,254 @@
...
+		case rpmdb.TagBasenames:
+			i.basename = v.([]string)
+		case rpmdb.TagFilenames:
+			// Filenames is the tag used in rpm4 -- this is a best-effort for
+			// supporting it.
+			//
+			// This takes the whole filenames value and splits it into an
+			// rpm5-style dir+base.
+			names := v.([]string)
+			println(i.Name, e.Tag.String(), names)
```
RTann: debugging?

internal/rpm/info.go
```diff
@@ -0,0 +1,254 @@
...
+			sort.Strings(names)
```
RTann: nit: `slices.Sort(names)`

rpm/packagescanner.go
```diff
@@ -63,32 +68,36 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
...
-			zlog.Debug(ctx).Msg("already seen, skipping")
-			continue
-		}
-		done[db.Path] = struct{}{}
-		ps, err := getDBObjects(ctx, sys, db, packagesFromDB)
+	var out []*claircore.Package
+	found, errFunc := rpm.FindDBs(ctx, sys)
```
RTann: nit: `found` -> `dbs` and then below `for found := range found` -> `for db := range dbs`

internal/rpm/files.go
```diff
@@ -0,0 +1,149 @@
...
+
+	return b.String()
+}
+
+// Len reports the size of the set.
+func (s *PathSet) len() int {
+	return len(s.paths)
+}
+
+// NoCopy is a ZST to trip the "copylocks" vet check.
```
RTann: what is ZST in this context?

internal/rpm/files.go
```diff
@@ -0,0 +1,149 @@
...
+	_, exists := s.paths[p]
+	return exists
+}
+
+// GoString implements [fmt.GoStringer].
+func (s *PathSet) GoString() string {
+	var b strings.Builder
+
+	b.WriteByte('[')
+	for k := range maps.Keys(s.paths) {
```
RTann: nit: `for k := range s.paths`

internal/rpm/files.go
```diff
@@ -0,0 +1,149 @@
...
+
+// PkgCache is a process-wide cache for [PathSet]s.
+//
+// This allows for all callers to use a single [PathSet] for a given layer.
+var pkgCache fileCache
+
+// FileCache implements a cache of [PathSet] values keyed by the layer digest.
+type fileCache struct {
+	// map[unique.Handle[string]][weak.Pointer[PathSet]]
+	m  sync.Map
```
RTann, line 27: I do wish there were a generic version to make it more obvious upon definition

internal/rpm/files.go
```diff
@@ -0,0 +1,149 @@
...
+	// that retrieves the original pointer needs to assume that it may have
+	// become nil and handle that case before dereferencing it.
+
+	key := unique.Make(layer.Hash.String())
+	if v, ok := c.m.Load(key); ok {
+		if s := v.(weak.Pointer[PathSet]).Value(); s != nil {
+			return s, nil
+		}
+	} else {
+		c.m.CompareAndDelete(key, v)
```
RTann: Can you explain why this is necessary? Is it just in case the value is added to the map between the time we check above and now? Would `v == nil` now?

internal/rpm/files.go
```diff
@@ -0,0 +1,149 @@
...
+	// This makes use of [weak.Pointer], which may be odd at first.
+	//
+	// The weak package has full documentation, but roughly: because a
+	// [weak.Pointer] does not keep a value alive for GC purposes, every use
+	// that retrieves the original pointer needs to assume that it may have
+	// become nil and handle that case before dereferencing it.
+
+	key := unique.Make(layer.Hash.String())
+	if v, ok := c.m.Load(key); ok {
+		if s := v.(weak.Pointer[PathSet]).Value(); s != nil {
```
RTann: what happens when `s == nil`? Is it assumed the `runtime.AddCleanup` got to it first?

internal/rpm/assertions.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/quay/claircore/internal/rpm/bdb"
+	"github.com/quay/claircore/internal/rpm/ndb"
+	"github.com/quay/claircore/internal/rpm/sqlite"
+)
+
+// Assert that the various database-specific packages implement [HeaderReader].
+var (
+	_ HeaderReader = (*sqlite.RPMDB)(nil)
```
RTann, line 11: nit: `PackageDB` for consistency?

rpm/ndb/package.go
```diff
@@ -67,6 +68,18 @@ func (db *PackageDB) Parse(r io.ReaderAt) error {
 	return nil
 }
 
+// Headers returns an iterator over all the header blobs in the database.
```
crozzy, line 71: Nit: Should `AllHeaders()` call `Headers()`? I suppose `slices.Collect` won't work with a `Seq2`

crozzy: Ahh, I just saw `tryCollect()` worth making them all consistent?

internal/rpm/database.go
```diff
@@ -0,0 +1,227 @@
...
+// LoadPackageInfo maps a sequence yielding package header bytes to a sequence
+// yielding package [Info] values.
+//
+// Any errors from the underlying sequence are passed through, and any errors
+// encountered are reported. Continuing after an error is reported starts
+// processing the next package header.
+func loadPackageInfo(ctx context.Context, headers iter.Seq2[io.ReaderAt, error]) iter.Seq2[Info, error] {
+	return func(yield func(Info, error) bool) {
+		var h rpmdb.Header
+		var z Info
```
RTann: can you comment why this is necessary? It looks like it's just meant as a blank info to return upon error. Perhaps we can rename it to make it clear?

crozzy (approved):
Looks good to me in the errm 3rd pass, just one nit

internal/rpm/database.go
```diff
@@ -0,0 +1,227 @@
...
+		seq := loadPackageInfo(ctx, headers)
+		var ok bool
+		ct := 0
+		defer func() {
+			zlog.Debug(ctx).
+				Int("packages", ct).
+				Msg("processed rpm db")
+		}()
+		retErr := func(err error) (cont bool) {
+			trace.WithRegion(ctx, "internal/rpm.Database.PackagesYield", func() { cont = yield(claircore.Package{}, err) })
```
RTann, line 53: `loacPackageInfo` seems to keep a blank package around to minimize potentially allocating multiple times due to error. Should we do the same here?

internal/rpm/database.go
```diff
@@ -0,0 +1,227 @@
...
+
+			pkg := claircore.Package{
+				Kind:           claircore.BINARY,
+				Name:           info.Name,
+				Arch:           info.Arch,
+				Module:         info.ModuleStream(),
+				RepositoryHint: info.Hint(),
+				PackageDB:      db.pkgdb,
+			}
+			v := info.NEVRA()
```
RTann, line 77: nit: is this variable used anywhere? If not, maybe just `pkg.Version = info.NEVRA().EVR()`

internal/rpm/database.go
```diff
@@ -0,0 +1,227 @@
...
+			pkg.Version = v.EVR()
+			printSourceVersionWarning(ctx)
+
+			srcRPM := info.SourceRPM
+			// Remove `.(no)src.rpm`
+			srcRPM = strings.TrimSuffix(srcRPM, ".rpm")
+			srcRPM = strings.TrimSuffix(srcRPM, ".src")
+			srcRPM = strings.TrimSuffix(srcRPM, ".nosrc")
+			pkg.Source, ok = srcs[srcRPM]
+			for !ok {
```
RTann, line 87: I get why this is done, but it does feel awkward. How would a closure sound?

internal/rpm/database.go
```diff
@@ -0,0 +1,227 @@
...
+				if v.Name == nil {
+					zlog.Info(ctx).
+						Msg("no name parse out of SOURCERPM tag, skipping")
+					break
+				}
+				src := claircore.Package{
+					Kind:    claircore.SOURCE,
+					Name:    *v.Name,
+					Version: v.EVR(),
+					// RepositoryHint: info.Hint(), // ???
```
RTann: Is this needed?

internal/dnf/dnf.go
```diff
@@ -0,0 +1,262 @@
...
+	Enum int
+}
+
+// Possible is the slice of dbDesc that's examined by openHistoryDB.
+var possible = []dbDesc{
+	// Defined here:
+	// https://github.com/rpm-software-management/dnf5/blob/13886935418e28482de7b675169482b85303845d/include/libdnf/transaction/transaction_item_action.hpp#L35
+	{Path: `usr/lib/sysimage/libdnf5/transaction_history.sqlite`, Enum: 5}, // dnf5
+	// Defined here:
+	// https://github.com/rpm-software-management/libdnf/blob/93759bc5cac262906e52b6a173d7b157914ec29e/libdnf/transaction/Types.hpp#L45
```
RTann: nit: I like https://github.com/rpm-software-management/libdnf/blob/0.74.0/libdnf/transaction/Types.hpp#L45 instead because it makes the tag clear

internal/dnf/dnf.go
```diff
@@ -0,0 +1,262 @@
...
+// denote a "removed" action.
+type dbDesc struct {
+	Path string
+	Enum int
+}
+
+// Possible is the slice of dbDesc that's examined by openHistoryDB.
+var possible = []dbDesc{
+	// Defined here:
+	// https://github.com/rpm-software-management/dnf5/blob/13886935418e28482de7b675169482b85303845d/include/libdnf/transaction/transaction_item_action.hpp#L35
```
RTann: latest version seems to be https://github.com/rpm-software-management/dnf5/blob/5.2.13.1/include/libdnf5/transaction/transaction_item_action.hpp#L37 (and this link includes the tag which I think is helpful)

RTann: the only reference to (most) of the path is here: https://github.com/rpm-software-management/dnf5/blob/692386325ac14141d969e6f762cf072bfd201646/dnf5.spec#L420

internal/dnf/dnf.go
RTann: Do we not want to link to https://github.com/rpm-software-management/libdnf/blob/0.74.0/libdnf/transaction/Swdb.hpp#L57 instead?

internal/dnf/dnf.go
```diff
@@ -0,0 +1,262 @@
...
+		default:
+			return nil, fmt.Errorf("internal/dnf: unexpected error handling fs.FS: %w", err)
+		}
+	}
+	if found == nil {
+		return nil, nil
+	}
+	var h *historyDB
+
+	f, err := sys.Open(found.Path)
```
RTann, line 155: can we just call this instead of `fs.Stat` earlier?

internal/dnf/dnf.go
```diff
@@ -0,0 +1,262 @@
...
+				zlog.Warn(ctx).Err(err).Msg("unable to unlink sqlite db file")
+			}
+		}
+	}()
+
+	zlog.Debug(ctx).Str("file", spool.Name()).Msg("copying sqlite db out of tar")
+	if _, err := io.Copy(spool, f); err != nil {
+		return nil, fmt.Errorf("internal/dnf: error spooling sqlite db: %w", err)
+	}
+	if err := spool.Sync(); err != nil {
```
RTann, line 187: just making sure *everything* is committed to disk before running the SQL queries?

internal/dnf/dnf.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	if err := db.PingContext(ctx); err != nil {
+		return nil, fmt.Errorf("internal/dnf: error reading sqlite db: %w", err)
+	}
+
+	h = &historyDB{
+		db: db,
+		rm: found.Enum,
+	}
+	runtime.AddCleanup(db, func(p string) {
```
RTann: just curious if any testing has shown this does get deleted eventually? If desired, the example in https://pkg.go.dev/runtime#AddCleanup shows a decent-ish way to verify it (perhaps only if we add some log, though)

internal/dnf/dnf.go
```diff
@@ -0,0 +1,262 @@
...
+		os.Remove(p)
+	}, spool.Name())
+	// This is an internal function, so add an extra caller frame.
+	_, file, line, _ := runtime.Caller(2)
+	runtime.SetFinalizer(h, func(_ *historyDB) {
+		panic(fmt.Sprintf("%s:%d: historyDB not closed", file, line))
```
RTann, line 206: idea: can we just `Close` for those who may forget?

internal/dnf/testdata/var/lib/dnf/history.sqlite
RTann, line 1: can we add test for dnf5, too?

internal/rpm/files.go
```diff
@@ -0,0 +1,149 @@
...
+		c.m.CompareAndDelete(key, wp)
+	}
+}
+
+// GetPathSet returns a [PathSet] for the provided layer, using a cached
+// version if possible.
+func (c *fileCache) GetPathSet(ctx context.Context, layer *claircore.Layer) (*PathSet, error) {
+	// This makes use of [weak.Pointer], which may be odd at first.
+	//
+	// The weak package has full documentation, but roughly: because a
```
RTann, line 42: The purpose of this is to simplify the implementation, right? Before, we had to track references (à la the FetchArena), but the weak pointer removes the need for this, as it'll tell us when nobody is using it anymore (when the `Value` is `nil`). Reading https://go.dev/blog/cleanups-and-weak#weak-pointers helped with this, too

internal/rpm/files.go
```diff
@@ -0,0 +1,149 @@
...
+	// [weak.Pointer] does not keep a value alive for GC purposes, every use
+	// that retrieves the original pointer needs to assume that it may have
+	// become nil and handle that case before dereferencing it.
+
+	key := unique.Make(layer.Hash.String())
+	if v, ok := c.m.Load(key); ok {
+		if s := v.(weak.Pointer[PathSet]).Value(); s != nil {
+			return s, nil
+		}
```
RTann: The example I saw in https://go.dev/blog/cleanups-and-weak#weak-pointers calls `cache.CompareAndDelete(filename, value)` here and then tries this all over again. I'm thinking we don't need this because we do end up overwriting the value later, but I just wanted to bring it up here. I'm mostly wondering if the old weak pointer will be deleted by the GC after we overwrite it

internal/rpm/files.go
```diff
@@ -0,0 +1,149 @@
...
+		seq, errFunc := FindDBs(ctx, sys)
+		defer func() {
+			err = errors.Join(err, errFunc())
+		}()
+
+		for found := range seq {
+			zlog.Debug(ctx).
+				Stringer("db", found).
+				Msg("found possible databases")
```
RTann: nit: `database`

internal/rpm/files.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// NewPathSet returns a [PathSet] for the provided layer.
+func NewPathSet(ctx context.Context, layer *claircore.Layer) (*PathSet, error) {
+	return pkgCache.GetPathSet(ctx, layer)
+}
+
+// PathSet is used to check if a path is an RPM-owned file.
+type PathSet struct {
+	_noCopy noCopy
```
RTann, line 127: Curious why this cannot be copied

---

Is it to prevent concurrent access to `paths` and remove the need for locking?

internal/rpm/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+			},
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
+		set, err := NewPathSet(ctx, &rl[0])
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer runtime.KeepAlive(set)
```
RTann, line 54: can you explain why this is here?

internal/dnf/dnf.go
hdonnay: Could, but it's inside a `for` scope, so I just opened the file in the desired scope after the check.

internal/rpm/bdb/bdb.go
hdonnay: Yeah, this would be the sort of change that warrants a benchmark

internal/rpm/bdb/bdb.go
hdonnay: I think I forgot a `continue`, will check

internal/dnf/dnf.go
hdonnay: Yeah, I'm not super happy with this and could be convinced to have the `historyDB` just take ownership of `spool`. Ultimately, we really should figure out a way to just use a file descriptor.

internal/dnf/dnf.go
hdonnay: No, the point of this is to help callers not forget.

internal/dnf/dnf.go
hdonnay: The spool file is unlinked in the `defer`, so this is probably overkill, anyway.

internal/rpm/assertions.go
hdonnay: Will create a follow-up for this.

internal/rpm/database.go
hdonnay: I think a closure would be more verbose for no real benefit. I could be convinced to rewrite it as an `if` with a `goto`, if that would be less weird.

internal/rpm/database.go
hdonnay: It probably can't hurt.

internal/rpm/database.go
hdonnay: Yeah, this is here just to avoid having type `Info{}` in multiple places. It's `z` for "zero". I'll just put the literals in.

internal/rpm/database.go
hdonnay: The `EVR` method takes a pointer receiver, so can't be chained onto a function returning a non-pointer.

internal/rpm/files.go
hdonnay: When `s == nil`, the pointed-to object has been garbage collected.

internal/rpm/files.go
hdonnay: I think this is leftover, will clean up.

internal/rpm/files.go
hdonnay: It's to enforce there only being one copy so that the `weak.Pointer` can work correctly. If a caller does something like
```
s1, _ := NewPathSet(ctx, layer)
s2 := *s1
```
then that would create another map referencing the same backing memory, preventing the value that `s1` points to from being collected.

internal/rpm/files.go
hdonnay: "Zero-sized type;" I'll fix the comment.

internal/rpm/files.go
hdonnay: Yeah, the `CompareAndDelete` should be the `else` of _this_ `if` statement. We don't need to loop because the `singleflight` is used to coordinate getting a fresh value.

internal/rpm/find.go
hdonnay: Using `filepath` would use the platform separator, but `fs.FS` only uses `/`.

internal/rpm/find.go
hdonnay: Yeah, I think this is a leftover from some refactoring -- I'll rework it.

internal/dnf/testdata/var/lib/dnf/history.sqlite
hdonnay: I don't know where to get one off the top of my head; will think about it.

rpm/ndb/package.go
hdonnay: `AllHeaders` gets removed in later commits.

internal/rpm/files_test.go
hdonnay: To keep `set` alive for this scope so the testcases are getting the cached value and the GC isn't collecting the cached value in a nondeterministic way.

test/rpmtest/options.go
```diff
@@ -42,13 +50,46 @@ func HintCompare(t testing.TB) cmp.Option {
...
 		}),
 	)
 }
 
+func isSubset[S ~[]E, E comparable](super, sub S) bool {
+	if super == nil {
+		return false
+	}
+	n := 0
+	for _, e := range sub {
```
RTann, line 89: up to you, but you can also short circuit before reaching this with `if len(super) < len(sub)`

rpm/packagescanner.go
```diff
@@ -24,27 +26,30 @@ var (
...
 func (*Scanner) Kind() string { return pkgKind }
 
 // Scan attempts to find rpm databases within the layer and enumerate the
 // packages there.
 //
 // A return of (nil, nil) is expected if there's no rpm database.
+//
+// Deprecated: In-tree [indexer.PackageScanner] implementations should almost
+// certainly use the "internal/rpm" and "internal/dnf" packages.
 func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircore.Package, error) {
```
RTann, line 53: Is this just kept around for compatibility purposes?

internal/dnf/dnf.go
```diff
@@ -0,0 +1,265 @@
...
+
+// BUG(hank) The dnf mapping is less useful than it could be because there's no
+// reliable way to turn the "repoid" that it reports into something with meaning
+// outside of the Red Hat build system's builder's context. See [CLAIRDEV-45]
+// for more information.
+//
+// [CLAIRDEV-45]: https://issues.redhat.com/browse/CLAIRDEV-45
+
+// PackageSeq is an alias for the Package iterator type.
+type PackageSeq = iter.Seq2[claircore.Package, error]
```
RTann, line 36: Forgot to ask before: why choose `claircore.Package` instead of `*claircore.Package`?

internal/dnf/dnf.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	wrapped := func(yield func(claircore.Package, error) bool) {
+		defer h.Close()
+
+		for pkg, err := range seq {
+			if err != nil {
+				if !yield(claircore.Package{}, err) {
```
RTann: Forgot to ask before: why not just pass `pkg` here, too?

internal/dnf/dnf.go
```diff
@@ -0,0 +1,265 @@
...
+// repoid, if possible.
+//
+// If any error is reported, the [claircore.Package] value is not modified.
+func (h *historyDB) AddRepoid(ctx context.Context, pkg *claircore.Package) error {
+	if h == nil {
+		return nil
+	}
+
+	// TODO(hank) Shouldn't need to build a string like this.
+	ver, err := rpmver.Parse(fmt.Sprintf("%s-%s.%s", pkg.Name, pkg.Version, pkg.Arch))
```
RTann: seems like we fill out `ver.Version`, `ver.Release`, and `ver.Archicture` (unless `pkg.Name` already has a `-`). Is this expected?

internal/dnf/dnf.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		// TODO(hank) Log?
```
RTann: I'm not opposed to at least a debug level log :)

internal/dnf/dnf.go
RTann: We also seem to assume the `epoch` is always `0`. I think I saw a reference to an issue about this somewhere. Can you mention is somewhere here, if that's what's happening here, too?

internal/rpm/bdb/bdb.go
```diff
@@ -284,9 +305,11 @@ func (r *rope) ReadAt(b []byte, off int64) (int, error) {
 			idx++
 			if idx != len(r.rd) {
 				rdoff = 0 // Reading from the start, now that we're on the next one.
-				continue
+				break     // May return EOF or nil on an exact-sized read, so hit the post-switch check.
```
RTann, line 308: We are essentially opting to return `err == nil` in this case, even if the underlying `io.ReaderAt` opts to `err == io.EOF`, right?

If `n != len(b)`, can we be stuck in an infinite loop? I guess it's effectively what we had before, and `idx` will keep increasing, but I want to confirm

internal/rpm/files.go
```diff
@@ -0,0 +1,152 @@
...
+			return s, nil
+		}
+		// Unable to upgrade: it's been garbage collected.
+		// Attempt to delete it so the next caller just has the load fail,
+		// then treat the key as novel.
+		c.m.CompareAndDelete(key, v)
+	}
+
+	ch := c.sf.DoChan(key.Value(), func() (any, error) {
+		set := &PathSet{
```
RTann: how would you feel about another `Load` over here in case we run into:

thread A calls `Load` on line 48 and doesn't find anything
thread B exits the singleflight and `Store`s the set
thread A now goes into the singleflight and does it all over again since it checked `Load` before B finished

RTann: Thinking about it some more, I do wonder if we would want to consider the `for` loop pattern instead of the `singleflight` pattern. I do see that's how the `unique` package is implemented

internal/rpm/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+		for _, tc := range tc.Testcases {
+			t.Run(tc.Name, func(t *testing.T) {
+				t.Parallel()
```
RTann, line 58: https://go.dev/doc/gc-guide#Testing_object_death suggests avoiding testing object death (perhaps also things related to object death) in parallel with other tests. Should we remove this line?

internal/rpm/rpm.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Validator is the extra interface an RPM database can implement if it needs
+// extra checks after opening.
+type validator interface {
+	Validate(context.Context) error
+}
+
+// Does what it says on the tin.
+func flat_map[T any, U any](seq iter.Seq[T], f func(T) (U, bool)) iter.Seq[U] {
```
RTann: Is this used anywhere?

internal/rpm/rpm.go
```diff
@@ -0,0 +1,37 @@
...
+// formats.
+package rpm
+
+import (
+	"context"
+	"io"
+	"iter"
+)
+
+const Version = "10"
```
RTann, line 11: confirming there are no changes to merit a version bump?

internal/rpm/rpmdb/tag_table.go
RTann, line 1: going to just assume this is right

rhel/coalescer.go
```diff
@@ -50,6 +49,10 @@ func (*Coalescer) Coalesce(ctx context.Context, artifacts []*indexer.LayerArtifa
 	}
 	// The same thing has to be done in reverse, because the first layer(s) are missing
 	// the relevant information.
+	//
+	// With [ENGCMP-5332], this shouldn't be needed, so check back in 5 years.
+	//
+	// [ENGCMP-5332]: https://issues.redhat.com/browse/ENGCMP-5332
```
RTann, line 55: looks like it's closed as of late April

rpm/packagescanner.go
hdonnay: Yes.

internal/dnf/dnf.go
hdonnay: That makes sense.

internal/dnf/dnf.go
hdonnay: I think the pervasive use of pointers to our domain types was a mistake because it makes reasoning about garbage harder.

internal/dnf/dnf.go
hdonnay: > seems like we fill out `ver.Version`, `ver.Release`, and `ver.Archicture` (unless `pkg.Name` already has a `-`). Is this expected?

Yes.

> We also seem to assume the `epoch` is always `0`. I think I saw a reference to an issue about this somewhere. Can you mention is somewhere here, if that's what's happening here, too?

The `Version` field should have the epoch.

internal/rpm/bdb/bdb.go
hdonnay: No, hitting EOF on the last segment will fallthrough to the error return.

internal/rpm/files.go
hdonnay: The `singleflight` is for work de-duplication, it's not an either-or, I suppose.

internal/rpm/files_test.go
hdonnay: It's not testing object death, it's just ensuring the cache is used.

internal/rpm/rpm.go
hdonnay: doesn't seem like

internal/rpm/rpm.go
hdonnay: Shouldn't be

rhel/coalescer.go
hdonnay: Right, so we should be able to assume it's everywhere in like 5 years

internal/dnf/dnf.go
RTann: for this particular case or in general? I see

```
type Version struct {
	Name         *string
	Architecture *string
	Epoch        string
	Version      string
	Release      string
}
```

internal/rpm/find.go
```diff
@@ -0,0 +1,271 @@
...
+	// variable and a channel -- but the API is nice.
+	eg, ctx := errgroup.WithContext(ctx)
+
+	ch := make(chan FoundDB)
+	eg.Go(func() error {
+		defer close(ch)
+		return fs.WalkDir(sys, ".", walk(ctx, ch, sys))
+	})
+	seq := func(yield func(FoundDB) bool) {
+		defer done(iterationDone)
```
RTann, line 40: This'll only have an effect if the context were not already cancelled via other means, right?

internal/rpm/find.go
RTann: what are your thoughts on making this the `default` instead?

internal/rpm/find.go
```diff
@@ -0,0 +1,271 @@
...
+	handle unique.Handle[fs.FS]
+}
+
+// String implements [fmt.Stringer].
+func (f FoundDB) String() string {
+	return f.kind.String() + ":" + f.path
+}
+
+// Filename constructs the path to the main database file.
+func (f FoundDB) filename() string {
```
RTann, line 140: why not just store this along with the rest of the path?

internal/rpm/find.go
```diff
@@ -0,0 +1,271 @@
...
+// OpenDB attempts to open the database pointed to by "found."
+func OpenDB(ctx context.Context, sys fs.FS, found FoundDB) (*Database, error) {
+	// Previous incarnations of this package ended up duplicating large chunks
+	// of logic around opening the databases. This pass attempts to fix that by
+	// using new go features.
+
+	switch {
+	case found.kind == 0:
+		return nil, errors.New("internal/rpm: programmer error: passed zero FoundDB")
+	case unique.Make(sys) != found.handle:
```
RTann, line 164: is the purpose of saving the handle in FoundDB for this sanity check? When would this happen?

internal/rpm/find.go
```diff
@@ -0,0 +1,271 @@
...
+
+	cleanup := &databaseCleanup{}
+	db := Database{
+		pkgdb:   found.path,
+		cleanup: cleanup,
+	}
+
+	spool, err := os.CreateTemp(os.TempDir(), fmt.Sprintf(`rpm.package.%v.`, found.kind))
+	if err != nil {
+		return nil, fmt.Errorf("rpm: error spooling db: %w", err)
```
RTann: I've seen a few instances of `rpm:` and `internal/rpm:` (and an instance where it's missing). Were some instances missed or was this on purpose?

internal/rpm/find.go
```diff
@@ -0,0 +1,271 @@
...
+type databaseCleanup struct {
+	spool *os.File
+	close func() error
+}
+
+func (c *databaseCleanup) Close() (err error) {
+	if c.spool != nil {
+		err = errors.Join(err, c.spool.Close(), os.Remove(c.spool.Name()))
+	}
+	if c.close != nil {
```
RTann, line 267: wondering if the order matters here. Since SQLite requires the file to be accessible, not sure if the SQLite's close will return an error if it's called after the file is deleted

internal/rpm/sqlite/sqlite.go
```diff
@@ -63,30 +64,35 @@ func (db *RPMDB) Close() error {
 	return db.db.Close()
 }
 
-// AllHeaders returns ReaderAts for all RPM headers in the database.
-func (db *RPMDB) AllHeaders(ctx context.Context) ([]io.ReaderAt, error) {
+// Headers returns all RPM headers in the database.
```
RTann: nit: `// Headers returns an iterator over all RPM headers in the RPMDB.`

internal/rpm/sqlite/sqlite.go
```diff
@@ -63,30 +64,35 @@ func (db *RPMDB) Close() error {
...
+func (db *RPMDB) Headers(ctx context.Context) iter.Seq2[io.ReaderAt, error] {
 	// Keys are sorted coming out of this query.
 	rows, err := db.db.QueryContext(ctx, allpackages)
```
RTann, line 70: just curious why this is outside the iterator

internal/rpm/info.go
```diff
@@ -0,0 +1,252 @@
...
+			i.Digest = v.([]string)[0]
+		case rpmdb.TagSigPGP:
+			i.Signature = v.([]byte)
+		case rpmdb.TagDirnames:
+			i.dirname = v.([]string)
+		case rpmdb.TagDirindexes:
+			i.dirindex = v.([]int32)
+		case rpmdb.TagBasenames:
+			i.basename = v.([]string)
+		case rpmdb.TagFilenames:
```
RTann, line 78: So this is rpm4, only, and the three previous tags are rpm5, only? Can you make it clear that we'll never (hopefully) run into a case where those three tags + this one exist in the same rpm DB?

internal/rpm/info.go
```diff
@@ -0,0 +1,252 @@
...
+			// Filenames is the tag used in rpm4 -- this is a best-effort for
+			// supporting it.
+			//
+			// This takes the whole filenames value and splits it into an
+			// rpm5-style dir+base.
+			names := v.([]string)
+			slices.Sort(names)
+			i.dirname = make([]string, 0)
```
RTann, line 87: is this needed?

internal/rpm/info.go
```diff
@@ -0,0 +1,252 @@
...
+	//
+	// The tested strings are absolute paths.
+	filePatterns = sync.OnceValue(func() *regexp.Regexp {
+		pat := []string{
+			`^.*/[^/]+\.[ejw]ar$`,                         // Jar files
+			`^.*/site-packages/[^/]+\.egg-info/PKG-INFO$`, // Python packages
+			`^.*/package.json$`,                           // npm packages
+			`^.*/[^/]+\.gemspec$`,                         // ruby gems
+			`^/usr/s?bin/[^/]+$`,                          // any executable
+			"^/usr/libexec/[^/]+/[^/]+$",                  // sometimes the executables are here too
```
RTann: nit: for consistency

```suggestion
			`^/usr/libexec/[^/]+/[^/]+$`,                  // sometimes the executables are here too
```

internal/rpm/info.go
```diff
@@ -0,0 +1,252 @@
...
+	for idx := range h.Infos {
+		e := &h.Infos[idx]
+		if _, ok := wantTags[e.Tag]; !ok {
+			continue
+		}
+		v, err := h.ReadData(ctx, e)
+		if err != nil {
+			return err
+		}
+		switch e.Tag {
```
RTann, line 51: nit: maybe add a default to catch programmer error (update to `wantTags` without adding a case for it)

internal/rpm/info.go
```diff
@@ -0,0 +1,252 @@
...
+		if pat.MatchString(p) {
+			s.paths[p] = struct{}{}
+		}
+	}
+}
+
+// NEVRA constructs a Name-Epoch-Version-Release-Architecture [rpmver.Version].
+func (i *Info) NEVRA() rpmver.Version {
+	return rpmver.Version{
+		Name:         &([]string{i.Name})[0],
```
RTann: can you explain the purpose of this?

internal/rpm/info.go
```diff
@@ -0,0 +1,252 @@
...
+		Release:      i.Release,
+	}
+}
+
+// Hint constructs a string suitable to be used as the "RepositoryHint".
+func (i *Info) Hint() string {
+	v := url.Values{}
+	if i.Digest != "" {
+		switch i.DigestAlgo {
+		case 8:
```
RTann, line 167: is this specified in both rpm4 and 5?

internal/rpm/database.go
```diff
@@ -0,0 +1,228 @@
...
+
+		for r, err := range headers {
+			if err != nil {
+				if !yield(Info{}, fmt.Errorf("rpm: unable to read header: %w", err)) {
+					return
+				}
+				continue
+			}
+
+			h = rpmdb.Header{}
```
RTann, line 202: I recall asking about something similar, but I don't remember where it was in this PR: why not just `var h rpmdb.Header` here?

internal/rpm/database.go
```diff
@@ -0,0 +1,228 @@
...
+			srcRPM = strings.TrimSuffix(srcRPM, ".src")
+			srcRPM = strings.TrimSuffix(srcRPM, ".nosrc")
+			pkg.Source, ok = srcs[srcRPM]
+			for !ok {
+				v, err := rpmver.Parse(srcRPM)
+				if err != nil {
+					zlog.Info(ctx).
+						Err(err).
+						Msg("unable to parse SOURCERPM tag, skipping")
+					break
```
RTann, line 93: so instead of erroring out the whole thing, we just ignore it and keep going?

internal/dnf/dnf.go
RTann: I think I recall another function elsewhere returning `iter.Seq2[claircore.Package, error]`. Do we want to consider putting this somewhere more generic or just keeping things as-is?

internal/dnf/dnf.go
```diff
@@ -0,0 +1,272 @@
...
+
+// OpenHistoryDB does what it says on the tin.
+//
+// This function may return a nil *historyDB, which is still safe to use.
+func openHistoryDB(ctx context.Context, sys fs.FS) (*historyDB, error) {
+	var found *dbDesc
+	for _, v := range possible {
+		switch _, err := fs.Stat(sys, v.Path); {
+		case errors.Is(err, nil):
+			found = &v
```
RTann, line 143: do we want to short-circuit here, too?

rhel/coalescer_test.go
```diff
@@ -2,340 +2,466 @@ package rhel
...
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/quay/zlog"
 
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/indexer"
 	"github.com/quay/claircore/test"
+	"github.com/quay/claircore/toolkit/types/cpe"
```
RTann: not sure if this belongs here or the other section because this is from a different module

rhel/coalescer_test.go
```diff
@@ -2,340 +2,466 @@ package rhel
...
+		opts := cmp.Options{cmpopts.IgnoreUnexported(claircore.Digest{})}
+		if !cmp.Equal(got, want, opts) {
+			t.Error(cmp.Diff(got, want, opts))
 		}
-	}
+	})
 }
 
-func TestCoalescerUpdatedPackage(t *testing.T) {
+// TestCoalescer tests the private method coalesce on the [Coalescer].
```
RTann: private?

rhel/repositoryscanner.go
```diff
@@ -204,49 +209,126 @@ func (r *RepositoryScanner) Scan(ctx context.Context, l *claircore.Layer) (repos
...
-// MapContentSets returns a slice of CPEs bound into strings, as discovered by
-// examining information contained within the container.
-func mapContentSets(ctx context.Context, sys fs.FS, cm *mappingFile) ([]string, error) {
-	// Get CPEs using embedded content-set files.
-	// The files are stored in /root/buildinfo/content_manifests/ and will need to
-	// be translated using mapping file provided by Red Hat's PST team.
-	// For RHCOS, the files are stored in /usr/share/buildinfo/.
+// BugURL constructs a link directly to the Red Hat Jira instance.
+func bugURL(id string, err error) string {
+	const desc = "A Clair instance noticed an invalid CPE:{code}%s{code}\nThe reported error was:{code}%v{code}"
```
RTann, line 312: Clair isn't the only client 😄

rhel/repositoryscanner.go
```diff
@@ -299,27 +381,16 @@ type repo struct {
...
-				Str("repository", r).
-				Msg("repository not present in a mapping file")
-			continue
-		}
-		for _, cpe := range cpes.CPEs {
-			s[cpe] = struct{}{}
-		}
+// GetOne takes a repoid and reports the CPEs and if the repoid was known
+// beforehand.
+func (m *mappingFile) GetOne(ctx context.Context, repoid string) (cpes []string, ok bool) {
```
RTann, line 389: nit: the named returns aren't really used, so we can delete them

internal/dnf/dnf.go
hdonnay: They're different types

rhel/repositoryscanner.go
```diff
@@ -204,49 +209,126 @@ func (r *RepositoryScanner) Scan(ctx context.Context, l *claircore.Layer) (repos
...
+			// Embedded content-sets are unavailable in very old images.
+			// For these, use fallback option and query Red Hat Container API.
+			ctx, done := context.WithTimeout(ctx, r.cfg.Timeout)
+			defer done()
+
+			// !!! This ends up being a weird nonlocal return. !!!
+			var cpes []string
+			cpes, err = mapContainerAPI(ctx, sys, r.apiFetcher)
+			if err != nil {
+				return
```
RTann, line 257: do we want to log here?

rhel/repositoryscanner.go
```diff
@@ -204,49 +209,126 @@ func (r *RepositoryScanner) Scan(ctx context.Context, l *claircore.Layer) (repos
...
-				Str("url", `https://bugzilla.redhat.com/enter_bug.cgi?product=Container%20Factory`).
+				Str("url", bugURL(cpeID, err)).
 				Str("cpeID", cpeID).
 				Msg("invalid CPE, please report a bug upstream")
 			continue
 		}
 
 		repositories = append(repositories, r)
 	}
+	slices.SortFunc(repositories, func(a, b *claircore.Repository) int {
```
RTann, line 294: can you explain the purpose of this?

RTann (commented):
I think I finally got through it

internal/rpm/find.go
hdonnay: Yes

internal/rpm/find.go
hdonnay: Yeah, like the error message says, it's to catch a programmer error.

internal/rpm/find.go
hdonnay: Shouldn't be a problem, because the file is already open

internal/rpm/info.go
hdonnay: yes

internal/rpm/find.go
hdonnay: Just missed, I'll give them a once-over

internal/rpm/info.go
hdonnay: I suppose not, as it's always just appended to.

internal/rpm/info.go
hdonnay: This is a construction to get a pointer. It's needed in some cases because the string header exists such that it's not allowed to take a pointer to it. I think this is an artifact of evolution because after checking, that doesn't seem to be the case here.

internal/rpm/database.go
hdonnay: This zeroes the `Header`, because the iterator uses the same variable for every header it parses. Declaring the variable outside the loop is just a style preference, I suppose.

internal/rpm/database.go
hdonnay: Yes. All the source-package handling is best effort until the new tag is used.

rhel/repositoryscanner.go
hdonnay: It removes duplicated repositories. This can apparently happen with the Rube Goldberg machine that is the build system.

rhel/repositoryscanner.go
hdonnay: I added the named return(s) to document that the returned strings will be CPEs.

rhel/coalescer_test.go
hdonnay: Yeah, I dunno, that's just what was in the comment already.

internal/rpm/find.go
hdonnay: This is just the way it worked out. The name of the file is based on the `Kind`, the `Kind` needs to be tracked, so building the full path on demand is easy enough.

internal/rpm/sqlite/sqlite.go
hdonnay: No particular reason

internal/rpm/find.go
RTann: isn't it closed and deleted by this point?

crozzy (approved):
LGTM

PR #870, opened 2023-03-14
linux: move package to root

crozzy:
It doesn't really make sense for an "implementer" to exist inside the indexer/ dir so in the spirit of cleaning up indexer/ move the linux specific package.


RTann (approved):
Makes sense to me

hdonnay (approved):

crozzy:
Clean rebase

PR #879, opened 2023-03-20
chore: do not assume challenge endpoint called token

RTann:
I ran into this when I was testing out images pulled from quay.io. The returned `realm` for that was `https://quay.io/v2/auth`

### DIFF
--- a/test/fetch/layer.go
+++ b/test/fetch/layer.go
@@ -154,10 +154,6 @@ func (d *client) doAuth(ctx context.Context, c *http.Client, name, h string) err
 	if err != nil {
 		return err
 	}
-	u, err = u.Parse("token")
-	if err != nil {
-		return err
-	}
 	v := url.Values{
 		"service": {attrs["service"]},
 		"scope":   {attrs["scope"]},


crozzy:
I'm sort of surprised by this, is quay.io the only registry we've tried to pull from with auth?

RTann:
> I'm sort of surprised by this, is quay.io the only registry we've tried to pull from with auth?

with `auth` instead of `token`? I just checked, and the instances of `gcr.io` and `registry.access.redhat.com` did not require any auth (I ran `curl` and got a 200 without a `www-authenticate` header). I only saw auth required for `docker.io`. I did for `quay.io` at one point, but I think that was only when I was using the incorrect path. I don't see it required anymore now that I fixed the path. 

```
// Correct
$ curl https://quay.io/v2/projectquay/clair-fixtures/blobs/sha256:8cf469d41e77c8c4aaa2191f42f55b9758f77c640f2f7015b799b26458903f9b --head
HTTP/2 200 
date: Mon, 20 Mar 2023 23:04:22 GMT
content-type: application/octet-stream
content-length: 15212591
server: nginx/1.20.1
docker-content-digest: sha256:8cf469d41e77c8c4aaa2191f42f55b9758f77c640f2f7015b799b26458903f9b
cache-control: max-age=31436000
x-frame-options: DENY
strict-transport-security: max-age=63072000; preload

// Incorrect
$ curl https://quay.io/v2/projectquay/clair-fixtures:test/blobs/sha256:8cf469d41e77c8c4aaa2191f42f55b9758f77c640f2f7015b799b26458903f9b --head
HTTP/2 401 
date: Mon, 20 Mar 2023 23:04:29 GMT
content-type: application/json
content-length: 112
server: nginx/1.20.1
www-authenticate: Bearer realm="https://quay.io/v2/auth",service="quay.io",scope="repository:projectquay/clair-fixtures%3Atest:pull"
docker-distribution-api-version: registry/2.0
```

I also tried giving bad resources to `registry.access.redhat.com` and `gcr.io`, but neither gave me `www-authenticate`

RTann:
So I guess this PR wasn't needed because auth is not required when I used the correct link. However, maybe it'd be helpful in the future

crozzy (approved):
Seems like something we want to not come up against this in the future

PR #882, opened 2023-03-27
tests: Rename testdata to .testdata 

crozzy:
We recently had an issue where test data file names weren't
compatible with the go tools. That led to the decision to stop any
testdata from being pulled when using claircore as a dep. This should
cut-down on final image size too.


hdonnay:
I'm not clear on what this is supposed to fix. What's the actual error? The `testdata` directory is specifically recognized by the `go` tool for the purpose of holding ancillary data.

crozzy:
The error is that the claircore library couldn't be imported but my fault as I didn't know `testdata` was a specific case just thought `_` and `.` proceeding was ignored. I guess it doesn't matter if it's ignored in the end as the `git archive` with `--format=zip` will consider the entire fs before go does it's trimming. Replacing with this: https://github.com/quay/claircore/pull/883

PR #886, opened 2023-03-29
Parse OSV maven data and complete java support

daynewlee:
Enable OSV maven
Match CVE data with affected packages


RTann:
@daynewlee can you base the PR on Hank's branch so we can see the individual changes your PR makes?

PR #903, opened 2023-04-12
python: migrate Python support to OSV

daynewlee:


hdonnay:
before you get too far down this path, `github.com/quay/claircore/pkg/pep440` exists.

PR #908, opened 2023-04-14
etc: add import helper

hdonnay:


crozzy (approved):

crozzy:
Do we want to add something like this as discussed:
```yaml
  import-check:
    needs: ['config']
    name: Import Check
    runs-on: ubuntu-latest
    container: quay.io/projectquay/golang:${{ fromJSON(needs.config.outputs.go_versions)[0] }}
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Go Imports
        run: ./etc/importfix.sh
      - name: Git Diff
        run: git diff --exit-code
```

hdonnay:
Do you think that's worth putting in this PR?

crozzy:
It seems like at the moment things are pretty much in line as far as existing code save for two minor formatting issues, do we want to just fix them and then do a full `goimports -d $(find . -type f -name '*.go' -not -path "./vendor/*")` every time?

hdonnay:
going to rebase and merge

PR #911, opened 2023-04-18
indexer: Add index to manifest_index table to aid deletions

crozzy:
By deleting a broad index for space reasons (
https://github.com/quay/claircore/commit/d9abe3db42bb533962ff163b3521587fc4b3f0cc) I inadvertantly slowed the delete manifests query due to a delete cascading to the manifest_index table and having to do a seq scan. This change adds a dedicated index for queries involving manifest_id.

### DIFF
--- a/datastore/postgres/deletemanifests_test.go
+++ b/datastore/postgres/deletemanifests_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/quay/zlog"
 
 	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
 	"github.com/quay/claircore/test"
 	"github.com/quay/claircore/test/integration"
 	pgtest "github.com/quay/claircore/test/postgres"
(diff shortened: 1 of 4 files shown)


crozzy:
Query log after adding index:
```
2023-04-18 14:41:37.556 PDT [2525748] LOG:  duration: 0.006 ms  plan:
	Query Text: DELETE FROM ONLY "public"."manifest_index" WHERE $1 OPERATOR(pg_catalog.=) "manifest_id"
	Delete on public.manifest_index  (cost=4.20..13.67 rows=0 width=0) (actual time=0.005..0.005 rows=0 loops=1)
	  Buffers: shared hit=2
	  ->  Bitmap Heap Scan on public.manifest_index  (cost=4.20..13.67 rows=6 width=6) (actual time=0.004..0.005 rows=0 loops=1)
	        Output: ctid
	        Recheck Cond: ('1'::bigint = manifest_index.manifest_id)
	        Buffers: shared hit=2
	        ->  Bitmap Index Scan on idx_manifest_index_manifest_id  (cost=0.00..4.20 rows=6 width=0) (actual time=0.002..0.002 rows=0 loops=1)
	              Index Cond: (manifest_index.manifest_id = '1'::bigint)
	              Buffers: shared hit=2
```

hdonnay (changes requested):
Could you please split the change to the integration package to another commit?

hdonnay (dismissed):
LGTM

hdonnay (approved):

PR #915, opened 2023-04-24
python: migrate data source to OSV

daynewlee:


RTann:
I'm cool with it, but up to @crozzy and @hdonnay if pyup.io support should be removed. If we do end up removing it, then let's add the `needs-changelog` label to this PR

python/matcher.go
```diff
@@ -19,40 +19,54 @@ type Matcher struct{}
 // Name implements driver.Matcher.
 func (*Matcher) Name() string { return "python" }
 
-// Filter implements driver.Matcher.
-func (*Matcher) Filter(record *claircore.IndexRecord) bool {
-	return record.Package.NormalizedVersion.Kind == "pep440"
+func (*Matcher) Filter(r *claircore.IndexRecord) bool {
```
RTann, line 22: I think the `pep440` check still makes sense for this, but I'm not 100% sure. I'll defer this to someone else

python/matcher.go
```diff
@@ -19,40 +19,54 @@ type Matcher struct{}
...
-		zlog.Warn(ctx).
-			Str("package", record.Package.Name).
-			Stringer("version", &v).
-			Msg("unable to parse python package version")
-		return false, nil
+		return false, err
+	}
+	upperVersion := decodedVersions.Get("fixed")
+	if upperVersion == "" {
+		return false, fmt.Errorf("pypi: missing upper version")
```
RTann, line 44: I'm not sure if pypi makes sense here. wouldn't this be an osv problem?

test/periodic/updater_test.go
RTann, line 1: Should we replace the removed one with and OSV test?

python/testdata/indexreport-buster-jackson-databind.json
RTann, line 1: can we rename this to rhel8 or something? this is definitely not buster (debian 10)

python/matcher_test.go
```diff
@@ -2,6 +2,7 @@ package python_test
 
 import (
 	"context"
+	"github.com/google/go-cmp/cmp"
```
RTann, line 5: nit: put this in the block directly below with zlog

python/matcher_test.go
```diff
@@ -39,7 +40,7 @@ func (tc matcherTestcase) Run(t *testing.T) {
 
 // TestMatcher tests the python matcher.
 func TestMatcher(t *testing.T) {
-	tt := []matcherTestcase{
+	testcases := []matcherTestcase{
```
RTann, line 43: can you add a test case for a package version less than the "introduced" version?

python/matcher.go
```diff
@@ -19,40 +19,54 @@ type Matcher struct{}
...
+	}
+
+	rv, err := pep440.Parse(record.Package.Version)
+	if err != nil {
+		return false, err
 	}
-	spec, err := pep440.ParseRange(vuln.Package.Version)
+
+	v2, err := pep440.Parse(upperVersion)
```
RTann, line 52: nit: maybe `uv`?

python/matcher.go
```diff
@@ -2,8 +2,8 @@ package python
 
 import (
 	"context"
-
-	"github.com/quay/zlog"
```
RTann, line 6: let's keep the logs. It can be helpful to log any parsing failures

updater/osv/osv.go
```diff
@@ -594,12 +598,13 @@ func (e *ecs) Insert(ctx context.Context, name string, a *advisory) (err error)
 				vs = b.String()
 			}
 			pkgName := af.Package.PURL
-			if af.Package.Ecosystem == ecosystemMaven {
+			if af.Package.Ecosystem == ecosystemMaven || af.Package.Ecosystem == ecosystemPyPI {
```
RTann, line 601: nit: maybe a switch here, too? might be more readable and extensible than the if in this context

python/matcher.go
hdonnay: Yeah, I think it still does, even if only in a belt-and-suspenders way.

python/testdata/indexreport-buster-rhel8-databind.json
hdonnay, line 1: What's up with this name? Why "buster" and "rhel8" -- is it both Debian and RHEL? What does "databind" mean, here?

hdonnay (changes requested):
The first commit needs to be split into adding functionality to `osv` and removing `pyupio`.

We'll need a follow-up to clean up the leaked vulnerabilities, after this.

python/matcher_test.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/quay/zlog"
 
+	"github.com/google/go-cmp/cmp"
```
hdonnay, line 9: Fix grouping -- once rebased on latest `main`, `git rebase -x etc/importfix.sh origin/main` should do it

hdonnay (changes requested):
LGTM after the one formatting nit.

daynewlee:
Applied changes in etc/importfix.sh, then ran:
git rebase -x './etc/importfix.sh.patched; git diff --quiet || git commit -a -n --no-edit --fixup=HEAD' upstream/main

python/matcher.go
```diff
@@ -19,40 +21,65 @@ type Matcher struct{}
 // Name implements driver.Matcher.
 func (*Matcher) Name() string { return "python" }
 
-// Filter implements driver.Matcher.
-func (*Matcher) Filter(record *claircore.IndexRecord) bool {
-	return record.Package.NormalizedVersion.Kind == "pep440"
+func (*Matcher) Filter(r *claircore.IndexRecord) bool {
```
RTann, line 24: Let's just completely revert this part back since no update is needed

python/matcher.go
```diff
@@ -19,40 +21,65 @@ type Matcher struct{}
...
+		return true, nil
+	}
+
+	decodedVersions, err := url.ParseQuery(vuln.FixedInVersion)
+	if err != nil {
+		return false, err
+	}
+	upperVersion := decodedVersions.Get("fixed")
+	if upperVersion == "" {
+		return false, fmt.Errorf("osv: missing upper version")
```
RTann, line 45: thinking about it, maybe this should say `python` instead of `osv` because, even though the data is OSV, the function is under the `python` directory. I'll let someone else chime in about this because I'm not 100% sure the convention used in this repo

crozzy: agree

hdonnay (approved):
LGTM

PR #919, opened 2023-04-27
rhel: use CVSS score as Severity

RTann:
Red Hat rescores vulnerabilities all the time, so the scores found by NVD for vulnerabilities will differ from Red Hat's findings for packages inside of RHEL-based images.

This PR replaces the un-normalized Severity with an applicable CVSS score.

When it comes to multiple CVSS versions, prefer CVSSv3 over CVSSv2.
When it comes to advisories, opt to use the highest CVSS score among all related CVEs.

### DIFF
--- a/rhel/parser.go
+++ b/rhel/parser.go
@@ -5,6 +5,8 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
+	"strconv"
+	"strings"
 
 	"github.com/quay/goval-parser/oval"
 	"github.com/quay/zlog"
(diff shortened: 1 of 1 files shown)


hdonnay:
What's the reasoning for picking one of the "child" advisories instead of the severity assigned by Red Hat? I don't follow how NVD matters -- they do their own scoring that vendors frequently disagree with.

RTann:
> What's the reasoning for picking one of the "child" advisories instead of the severity assigned by Red Hat? I don't follow how NVD matters -- they do their own scoring that vendors frequently disagree with.

I think we talked about this before. ACS needs the CVSS scores for vulnerabilities, which, for the most part, is covered by NVD. The issue is that Red Hat rescores vulnerabilities, and people want to see Red Hat's scores (we've had these issues come up in the past, and only for Red Hat. Nobody has cared about other vendors' scores from what I can tell). From what I can remember, we said replacing the severity with the CVSS score was ok, but perhaps I am mistaken.

When it comes to RHSAs, ACS has opted to just choose the highest CVSS score among all the related CVEs. Technically, RHSAs do not have CVSS scores, just severity, but we opted for this approach because we show RHSAs as the top-level vulnerability instead of the CVE. We can discuss more offline, if needed

crozzy:
Replacing the Severity does seem like it's going to land us in hot-water for people expecting the RHSA severity. This information seems to be a better fit as an `enrichment` (in fact this is exactly what the current enrichments implementation does), the issue would be how/where to drop the hints for the enricher considering the fields it searches (`name`, `description`, `links`) are all public. I suppose if we added the NVD CVE links during the `Parse()` it'd just work and ACS (or any consumer) could choose the advisory severity from the vulnerability (as today) or the NVD severity from the vulnerabilities enrichment.

PR #922, opened 2023-05-01
osv: canonicalize ranges

hdonnay:
This commit makes the OSV updater actually bother to fix non-canonical version ranges, instead of just complaining about them into the log.

### DIFF
--- a/updater/osv/osv.go
+++ b/updater/osv/osv.go
@@ -373,10 +373,7 @@ func (u *updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
 		}
 		name := strings.TrimSuffix(path.Base(zf.Name), ".zip")
 
-		var skipped struct {
-			Withdrawn  []string
-			Unaffected []string
-		}
+		var skipped stats
 		var ct int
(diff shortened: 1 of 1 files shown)


crozzy (commented):
I think we discussed throwing these out until we get reasonable data?

updater/osv/osv.go
```diff
@@ -626,6 +627,34 @@ func (e *ecs) NewVulnerability() *claircore.Vulnerability {
 	return &e.Vulnerability[i]
 }
 
+// RemoveVulnerability does what it says on the tin.
+//
+// Will cause copying if the vulnerability is not the most recent returned from
+// NewVulnerability.
+func (e *ecs) RemoveVulnerability(v *claircore.Vulnerability) {
```
crozzy, line 634: Does this need to be public?

crozzy (commented):
LGTM, just one nit

updater/osv/osv.go
hdonnay: The receiver type is unexported, so this just keeps with the convention.

crozzy (approved):

PR #924, opened 2023-05-04
etc: make importfix script ignore deleted files

hdonnay:


hdonnay:
Now with fixes for the OSX userspace.

crozzy (approved):

PR #926, opened 2023-05-04
chore: use non-containerized postgres and go

hdonnay:
This is an attempt to speed up CI, now that we don't have any external process dependencies.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,10 +39,12 @@ jobs:
     needs: ['config']
     name: Tidy
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -107,4 +98,4 @@ jobs:
             **/testdata/*.tar
             **/testdata/*.tar.gz
       - name: Tests
-        run: make integration
+        run: go test -race -tags integration ./...
```
crozzy, line 101: What is the idea here? Enable the cache?

hdonnay: Yeah, to enable the cache. It'll get blown away on every toolchain or dependency upgrade.

crozzy (approved):

PR #934, opened 2023-05-11
integration: change test cache dir

hdonnay:
This patchset moves the cache of integration assets out-of-tree and into standardized cache paths. This should eventually help with CI caching.


test/fetch/layer.go
```diff
@@ -37,7 +38,9 @@ func Layer(ctx context.Context, t *testing.T, c *http.Client, from, repo string,
 	for _, o := range opt {
 		opts[o] = true
 	}
-	cachefile := filepath.Join("testdata", blob.String()+".layer")
+	// Splitting the digest like this future-proofs things against some weirdo
+	// running this on Windows.
```
crozzy, line 42: image

crozzy (approved):
LGTM, you think we need mention somewhere in the docs about the fact we're caching outside the application dir structure? Fine to merge w/o

PR #937, opened 2023-05-15
updater: Extend default updater time to 6 hours

crozzy:
Given the expense of updating and the lack of value to such a high frequency this change extends the default update period from 30 mins to 6 hours.


hdonnay:
Needs an accompanying change to the default `clair` config if it's to have any real effect.

hdonnay (approved):

PR #941, opened 2023-05-16
cicd: gating lints

hdonnay:
This moves some checks to a catch-all "gating lints" job.

It also adds a caching scheme to not re-test commits.

### DIFF
--- a/test/integration/db.go
+++ b/test/integration/db.go
@@ -1,6 +1,7 @@
 package integration
 
 import (
+	"bytes"
 	"context"
 	crand "crypto/rand"
 	"encoding/binary"
@@ -9,6 +10,9 @@ import (
 	"io"
 	"math/rand"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
 	"sync"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


hdonnay:
image
Example of the filename check firing.

hdonnay:
image
Example of the previous-ok check firing.

test/integration/db.go
```diff
@@ -251,7 +262,6 @@ func DBSetup() func() {
 		up.Do(func() {}) // Trip the sync.Once
```
crozzy, line 262: I'm not sure I entirely understand this line, the `sync.Once` seems to be called either way as part of `NeedDB()`

crozzy (commented):
This looks good, just one comment

test/integration/db.go
hdonnay: If the process doesn't need to do any work because there's an external database, tripping the Once means those functions won't get called.

crozzy: Oh ok, I was conflating the two cases: "GH actions external DB" and just "external DB"

crozzy (approved):

crozzy (approved):

PR #942, opened 2023-05-16
java: expose matcher

RTann:
This is so other users of this library may use the matcher, as necessary.


hdonnay (changes requested):
Needs a doc comment if it's going to be exported. LGTM after that

hdonnay (approved):

PR #944, opened 2023-05-22
whiteout: Account for null filepath column

crozzy:
For all existing entries in the package_scanartifact table, filepath will be null. This is not an issue until one of the scanner versions is updated and the indexing process will attempt to reindex the layers with the new scanner and retrieve the existing packages from the package_scanartifact (ie. reconstruct the index_report). This change ensures null is handled.

### DIFF
--- a/datastore/postgres/packagesbylayer.go
+++ b/datastore/postgres/packagesbylayer.go
@@ -120,7 +120,7 @@ WHERE
 		var spkg claircore.Package
 
 		var id, srcID int64
-		var nKind *string
+		var nKind, fPath *string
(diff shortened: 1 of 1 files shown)


crozzy:
Reproduction steps:

- Index a manifest before the whiteout changes were merged
- Update code to latest allowing DB migration to run
- Re-index manifest

Some scanners' versions will have been incremented and clair will re-index with those new scanners, for unchanged scanners it will retrieve the records from `xxxxx_scanartifact`. Those entries will have `package_scanartifact.filepath` null.

hdonnay (approved):

PR #947, opened 2023-05-23
matcher: Use latest_vuln view in lieu of querying vuln table

crozzy:
This view has always existed but never been used (AFAICT). Lately it has been problematic to query all of the vuln table as some vulnerabilities associated with older update_operations we're deleted or duplicated in new update_operations.

TODO

- [x] Database performance analysis
- [x] Tests

### DIFF
--- a/datastore/postgres/querybuilder_test.go
+++ b/datastore/postgres/querybuilder_test.go
@@ -1,17 +1,22 @@
 package postgres
 
 import (
+	"context"
 	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/uuid"
+	"github.com/quay/zlog"
 
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/datastore"
 	"github.com/quay/claircore/libvuln/driver"
 	"github.com/quay/claircore/pkg/pep440"
 	"github.com/quay/claircore/test"
+	"github.com/quay/claircore/test/integration"
+	pgtest "github.com/quay/claircore/test/postgres"
 )
 
 func TestGetQueryBuilderDeterministicArgs(t *testing.T) {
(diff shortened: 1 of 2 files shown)


crozzy:
The number of update_operations / updater should be bounded as long as GC is on. This has the potential to cause increased latency in vulnerability report generation for cases where GC is turned off.

RTann (dismissed):
Just for a sanity check, did you check that the unit test you added fails when we don't use `latest_vulns`?

crozzy:
> Just for a sanity check, did you check that the unit test you added fails when we don't use `latest_vulns`?

yes

datastore/postgres/querybuilder_test.go
```diff
@@ -1,17 +1,23 @@
...
 
 import (
+	"context"
 	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/uuid"
+
```
hdonnay, line 11: extraneous whitespace

datastore/postgres/querybuilder_test.go
```diff
@@ -221,3 +227,87 @@ func TestGetQueryBuilderDeterministicArgs(t *testing.T) {
...
+		res, err := store.Get(ctx, tc.Records, datastore.GetOpts{})
+		if err != nil {
+			t.Fatalf("failed to get vulnerabilities: %v", err)
+		}
+		vulns := []*claircore.Vulnerability{}
+		for _, vs := range res {
+			vulns = append(vulns, vs...)
+		}
+		if len(vulns) != tc.Vulnerable {
+			t.Fatalf("wrong number of vulns, expected %d got %d", tc.Vulnerable, len(vulns))
```
hdonnay, line 310: order should be "got", "want"

hdonnay (changes requested):
Some nits that should be easy to knock out in a rebase.

LGTM after that.

PR #949, opened 2023-05-24
ruby: add vuln matching

RTann:
Note: Ruby recommends semver, but it is not enforced. I am taking the stance here that we will not support non-semver versions. This is relevant because I did find an example which broke semver: https://osv.dev/vulnerability/GHSA-23f7-99jx-m54r (0.119.0.beta1 is not proper semver. It should be 0.119.0-beta1)

### DIFF
--- a/ruby/matcher.go
+++ b/ruby/matcher.go
@@ -1,0 +1,97 @@
+package ruby
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/Masterminds/semver"
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+var _ driver.Matcher = (*Matcher)(nil)
+
+// Matcher attempts to correlate discovered ruby packages with reported
+// vulnerabilities.
+type Matcher struct{}
+
+// Name implements driver.Matcher.
+func (*Matcher) Name() string { return "ruby" }
+
+// Filter implements driver.Matcher.
+func (*Matcher) Filter(record *claircore.IndexRecord) bool {
+	return record.Repository != nil && record.Repository.Name == repository
+}
+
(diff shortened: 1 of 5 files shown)


daynewlee:
It will be nice to have an integration test to run OSV updater and match Ruby CVEs with an index report. But it could be a pain to generate an index report.

daynewlee (commented):
It will be nice to have an integration test to run OSV updater and match Ruby CVEs with an index report. But it could be a pain to generate an index report.

RTann:
> It will be nice to have an integration test to run OSV updater and match Ruby CVEs with an index report. But it could be a pain to generate an index report.

a little bit of a pain, but worth it. Found some issues which I have since fixed :)

ruby/matcher.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/Masterminds/semver"
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+var _ driver.Matcher = (*Matcher)(nil)
+
+// Matcher attempts to correlate discovered python packages with reported
```
crozzy, line 17: nit comment

ruby/packagescanner.go
```diff
@@ -137,7 +137,7 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
 			Name:           name,
 			Version:        version,
 			Kind:           claircore.BINARY,
-			PackageDB:      g,
+			PackageDB:      `ruby:` + g,
```
crozzy, line 140: This change probably won't show up for existing manifests with a scanner version update, is it needed for matching? Do you think it warrants a version update?

crozzy (commented):

crozzy (approved):

PR #951, opened 2023-05-25
java: Add java.Matcher to defaults

crozzy:
As we now have a native java matcher we should add it to the defaults, it will be doing similar work to the CRDA matcher so people will most likely want to disable the CRDA matcher (by removing its config). The CRDA matcher will be removed in later versions.


hdonnay (changes requested):
The list of matchers is sorted. LGTM after that change

PR #953, opened 2023-05-30
indexer: Add ability for package scanner to return a default repository

crozzy:
This has become an issue with more language support where the packages need to be associated to a repository but the repo scanner just ends up being a packagescanner-lite (repeating work). This change lets packageScanners implement an optional method to return a default Repository which will be created during indexing.


hdonnay (approved):
LGTM

PR #954, opened 2023-05-31
gobin: Make matching use VersionFilter for db side filtering

crozzy:
In order to use the db for version filtering, the package created with the gobin package detector needs to add normalized versions. The detector also needs implement VersionFilter and VersionAuthoritative, then Vulnerable() becomes a no-op.

### DIFF
--- a/gobin/exe.go
+++ b/gobin/exe.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	_ "unsafe" // for error linkname tricks
 
+	"github.com/Masterminds/semver"
 	"github.com/quay/zlog"
 
 	"github.com/quay/claircore"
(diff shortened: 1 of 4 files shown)


version.go
```diff
@@ -130,3 +132,13 @@ func (r *Range) Contains(v *Version) bool {
 	// Lower <= v && Upper > v
 	return r.Lower.Compare(v) != 1 && r.Upper.Compare(v) == 1
 }
+
+// FromSemver is the SemVer to claircore.Version mapping used by this package.
+func FromSemver(v *semver.Version) (out Version) {
```
hdonnay, line 137: NAK on adding `semver` to the `claircore` API

hdonnay (changes requested):

version.go
crozzy: Reverted, lmk if you think of a decent common place to add it

gobin/exe.go
```diff
@@ -37,36 +38,77 @@ func toPackages(ctx context.Context, out *[]*claircore.Package, p string, r io.R
 
 	// TODO(hank) This package could use canonical versions, but the
 	// [claircore.Version] type is lossy for pre-release versions (I'm sorry).
+	var runtimeVer claircore.Version
```
crozzy, line 41: Judging by the comment, this needs more thought if we're going to correctly index pre-release versions

hdonnay (approved):
This seems okay if we're fine with misdetections on prerelease and untagged versions.

PR #956, opened 2023-06-01
oracle: support oracle linux 9

RTann:
I also created a ticket to track us potentially adding automatic updates for this in the future: https://issues.redhat.com/browse/PROJQUAY-5594


crozzy (changes requested):
LGTM, You should probably iterate the scanner version too

crozzy (approved):

PR #960, opened 2023-06-05
indexer: Cache scanner table

crozzy:
During the indexing cycle many operations try to read information from the scanner table, this can occasionally lead to resource contention depending on the DB. This reads the table into memory when the indexer store is instanciated and uses when asked for scanner info.

### DIFF
--- a/datastore/postgres/indexer_store.go
+++ b/datastore/postgres/indexer_store.go
@@ -2,7 +2,7 @@ package postgres
 
 import (
 	"context"
-	"fmt"
+	"sync"
 	"time"
 
 	"github.com/jackc/pgx/v5/pgxpool"
@@ -23,8 +23,7 @@ func InitPostgresIndexerStore(ctx context.Context, pool *pgxpool.Pool, doMigrati
 		pool.Reset()
 	}
 
-	store := NewIndexerStore(pool)
-	return store, nil
+	return NewIndexerStore(pool), nil
 }
 
 var _ indexer.Store = (*IndexerStore)(nil)
@@ -33,7 +32,9 @@ var _ indexer.Store = (*IndexerStore)(nil)
 //
 // All the other exported methods live in their own files.
 type IndexerStore struct {
-	pool *pgxpool.Pool
+	pool         *pgxpool.Pool
+	scanners     map[string]int64
+	scannersOnce sync.Once
 }
 
 func NewIndexerStore(pool *pgxpool.Pool) *IndexerStore {
@@ -45,28 +46,6 @@ func NewIndexerStore(pool *pgxpool.Pool) *IndexerStore {
 func (s *IndexerStore) Close(_ context.Context) error {
 	s.pool.Close()
 	return nil
-}
-
-const selectScanner = `
-SELECT
-	id
-FROM
-	scanner
-WHERE
-	name = $1 AND version = $2 AND kind = $3;
-`
-
-func (s *IndexerStore) selectScanners(ctx context.Context, vs indexer.VersionedScanners) ([]int64, error) {
-	ids := make([]int64, len(vs))
-	for i, v := range vs {
-		err := s.pool.QueryRow(ctx, selectScanner, v.Name(), v.Version(), v.Kind()).
-			Scan(&ids[i])
-		if err != nil {
-			return nil, fmt.Errorf("failed to retrieve id for scanner %q: %w", v.Name(), err)
-		}
-	}
-
-	return ids, nil
 }
 
 func promTimer(h *prometheus.HistogramVec, name string, err *error) func() time.Duration {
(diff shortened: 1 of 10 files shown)


datastore/postgres/registerscanners.go
```diff
@@ -82,5 +83,60 @@ SELECT
...
+
+const selectAllScanner = `
+SELECT
+	id, name, version, kind
+FROM
+	scanner;
+`
+
+func (s *IndexerStore) populateScanners(ctx context.Context) error {
+	s.scanners = make(map[string]int64)
```
RTann: This is already done in `indexer_store.go`

datastore/postgres/registerscanners.go
```diff
@@ -82,5 +83,60 @@ SELECT
 		registerScannerDuration.WithLabelValues("insert").Observe(time.Since(start).Seconds())
 	}
 
+	return s.populateScanners(ctx)
+}
+
+const selectAllScanner = `
```
RTann, line 92: nit: `selectAllScanners`?

datastore/postgres/registerscanners.go
```diff
@@ -82,5 +83,60 @@ SELECT
...
+	scanner;
+`
+
+func (s *IndexerStore) populateScanners(ctx context.Context) error {
+	s.scanners = make(map[string]int64)
+	rows, err := s.pool.Query(ctx, selectAllScanner)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve scanners: %w", err)
+	}
+	for rows.Next() {
```
RTann, line 106: I think we still need a `defer rows.Close()` (I don't think it'll close if `rows.Scan` fails) and a `rows.Err()` check, too

datastore/postgres/indexer_store.go
```diff
@@ -39,12 +38,14 @@ var _ indexer.Store = (*IndexerStore)(nil)
 //
 // All the other exported methods live in their own files.
 type IndexerStore struct {
-	pool *pgxpool.Pool
+	pool     *pgxpool.Pool
+	scanners map[string]int64
```
RTann: this may be a decent candidate for `map[unique.Handle[string]]int64` since I think this map will be rather static once it's initially populated

datastore/postgres/registerscanners.go
```diff
@@ -82,5 +83,60 @@ SELECT
 		registerScannerDuration.WithLabelValues("insert").Observe(time.Since(start).Seconds())
 	}
 
+	return s.populateScanners(ctx)
```
RTann: is it assumed that `RegisterScanners` is called only once and in the very beginning of the process?

datastore/postgres/registerscanners.go
```diff
@@ -82,5 +83,60 @@ SELECT
...
+}
+
+const selectAllScanner = `
+SELECT
+	id, name, version, kind
+FROM
+	scanner;
+`
+
+func (s *IndexerStore) populateScanners(ctx context.Context) error {
```
RTann, line 99: nit: `initScanners`

RTann (commented):
So let's say we are in the process of updating the Indexer and/or Matcher version(s) to add or delete a scanner.

Add: I think this is ok. The older versions won't even look for it (I believe)

Delete: The older versions will still claim the scanner exists, though I guess it'll still fail the next part of the scan process, maybe?

Can you add any commentary to potential race scenarios to see if we are ok with the possibilities?

hdonnay:
@crozzy do we (read: you) want to update this and land it or close this PR and try from scratch?

crozzy:
Hopefully dust it off, I'll update today or create a new PR

hdonnay (changes requested):
Needs updating

datastore/postgres/indexer_store.go
```diff
@@ -33,7 +32,8 @@ var _ indexer.Store = (*IndexerStore)(nil)
 //
 // All the other exported methods live in their own files.
 type IndexerStore struct {
-	pool *pgxpool.Pool
+	pool     *pgxpool.Pool
+	scanners sync.Map
```
hdonnay: It seems like this should be a `map[string]int64` and a `sync.Once`.

The use of a concurrent map makes it seem like there's an expectation that `populateScanners` can run concurrently to a `selectScanners` when it really shouldn't.

crozzy: yeah, that's fair enough, good idea

datastore/postgres/registerscanners.go
```diff
@@ -82,5 +83,67 @@ SELECT
...
+	return ids, nil
+}
+
+func (s *IndexerStore) selectScanner(v indexer.VersionedScanner) (int64, error) {
+	id, ok := s.scanners[makeScannerKey(v.Name(), v.Version(), v.Kind())]
+	if !ok {
+		return 0, fmt.Errorf("failed to retrieve id for scanner %q", v.Name())
+	}
+	return id, nil
+}
```
hdonnay: These should try the `Once`, and so should probably also take a `context.Context`.

hdonnay (changes requested):

hdonnay (approved):
LGTM

PR #963, opened 2023-06-07
osv: Split osv updater into updaterset and updater

crozzy:
Having one generic updater has issues when not all ecosystems have
updates. The osv updater will associate the new vulns from the updated
ecosystem with a new osv update_operation and the unchanged vulns will
be associated to an old update_operation. This throws up 3 potential
issues: 1) as we now use the latest_vuln view only the vulns from the
updated ecosytems are available to query 2) The GC can potentially
delete those un-updated vulns 3) when the notification system tries to
calculate the delta between previous and current update_operations it
will receive unexpected results. This pushes the updater creation to
the osv.factory which creates an updater per ecosystem (i.e. osv/go).
The fetching/parsing logic remains largely unchanged.

### DIFF
--- a/updater/osv/osv.go
+++ b/updater/osv/osv.go
@@ -15,7 +15,6 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"path"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


crozzy:
To test this locally:
* run clair with the osv updater
* query the DB for osv related vulns (or run a report that will show osv vulnerabilities)
* go to the DB and modify one of the osv ecosystems' fingerprint/etag (an ecosystem unrelated to the vulns in the previous bullet)
* Rerun the updaters (restart the matcher) and check every ecosystem is skipped expect the one with the modified fingerprint.
* query the DB for osv related vulns (or run a report that will show osv vulnerabilities)

hdonnay (dismissed):
LGTM

updater/osv/osv.go
```diff
@@ -209,8 +178,62 @@ func (u *updater) Fetch(ctx context.Context, fp driver.Fingerprint) (io.ReadClos
...
+	name      string
+	ecosystem string
+	c         *http.Client
+	uri       *url.URL
+}
+
+func (u *updater) Name() string { return u.name }
+
+type UpdaterConfig struct {
+	// The URL serving data dumps behind an S3 API.
```
hdonnay, line 207: I need to circle back and fix these references to the S3 api that aren't true anymore.

crozzy:
Needed change the ruby integration test after it was merged

hdonnay (approved):

PR #983, opened 2023-06-21
indexer: Scope versioned scanners to controller

crozzy:
The indexer.Options are mostly shared objects that are instanciated during startup (i.e. Layerscanner, FetchArena etc) and this is the desired behaviour, however the `Vscnrs` can be modified during indexing (when re-evaluating which scanners need to used for a manifest that is partially up to date) so they should be scoped directly to the `Controller`. This situation is exacerbated when the same manifest is submitted at the same time.


hdonnay (approved):
yeah LGTM

PR #985, opened 2023-06-23
tarfs: handle invalid UTF8 in extended names

hdonnay:
Found an edge case where the tar code doesn't validate that a name is valid UTF8 if it comes from an extended header instead of the standard ustar header. The path normalization function now escapes invalid runes into an escaped hex encoding of the bytes.

Reported-by: @ibazulic


crozzy (approved):
LGTM

PR #988, opened 2023-06-28
migrations: Make manifest_index.manifest_id index concurrent

crozzy:
This table has the potential to be large and normal indexing operations will need to be running concurrrently during this index creation. This will slow the index creation time but ensure the application can function during the creation.


hdonnay:
Not possible with our current migration toolkit.

PR #989, opened 2023-06-28
rpm: module name

hdonnay:
The way we handle the "module" data out of `rpm` regressed.

### DIFF
--- a/rpm/native_db.go
+++ b/rpm/native_db.go
@@ -59,9 +59,22 @@ func packagesFromDB(ctx context.Context, pkgdb string, db nativeDB) ([]*claircor
 			PackageDB: pkgdb,
 		})
 		p := &ps[idx]
-		if strings.Contains(info.Module, ":") {
-			p.Module = info.Module
+		var modStream string
+		if strings.Count(info.Module, ":") > 1 {
+			first := true
+			idx := strings.IndexFunc(info.Module, func(r rune) bool {
+				if r != ':' {
+					return false
+				}
+				if first {
+					first = false
+					return false
+				}
+				return true
+			})
+			modStream = info.Module[:idx]
 		}
+		p.Module = modStream
 		p.Version = constructEVR(&b, &info)
 		p.RepositoryHint = constructHint(&b, &info)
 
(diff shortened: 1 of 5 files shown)


rpm/native_db.go
```diff
@@ -59,9 +59,22 @@ func packagesFromDB(ctx context.Context, pkgdb string, db nativeDB) ([]*claircor
...
+				if r != ':' {
+					return false
+				}
+				if first {
+					first = false
+					return false
+				}
+				return true
+			})
+			modStream = info.Module[:idx]
```
crozzy, line 75: Is there ever a case where `if strings.Count(info.Module, ":") == 1` we just want to keep the string?

hdonnay: That should never happen, and the previous `rpm`-output handling code ignored the field in that case.

crozzy (approved):

rpm/native_db.go
```diff
@@ -59,9 +59,22 @@ func packagesFromDB(ctx context.Context, pkgdb string, db nativeDB) ([]*claircor
 			PackageDB: pkgdb,
 		})
 		p := &ps[idx]
-		if strings.Contains(info.Module, ":") {
-			p.Module = info.Module
+		var modStream string
+		if strings.Count(info.Module, ":") > 1 {
+			first := true
+			idx := strings.IndexFunc(info.Module, func(r rune) bool {
```
RTann, line 65: If there is only one instance of `:`, this fails, right? I honestly don't know what modules look like. Will they either have 0 or >= 2 instances of `:`?

If you're curious, this is what ACS does, which is what ClairCore did in 2021 or so: https://github.com/stackrox/scanner/blob/2.30.0/pkg/rpm/database.go#L271

RTann: ah actually `strings.Count(info.Module, ":") > 1` ensures we only get here if there are >= 2 instances of `:`, so never mind

hdonnay: Yes, this is restoring the [old, desired behavior](https://github.com/quay/claircore/commit/c3b3143dde0827e6a8a302f5627bce4be0c4dc07#diff-13d841be1e4d774855aaad9f350094f9763f988b9d9be7b503019726f5f5b3edL100-L109). Having tag contents that isn't `<name>:<stream>:<version>:<commit>` _shouldn't_ happen.

Issue #968, opened 2023-06-14
Support for Amazon Linux 2023


purplexed:
Hello,

Currently the clair project only supports Amazon Linux AMI 2018.03 and Amazon Linux 2 from the AWS images. Since Clair is used for the basic security scanning on ECR, are there any plans to support Amazon Linux 2023?

Thanks!

crozzy:
Hi!

There are no plans yet but it could be added fairly trivially. Part of the work should be to try and make it the distributions dynamic so that these issues don't come up in the future (see  ubuntu, debian, alpine etc etc)

stewartsmith:
My attempt at that trivial adding (revision 2): https://github.com/quay/claircore/pull/990

hdonnay:
Seems done

PR #990, opened 2023-07-01
Improve Amazon Linux detection, adding AL2023

stewartsmith:
A few fixes in here:
- Add the new Amazon Linux 2023
- Add tests for older Amazon Linux 1 versions, fixing up the regex for detecting the OS
- Use consistent names for variables matching public names.

Plus bump the version up to v2023.0.0 to match that it knows about AL2023.

There's no doubt a bit of a better way to structure this for being more future-proof, but this appears to pass the tests?


crozzy:
Thanks for this @stewartsmith, I'm taking a look

aws/releases.go
```diff
@@ -49,12 +53,24 @@ var linux2Dist = &claircore.Distribution{
...
-	case Linux1:
-		return linux1Dist
-	case Linux2:
-		return linux2Dist
+	case AmazonLinux1:
+		return AL1Dist
+	case AmazonLinux2:
+		return AL2Dist
+	case AmazonLinux2023:
+		return AL2Dist
```
crozzy, line 73: This is incorrect, I think there is an intrinsic problem in the tests as `releaseToDist()` is called on the `got` and `want` values (via `scanner.parse`), meaning they are both equally wrong.

aws/distributionscanner.go
```diff
@@ -12,13 +12,15 @@ import (
...
+// Amazon Linux keeps a consistent os-release file between
 // major releases.
 // All tested images on docker hub contained os-release file
+// scannerVersion matches the most recent major version of
+// Amazon Linux that we have tested with.
 
 const (
 	scannerName    = "aws"
-	scannerVersion = "v0.0.1"
+	scannerVersion = "v2023.0.0"
```
crozzy, line 23: We've gone down the route of very simple scanner versions (i.e. "2"), it's used to inform the index_state and semver (or the like) is really overkill

crozzy (changes requested):
Thanks for the PR, we've held off on adding any distributions without dynamic distribution discovery (see; ubuntu, debian, alpine etc.). I'm not sure how much you know about the Amazon Linux ecosystem but it'd be great to have a plan for that dynamic discovery. The problem with adding them piecemeal is that our release cycle needs to coincide with distro releases, which considering the number of supported distros we deal with ends up being a losing battle (as you can see with the issue that is linked to this PR).

stewartsmith:
> Thanks for the PR, we've held off on adding any distributions without dynamic distribution discovery (see; ubuntu, debian, alpine etc.). I'm not sure how much you know about the Amazon Linux ecosystem but it'd be great to have a plan for that dynamic discovery. The problem with adding them piecemeal is that our release cycle needs to coincide with distro releases, which considering the number of supported distros we deal with ends up being a losing battle (as you can see with the issue that is linked to this PR). So essentially, I don't mind merging these changes but we need an issue and a plan for dynamic discovery.

I'm deeply involved in the Amazon Linux ecosystem, it's my day job :)

I understand the ask for dynamic discovery for new OS versions. I guess we'll need something that includes the other important details such as CPE ID, repository locations, where to find CVE information etc etc. If there's something somewhat generic that you're aware of, I'd love to know about it.

I've created https://github.com/amazonlinux/amazon-linux-2023/issues/389 for us to track it.

aws/distributionscanner.go
stewartsmith: I'll replace with "2" then.

aws/releases.go
stewartsmith: Looks like I'll have to spend a moment understanding what's actually going on there rather than mostly just going through and adding/replacing the AL1/AL2 bits :)

crozzy: I just mean, the third branch here appears to erroneously returning `AL2Dist`, and given that it's disconcerting that we don't have a test to catch that.

stewartsmith: Oh, I guess I shouldn't look at reviews late at night. Thanks for the catch there, I'll see what I can figure out.

crozzy:
@stewartsmith are you still working on this? We have this scheduled for v4.7.1

stewartsmith:
> @stewartsmith another question, I'm guessing from this update that AL2022 is obsolete/out of support? I can't find any information about the support scope.

AL2022 was the tech preview for what became AL2023. If we didn't slip release date the name would have stuck, but we did, so we changed it to be current to when we were going GA.

stewartsmith:
> @stewartsmith are you still working on this? We have this scheduled for v4.7.1

Let me try and find my dev environment again and submit a V2.... this was very much an opportunistic PR for me without the thought of when I'd have time to follow up :) 

If my timing isn't great and someone else picks it up and makes fixes, I'm not at all going to be even remotely offended. I'm surprised I could be coherent enough in Go to not get an immediate NAK :)

stewartsmith:
Good news! I found the time to address the comments. I added a test case for https://github.com/quay/claircore/pull/990#discussion_r1253711497 as well

stewartsmith:
Will also fix https://github.com/quay/claircore/issues/968

crozzy (approved):

Issue #992, opened 2023-07-03
osv: duplicated ecosystems


hdonnay:
I noticed that the OSV updater reports duplicated ecosystems:
```
ecosystems=["almalinux","almalinux","almalinux","gsd","go","hex","maven","nuget","packagist","pub","pypi","rocky linux","rocky linux","rocky linux","rubygems","uvi","crates.io","npm"]
```

This has the smell of a loop variable bug.

crozzy:
It appear like the different versioned ecosytems will produce the duplicates i.e. (AlmaLinux, AlmaLinux:8, AlmaLinux:9) When `driver.UpdaterSet.Add()` is called it will return an error if an updater with the same name has already been registered. Looks like we're disregarding that error.

PR #993, opened 2023-07-03
osv: Dedupe ecosystems

crozzy:
When parsing ecosytems.txt and removing versions we record duplicates. Currently, the code will attempt to add these as duplicate updaters and return an error, however the error is not checked. Therefore this change won't actually change the behaviour of the osv updater but rather improve the logging.


crozzy:
closes https://github.com/quay/claircore/issues/992

hdonnay (approved):
LGTM

Issue #962, opened 2023-06-06
docs: mention test cache directory


hdonnay:
>[...] you think we need mention somewhere in the docs about the fact we're caching outside the application dir structure?

_Originally posted by @crozzy in https://github.com/quay/claircore/pull/934#pullrequestreview-1466228945_

PR #997, opened 2023-07-05
docs: add section for tests

hdonnay:
Closes: #962


crozzy (approved):
LGTM, just one nit, I think it's ok to merge apart from that

PR #1003, opened 2023-07-10
datastore: Revert lastest_vuln view usage

crozzy:
Currently the latest_vuln view is too resource intensive and needs to be reverted.


RTann:
Does this not need to be reverted, then?

crozzy:
No, we're going to move forward instead

crozzy:
https://github.com/quay/claircore/pull/1004 something like this

PR #1004, opened 2023-07-10
datastore: Pre-lookup valid update_operations before querying

crozzy:
By looking up the relevant update_operations (i.e. the latest ones) at the beginning of the vulnerability lookups we can avoid using the more expensive latest_vuln view.

### DIFF
--- a/datastore/postgres/get.go
+++ b/datastore/postgres/get.go
@@ -9,10 +9,10 @@ import (
 	"github.com/jackc/pgx/v4"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"github.com/quay/claircore/datastore"
 	"github.com/quay/zlog"
 
 	"github.com/quay/claircore"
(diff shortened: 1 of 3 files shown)


datastore/postgres/get.go
```diff
@@ -9,9 +9,10 @@ import (
 	"github.com/jackc/pgx/v4"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"github.com/quay/claircore/datastore"
 	"github.com/quay/zlog"
 
+	"github.com/quay/claircore/datastore"
+
```
hdonnay, line 15:
```suggestion
```

hdonnay (changes requested):
I think the imports need fixing, otherwise this is the same solution I hit on.

We really need to get on the reworking pass through here.

datastore/postgres/get.go
crozzy: woops, thanks

hdonnay (approved):

PR #1006, opened 2023-07-13
matcher: Use materialized view to speed up queries to update_operations

crozzy:
Querying the update_operations table for the latest operations creates a lot of CPU load on the DB. When the data is pulled into a materialized view we can take the CPU hit when the update_operation is being created as opposed to every time we receive a vulnerability report request.

### DIFF
--- a/datastore/postgres/enrichment.go
+++ b/datastore/postgres/enrichment.go
@@ -106,6 +106,7 @@ VALUES
 ON CONFLICT
 DO
 	NOTHING;`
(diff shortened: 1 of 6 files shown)


hdonnay (approved):
LGTM

I could go either way on whether the view update should be in the transaction or not. I guess I'm not sure if that even works or not.

crozzy:
> LGTM
> 
> I could go either way on whether the view update should be in the transaction or not. I guess I'm not sure if that even works or not.

I *think* it makes sense to be out due to the transaction needing to be committed before the refresh happens

PR #1012, opened 2023-07-24
postgres: rework query with join

hdonnay:
Looking at the query planner, this `JOIN` should actually speed things up.


crozzy (approved):
LGTM

hdonnay:
Test failed because of a GitHub 429 (???), will re-run.

PR #1014, opened 2023-07-25
gobin: reduce some log chatter

hdonnay:
Spotted when running across the quayio corpus. These changes should pollute the warning logs less.


crozzy (approved):
LGTM, I'd say we're probably good not iterating the version right?

hdonnay:
Yes -- After this change the created packages will have _a_ normalized version populated for the `runtime` package, but it's not used due to the lack of prerelease support for that workflows.

hdonnay:
Decided that urlencoding the whole path was going to be annoying, so I fixed the escaping in `zlog`.

crozzy (approved):
:+1:

PR #1015, opened 2023-07-26
Python: Account for fixed_in information that has no fixed version

crozzy:
Currently we are assuming all python OSV url-encoded version strings have a fixed in version, but it appears like if a fixed version is not included the package is still vulnerable. We need to account for this behaviour.

Example warning:

```
WRN missing upper version component=internal/matcher/Controller.Match fixed in version=introduced=2.2.0&lastAffected=3.0.1 matcher=python package=pypdf2 request_id=999390f4cd1c7a56
```

### DIFF
--- a/python/matcher.go
+++ b/python/matcher.go
@@ -37,15 +37,6 @@ func (*Matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, v
 		return true, nil
 	}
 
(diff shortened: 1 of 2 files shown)


hdonnay:
Isn't "lastAffected" effectively "fixed"? What's the difference?

hdonnay (approved):
LGTM

PR #1016, opened 2023-07-26
jar: don't info-log every properties file

hdonnay:
This gets real chatty in a loaded production deployment.


crozzy (approved):

hdonnay:
Clean rebase, will merge with green CI

PR #1020, opened 2023-07-27
datastore: refresh materialized view concurrently

crozzy:
Without concurrency we are seeing heavy locking on the production DB. This option will mean data is slightly latent but will not lock the materialized view for reading.

### DIFF
--- a/datastore/postgres/enrichment.go
+++ b/datastore/postgres/enrichment.go
@@ -106,7 +106,7 @@ VALUES
 ON CONFLICT
 DO
 	NOTHING;`
-		refreshView = `REFRESH MATERIALIZED VIEW latest_update_operations;`
+		refreshView = `REFRESH MATERIALIZED VIEW CONCURRENTLY latest_update_operations;`
 	)
 	ctx = zlog.ContextWithValues(ctx, "component", "datastore/postgres/UpdateEnrichments")
 
--- a/datastore/postgres/updatevulnerabilities.go
+++ b/datastore/postgres/updatevulnerabilities.go
@@ -80,7 +80,7 @@ func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string
 			(SELECT id FROM vuln WHERE hash_kind = $1 AND hash = $2))
 		ON CONFLICT DO NOTHING;`
 		undo        = `DELETE FROM update_operation WHERE id = $1;`
-		refreshView = `REFRESH MATERIALIZED VIEW latest_update_operations;`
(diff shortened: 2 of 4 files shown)


RTann:
what are the implications of reading potentially older data? we will just read the older data? Will it have any implications on the garbage collection process (we read data which no longer exists in the main table)?

datastore/postgres/enrichment.go
```diff
@@ -106,7 +106,7 @@ VALUES
 ON CONFLICT
 DO
 	NOTHING;`
-		refreshView = `REFRESH MATERIALIZED VIEW latest_update_operations;`
+		refreshView = `REFRESH MATERIALIZED VIEW CONCURRENTLY latest_update_operations;`
```
RTann, line 109: how often does this happen? Is this very resource intensive to do as often as it's done now?

datastore/postgres/updatevulnerabilities.go
```diff
@@ -80,7 +80,7 @@ func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string
 			(SELECT id FROM vuln WHERE hash_kind = $1 AND hash = $2))
 		ON CONFLICT DO NOTHING;`
 		undo        = `DELETE FROM update_operation WHERE id = $1;`
-		refreshView = `REFRESH MATERIALIZED VIEW latest_update_operations;`
+		refreshView = `REFRESH MATERIALIZED VIEW CONCURRENTLY latest_update_operations;`
```
RTann, line 83: similar questions

hdonnay (approved):

crozzy:
> what are the implications of reading potentially older data? we will just read the older data? Will it have any implications on the garbage collection process (we read data which no longer exists in the main table)?

By design Clair will retain 2 update operations meaning that even if the GC ran before the materialized view was updated (which shouldn't be possible) we should still be querying a valid update_operation. In reality the materialized view should be updated before the GC as the call to refresh still blocks code-wise, but it shouldn't lock DB wise. I think there is a potential concern here and that is if there an error after the update_operation during the process of persisting the vulnerabilities. This has lead to hypothesizing about the update_operation creation being included in the transaction which I think is probably a decent idea (in a different PR).

datastore/postgres/enrichment.go
crozzy: This should called ~ 400-500 times per update cycle (once per updater(rhel has a lot of updaters)). I don't think the frequency of this query is the issue, more the fact that it happens concurrently while we're sending 12000 matching queries / sec that reference this view.

hdonnay (approved):
still LGTM

RTann (approved):

PR #1025, opened 2023-07-31
gobin: fix unit test when run on ARM64

frostmar:
Unit test `TestScanner` builds and scans a layer containing a go executable. Clair doesn't recognise eg. Mac ARM64 go executables, so this test fails when run on an ARM Mac.

Fix: always (cross-)compile a Linux AMD64 executable, regardless of the host platform.

Alternatively it's simple to get gobin's `Detector.Scan(...)` to recognise the Mac ARM64 Mach-O executable header, but I don't whether other changes would be needed to fully support Mac ARM go binaries in Clair.


hdonnay:
This looks like the right fix. I don't think adding support for Mach-O makes a lot of sense, as OSX containers aren't a thing as far as I know.

Issue #1018, opened 2023-07-27
Introduce support for debain & alpine based distroless images


Kieran-Muller:
## The Problem

[Distroless images](https://github.com/GoogleContainerTools/distroless) somewhat resemble Debian images with their `/etc/os-release` files containing a number of references to detain and the fact that distroless is a slimmed down Debian image.

There are two issues faced with the distroless images currently:
1. The malformed `/etc/os-release` file is causing a bad version string in the dist table so Clair can't match any vulnerabilities correctly
2. Clair is unable to find packages due the non-standard way in which distroless stores them.

### Problem 1 
The malformed `/etc/os-release` file is causing a bad version string in the dist table so Clair can't match any vulnerabilities correctly.

Sample os-release file contents from a distroless based image.
```
$ cat /tmp/static/etc/os-release
PRETTY_NAME="Distroless"
NAME="Debian GNU/Linux"
ID="debian"
VERSION_ID="11"
VERSION="Debian GNU/Linux 11 (bullseye)"
HOME_URL="https://github.com/GoogleContainerTools/distroless"
SUPPORT_URL="https://github.com/GoogleContainerTools/distroless/blob/master/README.md"
BUG_REPORT_URL="https://github.com/GoogleContainerTools/distroless/issues/new"
```

Which leads to an index report as so: 
```json
{
  "manifest_hash": "sha256:f858d0eb854af55df05856e0ef2da291327894d0f01cd3efe2e2cfc6d13e088d",
  "state": "IndexFinished",
  "packages": {},
  "distributions": {
    "5873": {
      "id": "5873",
      "did": "debian",
      "name": "Debian GNU/Linux",
      "version": "11 (Debian GNU/Linux 11 (bullseye)",
      "version_code_name": "Debian GNU/Linux 11 (bullseye",
      "version_id": "11",
      "arch": "",
      "cpe": "",
      "pretty_name": "Debian GNU/Linux 11 (Debian GNU/Linux 11 (bullseye)"
    }
  },
  "repository": {},
  "environments": {},
  "success": true,
  "err": ""
} 
```

As opposed to something more correct:

```json
{
  "manifest_hash": "sha256:182f6734d63194209cf188018f6288480d11d0a7f234f50b60a4db81055cf3c5",
  "state": "IndexFinished",
  "packages": {},
  "distributions": {
    "653": {
      "id": "653",
      "did": "debian",
      "name": "Debian GNU/Linux",
      "version": "11 (bullseye)",
[...]

PR #1026, opened 2023-08-01
debian:distroless dist parsing support and tests

Kieran-Muller:
Linked issue: https://github.com/quay/claircore/issues/1018

This PR aims to introduce support for distroless /etc/os-release file parsing in order to allow for vulnerability matching. Further details can be found in the linked issue.

This PR contains:

- Test files of example /etc/os-release files for distroless images
- An update to the distributionscanner_test.go which adds "test cases" for the dist and distroless_dist file paths containing test data.
- Use of a regex pattern to get the version name from the VERSION field wrapped in brackets, which is then unpicked as before.

### DIFF
--- a/debian/distributionscanner.go
+++ b/debian/distributionscanner.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"regexp"
 	"runtime/trace"
 	"strconv"
 	"strings"
@@ -31,7 +32,7 @@ type DistributionScanner struct{}
 func (*DistributionScanner) Name() string { return "debian" }
 
 // Version implements [indexer.VersionedScanner].
-func (*DistributionScanner) Version() string { return "2" }
+func (*DistributionScanner) Version() string { return "3" }
 
 // Kind implements [indexer.VersionedScanner].
 func (*DistributionScanner) Kind() string { return "distribution" }
@@ -85,10 +86,13 @@ func findDist(ctx context.Context, sys fs.FS) (*claircore.Distribution, error) {
 		return nil, nil
 	}
 
+	// Regex pattern matches item within string that appear as so: (bookworm), (buster), (bullseye)
+	ver := regexp.MustCompile(`\(\w+\)$`)
+
 	name, nameok := kv[`VERSION_CODENAME`]
 	idstr := kv[`VERSION_ID`]
 	if !nameok {
-		name = strings.TrimFunc(kv[`VERSION`], func(r rune) bool { return !unicode.IsLetter(r) })
+		name = strings.TrimFunc(ver.FindString(kv[`VERSION`]), func(r rune) bool { return !unicode.IsLetter(r) })
 	}
 	if name == "" || idstr == "" {
 		zlog.Info(ctx).
--- a/debian/distributionscanner_test.go
+++ b/debian/distributionscanner_test.go
@@ -2,6 +2,7 @@ package debian
 
 import (
 	"context"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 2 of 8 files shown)


cathyyoung:
@hdonnay Kieran's out on vacation for a few days now, but are you able to review this please?

Kieran-Muller:
Alternatively, @crozzy if you're available for review :)

crozzy:
This is a positive improvement, I'm still slightly confused why I get a well-formed distro when running main manually: 
`clairctl -D -c local-dev/clair/config.yaml report -o json gcr.io/distroless/static-debian12 | jq`

```
  "distributions": {
    "1": {
      "id": "1",
      "did": "debian",
      "name": "Debian GNU/Linux",
      "version": "12 (bookworm)",
      "version_code_name": "bookworm",
      "version_id": "12",
      "arch": "",
      "cpe": "",
      "pretty_name": "Debian GNU/Linux 12 (bookworm)"
    }
```

But that `name` is a part of the `dist.Version` and the `dist.Version` is use as a matcher constraint so it makes sense that this would fix the matching

debian/distributionscanner_test.go
```diff
@@ -13,28 +14,40 @@ import (
 func TestDistributionScanner(t *testing.T) {
 	ver := regexp.MustCompile(`^\d+ \(\w+\)$`)
 	ctx := zlog.Test(context.Background(), t)
-	ents, err := os.ReadDir(`testdata/dist`)
+	ents, err := os.ReadDir(`testdata/distroless_dist`)
```
crozzy, line 17: We want to keep the `testdata/dist` here right? So we're still testing the normal debian os-release files

debian/distributionscanner.go
crozzy, line 1: We'll want to iterate the scanner version here to ensure things are re-indexed

crozzy (changes requested):
This looks like the right approach, I need to do a bit more manual testing, a couple of code comments

crozzy:
> This is a positive improvement, I'm still slightly confused why I get a well-formed distro when running main manually: `clairctl -D -c local-dev/clair/config.yaml report -o json gcr.io/distroless/static-debian12 | jq`
> 
> ```
>   "distributions": {
>     "1": {
>       "id": "1",
>       "did": "debian",
>       "name": "Debian GNU/Linux",
>       "version": "12 (bookworm)",
>       "version_code_name": "bookworm",
>       "version_id": "12",
>       "arch": "",
>       "cpe": "",
>       "pretty_name": "Debian GNU/Linux 12 (bookworm)"
>     }
> ```
> 
> But that `name` is a part of the `dist.Version` and the `dist.Version` is use as a matcher constraint so it makes sense that this would fix the matching

OK I know what the issue is here:
1 layer contains:
```
PRETTY_NAME="Distroless"
NAME="Debian GNU/Linux"
ID="debian"
VERSION_ID="12"
VERSION="Debian GNU/Linux 12 (bookworm)"
HOME_URL="https://github.com/GoogleContainerTools/distroless"
SUPPORT_URL="https://github.com/GoogleContainerTools/distroless/blob/master/README.md"
BUG_REPORT_URL="https://github.com/GoogleContainerTools/distroless/issues/new"
```
and another layer contains:
```
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
So depending on how the layers are processed, one or the other will be parsed, Clair operates under the assumption that there is only one distro per container image.

Could you also add a testcase [here](https://github.com/quay/claircore/blob/main/osrelease/scanner_test.go)? I think after that and after the other comments are addressed we'll be good to go. Thanks

debian/distributionscanner_test.go
Kieran-Muller: My mistake, you're right! I've corrected it.

Kieran-Muller:
> Could you also add a testcase [here](https://github.com/quay/claircore/blob/main/osrelease/scanner_test.go)? I think after that and after the other comments are addressed we'll be good to go. Thanks

I've added two samples to `TestParse()`. I had attempted to get something in to the `TestLayer()` test but I could not figure out how that was working and why it wasn't failing for my case when I wanted it to.

crozzy:
Great stuff, last thing, can you squash the commits into one? Then I think we'll be good to merge

Kieran-Muller:
Thanks to the help of @paulaldridge we have successfully got it squashed into 1 commit! 🥳

crozzy (approved):

crozzy:
Thanks for the patch @Kieran-Muller

PR #1030, opened 2023-08-02
postgres: remove hardcoded MaxConns config parameter

crozzy:
This had already been removed but when the code was restructured it was accidentally added back. This hopefully removes it once and for all.


hdonnay (approved):
:raised_hands:

PR #1037, opened 2023-08-08
python: account for just introduced key in OSV data

crozzy:
We weren't (I wasn't) accounting for the valid scenario when the vulnerability was open-ended defined by the "introduced" key and the absence of any other valid keys.


hdonnay (approved):
LGTM

PR #1042, opened 2023-08-15
datastore: remove arbitrary register scanner 1sec timeouts

frostmar:
1 second was too short to initialise with a remote Postgres, when running under the debugger in VSCode

### DIFF
--- a/datastore/postgres/registerscanners.go
+++ b/datastore/postgres/registerscanners.go
@@ -63,7 +63,7 @@ SELECT
 	var tctx context.Context
 	var done context.CancelFunc
 	for _, v := range vs {
-		tctx, done = context.WithTimeout(ctx, time.Second)
+		tctx, done = context.WithTimeout(ctx, 5*time.Second)
 		start := time.Now()
 		err = s.pool.QueryRow(tctx, exists, v.Name(), v.Version(), v.Kind()).
 			Scan(&ok)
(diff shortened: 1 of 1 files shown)


crozzy:
The issue here is that during concurrent indexing requests this table (`scanner`) appears to be locking. Given how small the table is, there is no way it should be taking more that 1 second to get results back (other than if query congestion is slowing things up). I have been iterating through ideas to speed the scanner table access up [here](https://github.com/quay/claircore/pull/960/files), the idea currently being that we read the scanner table after initial insertion and only access the map during the subsequent indexing processes.

More generally, our approach has been to remove the arbitrary timeouts from the datastore layer and let the passed-in context dictate the timeouts through the request lifetime. This means Clair is more adaptable to different customer environments where the DB resources are smaller/slower. 

All that to say, I think modifying this PR to remove the 1 second timeout would be merged and would address your problem then we can address the underlying issue of `scanner` table access going forward.

frostmar:
> Given how small the table is, there is no way it should be taking more that 1 second to get results back

I suspect with a remote database, a significant proportion of the 1sec might be spent setting up and initializing the first TLS connection to the DB? This was showing up for me only under the golang debugger in VSCode, so possibly that slows things down somewhat (although I can't offhand think of a plausible reason why)

I'll take a look to see if I can find a way to get rid of a constant timeout

frostmar:
Branch updated to remove the arbitrary 1sec timeouts.
It now just uses the caller-supplied context for all steps.

crozzy (approved):
LGTM, thanks!

PR #1044, opened 2023-08-16
Indexer: handle SetIndexReport() errors

frostmar:
In the indexer controller's state machine, inside `run()`, errors from `SetIndexReport()` were logged but otherwise unhandled. 
If an error occurred, a correct index report could be returned in the HTTP response, but an incomplete one from a previous state would exist in the database.

### DIFF
--- a/indexer/controller/controller.go
+++ b/indexer/controller/controller.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"errors"
+	"fmt"
 	"math/rand"
 	"time"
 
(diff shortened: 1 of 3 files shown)


crozzy:
Thanks for raising this, I think it's valid and the comment updates are appreciated.

crozzy (approved):
LGTM, thanks for the PR

crozzy:
fyi We create nightlies here: https://quay.io/repository/projectquay/clair?tab=tags the are build directly from clair/claircore's main, there is no guarantee of stability but they can be useful to test any changes out in a non-prod environment without having to wait for a release/build images yourself.

PR #1045, opened 2023-08-16
ci: create clair PR when new claircore tag is created

crozzy:
Something we have to do over and over is update clair with the latest claircore tag. This automates some of that workflow.

TODO:

- [x] How to authenticate? (with an app)

### DIFF
--- a/.github/workflows/update-clair.yml
+++ b/.github/workflows/update-clair.yml
@@ -1,0 +1,59 @@
+on:
+  push:
+    tags:
+      - v*.*.*
+
+name: pr-for-clair
+
+jobs:
+  send-pull-requests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup
+        run: |
+          tag=`basename ${{ github.ref }}`
+          branch="chore-update-claircore-to-${tag}"
+          echo "VERSION=${tag}" >> $GITHUB_ENV
+          echo "BRANCH=${branch}" >> $GITHUB_ENV
+
+      - name: Generate token
+        id: generate_token
+        uses: tibdex/github-app-token@v1
+        with:
+            app_id: ${{ secrets.CLAIR_APP_ID }}
+            private_key: ${{ secrets.CLAIR_APP_PRIVATE_KEY }}
+
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: main
+          repository: quay/clair
+
+      - name: Update claircore
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-clair.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          add-paths: go.mod,go.sum
+          token: ${{ steps.generate_token.outputs.token }}
+          base: main
+          commit-message: |
+            'chore: update claircore to ${{ env.VERSION }}
+            
+            Update claircore to latest tag
+            '
+          committer: No One <noreply@quay.io>
```
hdonnay, line 50: Can this be something like `Claircore-CI <${{ github.actor }}@users.noreply.github.com>`?

.github/workflows/update-clair.yml
```diff
@@ -0,0 +1,59 @@
...
+        id: cpr
+        uses: peter-evans/create-pull-request@v5
+        with:
+          add-paths: go.mod,go.sum
+          token: ${{ steps.generate_token.outputs.token }}
+          base: main
+          commit-message: |
+            'chore: update claircore to ${{ env.VERSION }}
+            
+            Update claircore to latest tag
```
hdonnay, line 48:
```suggestion
            Update claircore to latest release.
```

.github/workflows/update-clair.yml
```diff
@@ -0,0 +1,59 @@
+on:
+  push:
+    tags:
+      - v*.*.*
```
hdonnay, line 4:
```suggestion
      - v1.*.*
```

We're going to rev this at some point.

hdonnay (changes requested):
Some minor nits.

I think this is fine as-is. Do you have any strong feelings on patch-only updates vs minor and patch?

.github/workflows/update-clair.yml
crozzy: Not sure what the point in limiting to v1 tags is? Would we not want this behavior when it's iterated to v2 v3 etc (assuming we don't have a better mechanism by then)?

crozzy: I suppose I just thought it'd create PRs for everything and we'd just close what we didn't want. What would be the advantage of paring down?

hdonnay: Yeah, fair enough.

hdonnay (dismissed):

hdonnay (approved):

PR #1046, opened 2023-08-16
cicd: specify branch in codecov arguments

hdonnay:
I _think_ this is the reason that commits are not getting associated with the `main` branch when merged.


crozzy (approved):
I'm going to say FeelsGTM

PR #1048, opened 2023-08-21
claircore: fix toolkit versioning

hdonnay:
The `toolkit` module needs to be versioned properly so that consumers are using the correct version.


crozzy (approved):
LGTM

PR #1053, opened 2023-09-01
java: discount Red Hat patched packages from matching

crozzy:
Once RH has patched a package, the version it reports is no longer reliable to use for vulnerability matching. If the package is truely vulnerable it should be picked up on a container level by the rhcc matcher.


hdonnay:
Strongly against doing this or something akin to it in every Matcher implementation.

PR #1055, opened 2023-09-05
mdbook improvements

hdonnay:
These changes move the mdbook common code into a dedicated package, then moves the commands to dedicated packages.


crozzy (approved):
LGTM

PR #1056, opened 2023-09-05
rhcc: add a resolver to discount non-rhcc packages

crozzy:
This change adds an RHCC resolver that will check which packages came from the same layer as a Red Hat container layer and remove them from the final index report. This is to address the false positives that can be created when RH patches packages but they still report their vulnerable versions.

I'm sure there's a nicer way to do this but wanted to put something up so we have a reference of where a post-process step like this could exist.

### DIFF
--- a/rhel/rhcc/resolver.go
+++ b/rhel/rhcc/resolver.go
@@ -1,0 +1,59 @@
+package rhcc
+
+import (
+	"context"
+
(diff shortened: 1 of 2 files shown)


crozzy:
related to: https://github.com/quay/claircore/pull/1053

PR #1057, opened 2023-09-08
enricher: add RHCC enricher

crozzy:
This change introduces an enricher who's purpose is to scope down the vulnerability report for layers that have RHCC packages. This approach helps to keep the index report unchanged and therefore state is less of an issue, it also builds on existing machinary.


crozzy:
The result is something like this:
```json
{
  "message/vnd.clair.map.layer; enricher=clair.rhcc": [
    {
      "2": "sha256:013d2d7b1d1f8a94ffb762e6960fb5f483c60e38bd8740b1ec1f1a557d3d1bbf",
      "4": "sha256:013d2d7b1d1f8a94ffb762e6960fb5f483c60e38bd8740b1ec1f1a557d3d1bbf",
      "6": "sha256:d93e847446533e9af99b246c8cebea2527bf3e9db82e45fb59e3d9f1443a7d2c",
      "8": "sha256:d93e847446533e9af99b246c8cebea2527bf3e9db82e45fb59e3d9f1443a7d2c"
    }
  ]
}
```

enricher/rhcc/rhcc.go
```diff
@@ -0,0 +1,45 @@
...
+
+type Enricher struct{}
+
+var (
+	_ driver.Enricher = (*Enricher)(nil)
+)
+
+const (
+	// Type is the type of data returned from the Enricher's Enrich method.
+	Type = `message/vnd.clair.map.layer; enricher=clair.rhcc`
```
crozzy, line 20: I have omitted any schema here as the format is just a basic map

enricher/rhcc/rhcc.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+func (e *Enricher) Name() string { return "rhcc" }
+
+func (e *Enricher) Enrich(ctx context.Context, g driver.EnrichmentGetter, r *claircore.VulnerabilityReport) (string, []json.RawMessage, error) {
+	problematicPkgs := make(map[string]string)
```
hdonnay, line 26: Shouldn't this just be a `map[LayerID]bool`? It just needs to be a list of layers, right? I don't think `map[PkgID]LayerID` buys us a lot.

hdonnay (changes requested):
Question about the approach

enricher/rhcc/rhcc.go
crozzy: My thought was that it might be useful in the future for people to be able to cross-reference _which_ rhcc layer it is too and the added data seems negligible size-wise and nil complexity-wise.

hdonnay: I guess the `rhcc` packages should be at most one per layer. OK.

enricher/rhcc/rhcc.go
hdonnay, line 1: I missed this previously -- why is this in a separate package from the existing `rhcc` matcher?

crozzy: To be consistent with the cvss enricher, i.e someone creating a `Libvuln` object would do something like:
```go
import (
	"github.com/quay/claircore/enricher/cvss"
	"github.com/quay/claircore/enricher/rhcc"
	"github.com/quay/claircore/libvuln"
)

 s, err := libvuln.New(ctx, &libvuln.Options{
               ...
		Enrichers: []driver.Enricher{
			&cvss.Enricher{},
			&rhcc.Enricher{},
		},
	})
```
but I'm fine moving it (would also negate the need to make the `GoldRepo` public).

enricher/rhcc/rhcc.go
```diff
@@ -0,0 +1,48 @@
...
+	for id, _ := range r.Packages {
+		if envs, ok := r.Environments[id]; ok {
```
crozzy, line 28: Just do binary packages

rhel/rhcc/rhcc.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/quay/claircore/toolkit/types/cpe"
 )
 
-var goldRepo = claircore.Repository{
+var GoldRepo = claircore.Repository{
```
RTann, line 17: nit: Maybe add a godoc for this since it's now available outside the package

RTann (commented):
Haven't looked at the logic much yet, but I have a question. It was brought to my attention layers may be squashed, even if from the base image. For example `podman build --squash-all`. In that scenario, we will just see a single layer I believe, and I'm not sure how useful this would be in that scenario. I'm wondering what your thoughts are about this? I'm not saying to throw this out, but rather we may just need to make it clear to users that this may not be *that* useful for those images which are completely squashed

rhel/rhcc/rhcc.go
crozzy: Will do this in https://github.com/quay/claircore/pull/1416 as it will (hopefully) be merged first.

crozzy:
> Haven't looked at the logic much yet, but I have a question. It was brought to my attention layers may be squashed, even if from the base image. For example `podman build --squash-all`. In that scenario, we will just see a single layer I believe, and I'm not sure how useful this would be in that scenario. I'm wondering what your thoughts are about this? I'm not saying to throw this out, but rather we may just need to make it clear to users that this may not be _that_ useful for those images which are completely squashed

If the image is squashed we should still be able to identify every RH layer that has composed the squashed layer but yeah, any client making decisions off of this information would need to be aware that it's fallible if the input has been manipulated. Clients could potentially be told to ignore vulnerabilities in RH layers that are squashed together with application layers.

enricher/rhcc/rhcc.go
```diff
@@ -0,0 +1,48 @@
...
+
+const (
+	// Type is the type of data returned from the Enricher's Enrich method.
+	Type = `message/vnd.clair.map.layer; enricher=clair.rhcc`
+)
+
+func (e *Enricher) Name() string { return "rhcc" }
+
+func (e *Enricher) Enrich(ctx context.Context, g driver.EnrichmentGetter, r *claircore.VulnerabilityReport) (string, []json.RawMessage, error) {
+	problematicPkgs := make(map[string]string)
```
RTann, line 26: what makes them problematic? 😄

enricher/rhcc/rhcc.go
```diff
@@ -0,0 +1,48 @@
...
+	for id, p := range r.Packages {
+		if envs, ok := r.Environments[id]; ok && p.Kind == claircore.BINARY {
+			for _, e := range envs {
+				for _, repoID := range e.RepositoryIDs {
+					repo := r.Repositories[repoID]
+					if repo.Name == rhcc.GoldRepo.Name {
+						problematicPkgs[id] = e.IntroducedIn.String()
```
RTann, line 33: should we `break` here, too? Actually, this doesn't happen in practice, but in theory: what if there were two environments for this package which relate back to the `rhcc.GoldRepo` and are in different layers? Which layer will we show (not sure if layer order is guaranteed when looking at the envs)? If this isn't even possible nor do we want to deal with it theoretically, shouold we just break out of the `e.RepositoryIDs` loop?

enricher/rhcc/rhcc_test.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+func Digest(name string) claircore.Digest {
+	h := sha256.New()
+	io.WriteString(h, name)
```
RTann, line 19: nit: `panic` on error here, too?

enricher/rhcc/rhcc.go
crozzy: Yeah fair enough, the variable was existing based on the old implementation that identified vulns coming from "problematic packages" (packages that are from rhcc layers but not RH content).

enricher/rhcc/rhcc.go
crozzy: Yes, I can add a break. It likely always going to be one Repository anyway but at least the code is clearer

hdonnay (approved):
SGTM

Let's rescue this one.

PR #652, opened 2022-07-08
fetcher: unlinked file based fetcher

hdonnay:
Signed-off-by: Hank Donnay <hdonnay@redhat.com>

PR #1061, opened 2023-09-12
fetcher: new implementation using unlinked files

hdonnay:
I've tried to do this like 3 times and think I finally hit on a way to only gradually break the API.

## Todo:

- [x] Fix tests
- [x] Coverage pass
- [x] Wait for 1.19 to be removed from CI


hdonnay:
Supersedes and closes #652.

hdonnay:
I think this is in a place to get reviewed if you can find time. What it does (roughly in order):

- Move decompression logic to a dedicated package
- Modify the Layer type and add a LayerDescription; the former has state, the latter doesn't.
- Add a package to hold API migration helpers and add some helpers.
- Add a new fetcher interface that uses LayerDescription
- Implement a test fetcher with the new interface
- Implement a real fetcher with the new interface and using `O_TMPFILE`
- Unify the various ways of getting Layers out of the `test` hierarchy
- Update tests

There's still work to be done to move everything over to using the `FS` method of the `Layer`, but this is already very large. I can do that in a "stacked" PR, or this one, or later.

hdonnay:
I can also cherry-pick some the of not _strictly_ related `test` changes into other PRs if any reviewer would rather.

hdonnay:
Proposed changelog entry:
* * *
libindex: move to O_TMPFILE fetcher

This release uses a new fetcher (the component responsible for pulling layers
locally) that makes use of the O_TMPFILE flag to open(2). This ensures that
layer files will be cleaned up even in the event of an unclean shutdown,
including being sent a KILL signal.

libindex/fetcher.go
```diff
@@ -54,162 +55,221 @@ type RemoteFetchArena struct {
...
+		var err error
+		defer func() {
+			if err != nil {
+				span.SetStatus(codes.Ok, "")
+				return
 			}
-			ff = res.Val.(string)
-		case <-ctx.Done():
-			return ctx.Err()
+			span.SetStatus(codes.Error, "fetchInto error")
```
crozzy, line 163: maybe these guys should be inside the err check

crozzy (approved):
Concerns addressed face to face, LGTM

PR #1063, opened 2023-09-15
chore: audit `errgroup` usage

hdonnay:
While looking around in the code for `errors.Join` reimplementations, I noticed some other odd error-handling patterns, mostly around `errgroup`.

There's currently two `errgroup`+`semaphore` usages that I didn't touch: `libvuln/updates` and `datastore/postgres`. In both cases they're using the same pattern, and in both cases they're due to have more drastic changes than refactoring the internal concurrency.

## Todo
- [x] Update tests for better coverage.

### DIFF
--- a/indexer/controller/coalesce.go
+++ b/indexer/controller/coalesce.go
@@ -11,75 +11,79 @@ import (
 	"github.com/quay/claircore/indexer"
 )
 
-// coalesce calls each ecosystem's coalescer and merges the returned IndexReports
+// Coalesce calls each ecosystem's Coalescer and merges the returned IndexReports.
 func coalesce(ctx context.Context, s *Controller) (State, error) {
-	cctx, cancel := context.WithCancel(ctx)
-	defer cancel()
 	mu := sync.Mutex{}
 	reports := []*claircore.IndexReport{}
-	g := errgroup.Group{}
-	// dispatch a coalescer go routine for each ecosystem
+	g, gctx := errgroup.WithContext(ctx)
+	// Dispatch a Coalescer goroutine for each ecosystem.
 	for _, ecosystem := range s.Ecosystems {
+		select {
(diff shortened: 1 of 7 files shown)


indexer/controller/coalesce.go
```diff
@@ -11,75 +11,79 @@ import (
...
+		select {
+		case <-gctx.Done():
+			break
+		default:
+		}
 		artifacts := []*indexer.LayerArtifacts{}
-		pkgScanners, _ := ecosystem.PackageScanners(cctx)
-		distScanners, _ := ecosystem.DistributionScanners(cctx)
-		repoScanners, _ := ecosystem.RepositoryScanners(cctx)
+		pkgScanners, _ := ecosystem.PackageScanners(gctx)
```
crozzy, line 27: I guess the idea behind sharing this context is to be able to early out if the Coalesce goroutine fails and execution has already passed the `gctx.Done()` check?

indexer/controller/coalesce.go
```diff
@@ -11,75 +11,79 @@ import (
...
+			// Pack artifacts array in layer order.
 			artifacts = append(artifacts, la)
 		}
-		coalescer, err := ecosystem.Coalescer(cctx)
+		coalescer, err := ecosystem.Coalescer(gctx)
 		if err != nil {
 			return Terminal, fmt.Errorf("failed to get coalescer from ecosystem: %v", err)
 		}
-		// dispatch coalescer
+		// Dispatch.
```
crozzy, line 84: Not a part of this PR but I realized during this review that this pattern is pretty strange, we do all the expensive DB look ups for the coalescing context in serial and then do the (pretty cheap) coalescing in parallel, I suspect it's impact is negligible, what do you think?

crozzy (commented):
Looks good, just a couple of comments

PR #1064, opened 2023-09-19
gobin: improve handling of go versions

hdonnay:
This replaces the "runtime" package with its proper name, and documents and extends the main module handling.

See-also: PROJQUAY-5912

### DIFF
--- a/gobin/exe.go
+++ b/gobin/exe.go
@@ -49,20 +49,24 @@ func toPackages(ctx context.Context, out *[]*claircore.Package, p string, r io.R
 
 	// TODO(hank) This package could use canonical versions, but the
 	// [claircore.Version] type is lossy for pre-release versions (I'm sorry).
+
+	// TODO(hank) The "go version" is documented as the toolchain that produced
+	// the binary, which may be distinct from the version of the stdlib used?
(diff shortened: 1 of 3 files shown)


gobin/exe.go
```diff
@@ -70,15 +75,50 @@ func toPackages(ctx context.Context, out *[]*claircore.Package, p string, r io.R
...
+				}
+			case "vcs.time":
+				v = append(v, "built at "+s.Value)
+			case "vcs.modified":
+				switch s.Value {
+				case "true":
+					v = append(v, "dirty")
+				case "false":
+				default:
+					panic("unreachable")
```
crozzy, line 111: Do we want to risk the, however tiny, chance that someone is doing something messed up with their compiler and is able to down an indexer?

crozzy (commented):
LGTM, just one comment about the panic, I can't think of a situation that it wouldn't be "true" or "false" but most esoteric stuff we've seen has been hard to predict

gobin/exe.go
hdonnay: Good point, easy change.

hdonnay:
Will merge with green CI.

Issue #1065, opened 2023-09-20
jar: fix `Source` heuristic


hdonnay:
From a recent troubleshooting request, `jar.Parse` reports 

```
email-service-4.0.9.235707a.jar:rest-management-private-classpath/META-INF/maven/com.fasterxml.jackson.core/jackson-databind/pom.properties
```

for a source, but the logs say

```
{"level":"debug","parent":"email-service-4.0.9.235707a.jar","component":"java/jar/Parse","jar":"BOOT-INF/lib/ehcache-2.10.4.jar","path":"rest-management-private-classpath/META-INF/maven/com.fasterxml.jackson.core/jackson-databind/pom.properties","time":"2023-09-20T17:00:32-05:00","message":"found properties file"}
```

meaning I'd expect to to be reported as

```
email-service-4.0.9.235707a.jar:BOOT-INF/lib/ehcache-2.10.4.jar:rest-management-private-classpath/META-INF/maven/com.fasterxml.jackson.core/jackson-databind/pom.properties
```

It looks like a path component is getting dropped somewhere.

hdonnay:
Script for checking:

```go
///usr/bin/env go mod init jardump; trap 'rm go.mod go.sum' EXIT; /usr/bin/env go mod tidy && /usr/bin/env go run "$0" "$@"; exit "$?"

package main

import (
	"archive/zip"
	"context"
	"fmt"
	"log"
	"os"

	"github.com/quay/claircore/java/jar"
)

func main() {
	ctx := context.Background()
	f, err := os.Open(os.Args[1])
	if err != nil {
		log.Panic(err)
	}
	defer f.Close()
	fi, err := f.Stat()
	if err != nil {
		log.Panic(err)
	}
	z, err := zip.NewReader(f, fi.Size())
	if err != nil {
		log.Panic(err)
	}
	infos, err := jar.Parse(ctx, os.Args[1], z)
	if err != nil {
		log.Panic(err)
	}
	for _, i := range infos {
		fmt.Printf("%s\t%s@%s\n", i.Source, i.Name, i.Version)
	}
}
```

PR #1066, opened 2023-09-21
jar: improve recursive path tracking

hdonnay:
This should provide for more accurate details on where packages are discovered, especially when recursing into embedded jars.

Closes: #1065


hdonnay:
Yeah, good catch.

crozzy (approved):

PR #1075, opened 2023-09-25
datastore: refactor deleteManifests to deal with large datasets

crozzy:
Update the delete logic to try and follow the existing indexes. This splits the functionality into 4 calls, get manifest ID, get layer IDs, delete manifest, delete layers.

TODO:

- [ ] Logging feedback after delete
- [ ] Re add instrumentation
- [ ] Deal with stacking defers
- [ ] Enclose in a transaction

### DIFF
--- a/datastore/postgres/deletemanifests.go
+++ b/datastore/postgres/deletemanifests.go
@@ -2,7 +2,10 @@ package postgres
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
+	"github.com/jackc/pgx/v4"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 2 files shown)


datastore/postgres/deletemanifests.go
```diff
@@ -33,55 +36,58 @@ var (
...
+			}
+			lIDs = append(lIDs, layerID)
+		}
+		// TODO: feedback for how things went
+		// Delete manifest
+		_, err = s.pool.Exec(ctx, deleteManifest, manifestID)
+		if err != nil {
+			return nil, fmt.Errorf("unable to delete manifest: %w", err)
+		}
+		// Delete layers
```
crozzy, line 84: This is actually wrong, these layers could also be associated with other manifests.

PR #1076, opened 2023-09-27
datastore: refactor deleteManifests to deal with large datasets

crozzy:
Update the delete logic to try and follow the existing indexes. This splits the functionality into 4 calls, get manifest ID, get layer IDs, delete manifest, delete layers.

### DIFF
--- a/datastore/postgres/deletemanifests.go
+++ b/datastore/postgres/deletemanifests.go
@@ -2,7 +2,10 @@ package postgres
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
+	"github.com/jackc/pgx/v4"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/quay/zlog"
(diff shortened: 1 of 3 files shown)


crozzy:
`explain analyse SELECT id FROM manifest WHERE hash = 'sha256:3668078d16ac1cdb28544554a0578f71bb83b09c1e9c0666f12079d61826110e'`

```
Index Scan using manifest_hash_unique on manifest  (cost=0.15..8.17 rows=1 width=8) (actual time=0.019..0.020 rows=1 loops=1)
  Index Cond: (hash = 'sha256:3668078d16ac1cdb28544554a0578f71bb83b09c1e9c0666f12079d61826110e'::text)
Planning Time: 0.071 ms
Execution Time: 0.038 ms
```

`explain analyse SELECT layer_id FROM manifest_layer WHERE manifest_id = 4;`

```
Bitmap Heap Scan on manifest_layer  (cost=4.21..14.37 rows=8 width=8) (actual time=0.033..0.036 rows=4 loops=1)
  Recheck Cond: (manifest_id = 4)
  Heap Blocks: exact=1
  ->  Bitmap Index Scan on manifest_layer_pkey  (cost=0.00..4.21 rows=8 width=0) (actual time=0.019..0.020 rows=4 loops=1)
        Index Cond: (manifest_id = 4)
Planning Time: 0.114 ms
Execution Time: 0.084 ms
```

`explain analyse DELETE FROM manifest WHERE id = 4;`

```
Delete on manifest  (cost=0.15..8.17 rows=1 width=6) (actual time=0.028..0.028 rows=0 loops=1)
  ->  Index Scan using manifest_pkey on manifest  (cost=0.15..8.17 rows=1 width=6) (actual time=0.010..0.011 rows=1 loops=1)
        Index Cond: (id = 4)
Planning Time: 0.050 ms
Trigger for constraint indexreport_manifest_id_fkey: time=0.120 calls=1
Trigger for constraint manifest_index_manifest_id_fkey: time=1.681 calls=1
Trigger for constraint manifest_layer_manifest_id_fkey: time=0.049 calls=1
Trigger for constraint scanned_manifest_manifest_id_fkey: time=0.126 calls=1
Execution Time: 2.020 ms
```

`explain analyse DELETE FROM layer WHERE id = ANY('{9,10,11,12}');`

```
Delete on layer  (cost=8.64..17.11 rows=4 width=6) (actual time=0.078..0.078 rows=0 loops=1)
  ->  Bitmap Heap Scan on layer  (cost=8.64..17.11 rows=4 width=6) (actual time=0.009..0.010 rows=4 loops=1)
        Recheck Cond: (id = ANY ('{9,10,11,12}'::bigint[]))
        Heap Blocks: exact=1
        ->  Bitmap Index Scan on layer_pkey  (cost=0.00..8.64 rows=4 width=0) (actual time=0.004..0.004 rows=4 loops=1)
              Index Cond: (id = ANY ('{9,10,11,12}'::bigint[]))
Planning Time: 0.138 ms
Trigger for constraint dist_scanartifact_layer_id_fkey: time=0.336 calls=4
Trigger for constraint manifest_layer_layer_id_fkey: time=0.113 calls=4
Trigger for constraint package_scanartifact_layer_id_fkey: time=1.187 calls=4
Trigger for constraint repo_scanartifact_layer_id_fkey: time=0.190 calls=4
Trigger for constraint scanned_layer_layer_id_fkey: time=0.230 calls=4
Trigger for constraint file_scanartifact_layer_id_fkey: time=0.211 calls=4
Execution Time: 2.381 ms
```

crozzy:
`explain analyse select l.id from layer l left join manifest_layer ml on l.id = ml.layer_id where l.id = 22049 and ml.layer_id is null`

```
Nested Loop Anti Join  (cost=1.15..2332.28 rows=1 width=8) (actual time=0.030..0.032 rows=1 loops=1)
  Join Filter: (l.id = ml.layer_id)
  ->  Index Only Scan using layer_pkey on layer l  (cost=0.57..8.59 rows=1 width=8) (actual time=0.019..0.020 rows=1 loops=1)
        Index Cond: (id = 22049)
        Heap Fetches: 0
  ->  Index Only Scan using manifest_layer_layer_idx on manifest_layer ml  (cost=0.58..2323.67 rows=41510 width=8) (actual time=0.009..0.009 rows=0 loops=1)
        Index Cond: (layer_id = 22049)
        Heap Fetches: 0
Planning Time: 0.129 ms
Execution Time: 0.054 ms```

**Comment by crozzy** [comment]
Adding the index concurrently on a production snapshot from June took ~8 hours, this is bound to be larger in prod with read/write traffic also competing for priority. If we go down this route I'll add another admin job to Clair

File: datastore/postgres/migrations/indexer/08-index-manifest_layer.sql
Comment (hdonnay) [inline datastore/postgres/migrations/indexer/08-index-manifest_layer.sql:1]: I think this can just be a hash index? We just need to know the existence, so any ordering is just extra time building the index.

Comment (crozzy) [reply datastore/postgres/migrations/indexer/08-index-manifest_layer.sql:1]: I let it run for over a week and the index was still not near being created. I eventually stopped it yesterday to test the admin command, it was here: 

| blocks_total | blocks_done | tuples_total | tuples_done |
| ----------- | ----------- | ----------- | ----------- |
| 13641437 | 13641437 | 2132494811 | 380262621 |

File: datastore/postgres/deletemanifests.go [hunk 1]
```diff
@@ -33,55 +36,83 @@ var (
 
 func (s *IndexerStore) DeleteManifests(ctx context.Context, d ...claircore.Digest) ([]claircore.Digest, error) {
 	ctx = zlog.ContextWithValues(ctx, "component", "datastore/postgres/DeleteManifests")
-	rm, err := s.deleteManifests(ctx, d)
-	if err != nil {
-		return nil, err
-	}
-	return rm, s.layerCleanup(ctx)
+	return s.deleteManifests(ctx, d)
```
hdonnay, line 39: Just flatten this to a single method.

datastore/postgres/deletemanifests.go
```diff
@@ -33,55 +36,83 @@ var (
...
+ON l.id = ml.layer_id 
+WHERE l.id = $1
+AND ml.layer_id IS NULL;`
+		deleteManifest = `DELETE FROM manifest WHERE id = $1;`
+		deleteLayers   = `DELETE FROM layer WHERE id = ANY($1);`
+	)
+
+	deletedManifests := make([]claircore.Digest, 0, len(ds))
+	for _, d := range ds {
+		s.pool.BeginFunc(ctx, func(tx pgx.Tx) error {
```
hdonnay, line 58: What's the thinking on doing a transaction per loop, rather than a loop in a transaction?

datastore/postgres/deletemanifests.go
```diff
@@ -33,55 +36,83 @@ var (
...
+			for lRows.Next() {
+				var layerID int64
+				err = lRows.Scan(&layerID)
+				if err != nil {
+					return fmt.Errorf("unable to scan layer ID: %w", err)
+				}
+				lIDs = append(lIDs, layerID)
+			}
+			lRows.Close()
+			// TODO: feedback for how things went
```
hdonnay, line 87: ???

datastore/postgres/deletemanifests.go
```diff
@@ -33,55 +36,83 @@ var (
...
-	if err != nil {
-		return err
+			// Get all layer IDs
+			lRows, err := tx.Query(ctx, getLayers, manifestID)
+			if err != nil {
+				return fmt.Errorf("unable to query layers: %w", err)
+			}
+			defer lRows.Close()
+			lIDs := []int64{}
+			for lRows.Next() {
```
hdonnay, line 78: Need to check `lRows.Err()`

datastore/postgres/deletemanifests.go
```diff
@@ -33,55 +36,83 @@ var (
...
+	const (
+		getManifestID      = `SELECT id FROM manifest WHERE hash = $1`
+		getLayers          = `SELECT layer_id FROM manifest_layer WHERE manifest_id = $1;`
+		getDeletableLayers = `
+SELECT l.id FROM layer l 
+LEFT JOIN manifest_layer ml 
+ON l.id = ml.layer_id 
+WHERE l.id = $1
+AND ml.layer_id IS NULL;`
+		deleteManifest = `DELETE FROM manifest WHERE id = $1;`
```
hdonnay, line 52: These two commands should be able to be composed, eg `DELETE FROM manifest WHERE id IN (SELECT l.id FROM layer l 
LEFT JOIN manifest_layer ml 
ON l.id = ml.layer_id 
WHERE l.id = $1
AND ml.layer_id IS NULL);`

Not sure if it makes a ton of difference, as it just saves a roundtrip in a not-frequently used call.

hdonnay (changes requested):
In addition to the specific comments, the rewritten method needs a logs and metrics pass. They don't have to be the same exact messages and metrics, but it needs some observability.

datastore/postgres/deletemanifests.go
crozzy: My theory was that each manifest deletion can be logically broken down into one transaction's worth of work and it makes the code slightly cleaner when trying not to stack up deferred `Close()`s.

datastore/postgres/deletemanifests.go
crozzy: I'm not sure that'd work because we need to Delete the manifest first and let it cascade to the `manifest_layer` table to be able to surface the potentially delete-able layers, maybe the `getDeletableLayers` and `deleteLayers` queries could be combined.

crozzy: Re-reading your comment, maybe that's what you meant?

hdonnay: Yes, that's what I meant

hdonnay (approved):

PR #1077, opened 2023-09-28
test: DB harness and timeouts

hdonnay:
This is a raft of changes made when trying to hunt down some CI flakes.

I _think_ they may have been cause by Context timeout/cancels, so these changes remove them.
IIRC, these were added originally to catch come hangs in tests, but we should use the `-timeout` flag for that.

In a related change, I ported in some changes to the `test/postgres` functions from a database modernization branch I've got in my pocket.


test/postgres/doc.go
```diff
@@ -0,0 +1,18 @@
...
+// The [TestIndexerDB] and [TestMatcherDB] start database engines and create
+// suitable databases as needed. See the [test/integration] package for more
+// information on the specifics of the former.
+//
+// As part of the database setup, these functions can load SQL files. This
+// package searches both its embedded copy of the local "sql" directory and the
+// main test package's "testdata" director for these patterns:
+//   - all_*.psql
+//   - matcher_*.psql (for [TestMatcherDB])
+//   - indexer_*.psql (for [TestIndexerDB])
```
hdonnay, line 17: While writing the documentation, I thought through that it would be annoying that, if/when we have multiple database backends, all the SQL has to be in some dialect that's database agnostic because I made the mistake of camping on the `.sql` extension for the helper files. So, I changed this to use a ".psql" extension and added the metadata to make sure GitHub marks it as SQL.

crozzy: SGTM

crozzy (approved):

PR #1084, opened 2023-10-02
cicd: add cronjob to help with the `needs-changelog` label

hdonnay:
This is a github action to automatically remove the `needs-changelog` label.

All the parts have been tested piecewise, but this may require some followups when actually running in the runner environment.


crozzy (approved):
I can't see anything wrong here but, full disclosure, I'm not a shell scripting guru

PR #1086, opened 2023-10-02
toolkit: wfn account for empty strings

crozzy:
We need to be able to deal with empty strings when constructing WFNs. This means making sure the Scan() method knows not to raise an error and the String() method also need to check for whether the WFN is valid to ensure symmetry.


hdonnay (changes requested):
Two suggestions, but only the imports getting goofed needs to get fixed.

hdonnay (changes requested):
Commits need to get squashed, files look good.

PR #1089, opened 2023-10-03
chore: update toolkit to latest version

crozzy:
This updates the toolkit module to v1.1.1, which includes a small fix.


hdonnay:
Warrants a changelog entry mentioning that this fixes a bug introduced in 1.5.17

hdonnay (approved):

PR #1091, opened 2023-10-05
alpine: support edge

RTann:


alpine/distributionscanner.go
```diff
@@ -112,43 +131,53 @@ func (*DistributionScanner) scanFs(ctx context.Context, sys fs.FS) (d []*clairco
...
+	b, err := fs.ReadFile(sys, issuePath)
 	switch {
 	case errors.Is(err, nil):
-		// parse here
+		if isEdge := edgeIssueRegexp.Match(b); isEdge {
+			return &claircore.Distribution{
+				Name:       `Alpine Linux`,
+				DID:        `alpine`,
+				Version:    `edge`,
+				PrettyName: `Alpine Linux edge`,
```
crozzy, line 157: Could use the variable here ^^

crozzy (commented):
The code seems fine to me. It's worth mentioning that we don't currently ingest any edge vulnerabilities so these distros won't match to anything, but being able to identify the edge distros does seem useful.

RTann:
> The code seems fine to me. It's worth mentioning that we don't currently ingest any edge vulnerabilities so these distros won't match to anything, but being able to identify the edge distros does seem useful.

I think at least identifying could be useful. I can create another PR to start capturing edge vulns, and we can decide if it's worth having or not

crozzy (approved):

PR #1093, opened 2023-10-06
gobin: check for strconv.ErrRange when coverting to semver

crozzy:
The semver library will try to convert the separate version parts to int64, this can return errors (particularly when the version contains timestamps like 2023080721291691443750). We also want ints for our internal version type so we can't really use these versions without losing something.

I have another branch that starts a semver package but that will require more design thought.

### DIFF
--- a/gobin/exe.go
+++ b/gobin/exe.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strconv"
 	"strings"
 	_ "unsafe" // for error linkname tricks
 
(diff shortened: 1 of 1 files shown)


hdonnay (changes requested):
Can I trouble you to add a test, please.

crozzy:
https://issues.redhat.com/browse/PROJQUAY-6111

> Can I trouble you to add a test, please.

Certainly

crozzy:
Started writing tests, realized we'd need something like this: https://github.com/Masterminds/semver/pull/220/files, will see if there is traction there.

crozzy:
Closing in favor of https://github.com/quay/claircore/pull/1253

PR #1101, opened 2023-10-09
alpine: support edge vulnerabilities

RTann:
Followup to #1091 to support Alpine edge vulnerabilities

### DIFF
--- a/alpine/distributionscanner.go
+++ b/alpine/distributionscanner.go
@@ -26,11 +26,10 @@ const (
 	scannerName    = "alpine"
 	scannerVersion = "3"
 	scannerKind    = "distribution"
-)
 
-const (
 	issuePath = `etc/issue`
 
+	edgeVersion    = `edge`
 	edgePrettyName = `Alpine Linux edge`
 )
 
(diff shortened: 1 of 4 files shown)


RTann:
I updated my local Clair repo to use this version of ClairCore, and I ran the matcher. I found logs showing it was running the Alpine edge updater

alpine/distributionscanner.go
```diff
@@ -32,6 +32,7 @@ const (
 const (
 	issuePath = `etc/issue`
 
+	edgeVersion    = `edge`
```
crozzy, line 35: can we consolidate both of these `const`s?

crozzy (commented):
LGTM, just one nit

alpine/distributionscanner.go
RTann: not sure I follow

RTann: @crozzy do you have any updates?

crozzy: I meant have one `const` block instead of 2

RTann: oh didn't see that, will do

RTann: should be good now

hdonnay (approved):

PR #1105, opened 2023-10-13
all: use `Layer.FS` method

hdonnay:
Doing this should allow us to realize potential memory savings enabled by #1061.

All the in-tree Scanner implementations should now share a single `tarfs.FS` instance per layer.

### DIFF
--- a/dpkg/distroless_scanner.go
+++ b/dpkg/distroless_scanner.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/indexer"
-	"github.com/quay/claircore/pkg/tarfs"
 )
 
 const (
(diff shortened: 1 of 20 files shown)


dpkg/distroless_scanner.go
```diff
@@ -63,14 +62,9 @@ func (ps *DistrolessScanner) Scan(ctx context.Context, layer *claircore.Layer) (
...
-	rd, err := layer.Reader()
+	sys, err := layer.FS()
 	if err != nil {
-		return nil, fmt.Errorf("opening layer failed: %w", err)
-	}
-	defer rd.Close()
-	sys, err := tarfs.New(rd)
-	if err != nil {
-		return nil, fmt.Errorf("opening layer failed: %w", err)
+		return nil, fmt.Errorf("dpkg: opening layer failed: %w", err)
```
crozzy, line 67: Maybe add that it's the `dpkg-distroless` scanner?

crozzy (approved):
Nice, needs-changelog?

PR #1106, opened 2023-10-13
cicd: tweak per-patch codecov reporting

hdonnay:
I find the "patch" stats failing CI to be very annoying. Refactors that net positive in coverage can still get hit by it. This change still has it reported, but it won't fail the CI.


crozzy (approved):
agreed

PR #1114, opened 2023-10-16
gobin: screen for ELF executables and shared objects

hdonnay:
These are the only ELF files that would have go module information in them, anyway. Go does not use ELF relocatable objects for linking.


crozzy (approved):
LGTM, it seems pretty hard to test for, there is no concrete output for files we looked at (except the logs) so I guess as long as the existing tests pass it's good.

PR #1116, opened 2023-10-18
integration: poison default http resources

hdonnay:
This should help us root out inappropriate uses of the default `net/http` resources.


crozzy (approved):

crozzy:
:ok_hand:

PR #1123, opened 2023-10-24
Move more package scanners to default repo scanners

crozzy:
This work was started with the gobin PackageScanner but never fully realised with all potential ecosystems that could take advantage of just having one Scanner/Detector.


hdonnay (approved):
LGTM

I'd forgotten we (you) added this 😶

hdonnay (approved):

PR #1125, opened 2023-10-26
fix(ruby): use more accurate versioning

RTann:
Implemented Ruby versioning based on https://github.com/rubygems/rubygems/blob/1a1948424aca90013b37cb8a78f5d1d5576023f1/lib/rubygems/version.rb

### DIFF
--- a/ruby/matcher_test.go
+++ b/ruby/matcher_test.go
@@ -106,7 +106,7 @@ func TestVulnerable(t *testing.T) {
 			record: &claircore.IndexRecord{
 				Package: &claircore.Package{
 					Name:    "dependabot-omnibus",
-					Version: "0.120.0-beta2",
+					Version: "0.120.0.beta2",
 					Kind:    "binary",
 				},
 			},
@@ -118,7 +118,7 @@ func TestVulnerable(t *testing.T) {
 					Name:           "dependabot-omnibus",
 					RepositoryHint: "rubygems",
 				},
-				FixedInVersion: "fixed=0.125.1&introduced=0.119.0-beta1",
+				FixedInVersion: "fixed=0.125.1&introduced=0.119.0.beta1",
 			},
 			want: true,
 		},
--- a/ruby/version.go
+++ b/ruby/version.go
@@ -1,0 +1,210 @@
+package ruby
+
+import (
+	"errors"
+	"regexp"
+	"strings"
+)
+
+var (
+	anchoredVersion = regexp.MustCompile(`^\s*([0-9]+(\.[0-9a-zA-Z]+)*(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?)?\s*$`)
+
+	errInvalidVersion = errors.New("invalid gem version")
+)
+
+// Version is a RubyGem version.
+// This is based on the official [implementation].
+//
+// [implementation]: https://github.com/rubygems/rubygems/blob/1a1948424aca90013b37cb8a78f5d1d5576023f1/lib/rubygems/version.rb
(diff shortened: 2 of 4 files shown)


ruby/version.go
```diff
@@ -0,0 +1,162 @@
...
+		seg, ok := v.segs[i].(numericSegment)
+		if !ok || seg != 0 {
+			break
+		}
+	}
+	v.segs = v.segs[:i+1]
+
+	// Remove all zero segments preceding the first letter in a prerelease version.
+	if v.prerelease {
+		var start *int
```
RTann, line 124: Note: this does not account for 1.0.0.alpha == 1.alpha at this time, so need to go over this again

ruby/matcher_test.go
```diff
@@ -106,7 +106,7 @@ func TestVulnerable(t *testing.T) {
 			record: &claircore.IndexRecord{
 				Package: &claircore.Package{
 					Name:    "dependabot-omnibus",
-					Version: "0.120.0-beta2",
```
RTann, line 109: This was an invalid semver before, so I switched this to ensure our Rubygem version sees this as valid

ruby/matcher_test.go
```diff
@@ -106,7 +106,7 @@ func TestVulnerable(t *testing.T) {
...
+					Version: "0.120.0.beta2",
```
RTann, line 109: This was an invalid semver before, so I switched this to ensure our Rubygem version sees this as valid

crozzy: https://github.com/dependabot/dependabot-core/security/advisories/GHSA-23f7-99jx-m54r for reference

ruby/version.go
```diff
@@ -0,0 +1,170 @@
...
+	return segs
+}
+
+func partitionSegments(v string) (segs []Segment, prerelease bool) {
+	for _, s := range strings.Split(v, ".") {
+		if s == "" {
+			continue
+		}
+		n, err := strconv.ParseUint(s, 10, 64)
+		if err != nil {
```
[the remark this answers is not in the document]
RTann: I'm not sure how we'd handle `ErrRange`, unfortunately. Just not support it?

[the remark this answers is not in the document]
RTann: Another option would be to keep everything in string form and regex for digits and letters to determine how to proceed. It'd be safest, but probably less efficient since it deals with regexes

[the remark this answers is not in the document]
RTann: It's annoying, but it seems like Ruby allows any arbitrarily long sequence of numbers be an `Integer`. Doesn't seem like there is a concept of `max_int`, so I guess it's possible this number is way larger than the largest 64 bit number.

I was considering keeping those arbitrarily long integers as strings and comparing from there, but I don't want to implement that comparison... Maybe we can just not support it?

RTann:
@crozzy I decided to just not support Ruby versions with arbitrarily long integer sequences. Let me know if we should reconsider this, though.

crozzy:
> @crozzy I decided to just not support Ruby versions with arbitrarily long integer sequences. Let me know if we should reconsider this, though.

IIUC the discussed solution was to compare padded strings?

RTann:
> > @crozzy I decided to just not support Ruby versions with arbitrarily long integer sequences. Let me know if we should reconsider this, though.
> 
> IIUC the discussed solution was to compare padded strings?

Yep. Just added that now, if you can take a look

crozzy (approved):
LGTM

RTann:
Rebased. Will merge with CI

PR #1126, opened 2023-10-26
layerscanner: add layer hash to error context

crozzy:
This is a frustration in environments where we can see an error surfacing but it's difficult to tie it to a specific layer for debugging purposes.

### DIFF
--- a/indexer/layerscanner.go
+++ b/indexer/layerscanner.go
@@ -145,7 +145,10 @@ func (ls *LayerScanner) Scan(ctx context.Context, manifest claircore.Digest, lay
 				return context.Cause(ctx)
 			default:
 			}
-			return ls.scanLayer(ctx, l, s)
+			if err := ls.scanLayer(ctx, l, s); err != nil {
+				return fmt.Errorf("layer %q: %w", l.Hash, err)
+			}
+			return nil
 		}
 	}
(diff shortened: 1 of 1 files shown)


indexer/layerscanner.go
```diff
@@ -145,7 +145,10 @@ func (ls *LayerScanner) Scan(ctx context.Context, manifest claircore.Digest, lay
 				return context.Cause(ctx)
 			default:
 			}
-			return ls.scanLayer(ctx, l, s)
+			if err := ls.scanLayer(ctx, l, s); err != nil {
+				return fmt.Errorf("failed to scan layer %s: %w", l.Hash, err)
```
hdonnay, line 149:
```suggestion
				return fmt.Errorf("%s: %w", l.Hash, err)
```

The caller prepends their own context, which I think is something like "failed to scan all layer contents:", so reiterating that a _layer scan_ is what failed seems to add very little?

hdonnay (commented):
Suggestion on phrasing, unless you have a reason for it to be this way.

indexer/layerscanner.go
crozzy: I noticed we also use `%q` in other places for `layer.Hash`s so switched that too

hdonnay (approved):
LGTM

Issue #1124, opened 2023-10-26
Timeout while rescanning a previously scanned manifest leaves IndexReport empty forever


iainduncani:
## The Problem

We are finding that sometimes when we scan a manifest that has already been scanned we will end up with an empty IndexReport stored for that manifest:

```json
{
  "manifest_hash": "sha256:a2f196cc1f224ba665cde842a5be1d3c7ded333c279525242eba123677152172",
  "state": "",
  "packages": {},
  "distributions": {},
  "repository": {},
  "environments": {},
  "success": false,
  "err": ""
}
```

Once in this state any further request to rescan it will not do anything to change the IndexReport.  The logs from the call to scan the previously scanned manifest show:

```
info index request start
debug locking attempt
debug locking OK
info starting scan
info index request done
```

The logs on any future attempts show an almost identical output except they include the `manifest already scanned` line:

```
info index request start
debug locking attempt
debug locking OK
info starting scan
info manifest already scanned
info index request done 
```

## Suspected Code Path

Looking at the code I think this happens when an attempt to get the `IndexReport` in `checkManifest` timeouts.  The flow I believe is:

1. The `Controller` enters the [run method and calls `CheckManifest`](https://github.com/quay/claircore/blob/v1.5.19/indexer/controller/controller.go#L87) as the first state. At this point the `IndexReport` on the controller is empty
2. `checkManifest` discovers that the manifest is [already scanned](https://github.com/quay/claircore/blob/v1.5.19/indexer/controller/checkmanifest.go#L28-L31)
3. `checkManifest` attempts to load the `IndexReport` but the call to `s.Store.IndexReport` [times out (5s timeout) and an err is returned](https://github.com/quay/claircore/blob/v1.5.19/indexer/controller/checkmanifest.go#L65-L68).  `checkManifest` does not log anything in this case and `IndexReport` on the controller is not set because it was not loaded.
4. Back in the calling controller it goes through the checks as to whether an error occurred.  As this is a timeout `errors.Is(err, context.DeadlineExceeded)` will be true and it will [go into that case](https://github.com/quay/claircore/blob/v1.5.19/indexer/controller/controller.go#L100-L103)
5. As the timeout case does not exit the `for` loop early it just sets the `retry` variable to true the [`Controller` then proceeds to store the `IndexReport` into the database](https://github.com/quay/claircore/blob/v1.5.19/indexer/controller/controller.go#L116-L124).  This stores an empty `IndexReport` into the database.
[...]

PR #1131, opened 2023-10-31
Check manifest to check index report

iainduncani:
This is a proposed solution to  #1124.  This is just a draft as I would like feedback as to whether this is the right approach and will add tests if this is along the right lines.  The PR does two things:

1. Removes the retry logic from the Indexer's controller.  The retry logic never actually triggered a retry, any request that may timeout also returned the Terminal next state which is processed after the `if retry` section to break out of the for loop and exit.  Having the retry in there meant that empty index reports could be persisted if a query timed out in `checkManifest` (see the issue description).  There was then two options
   1. Fix the retry logic
   2. Delete the retry logic

   I opted for deleting the retry logic as it simplified the code.  If you would prefer the retry logic was fixed then I think it would just need a `continue` adding to the loop and fixing the persisting of an empty manifest when this state occurs (probably with a retry check when calling `SetIndexReport`).  I don't know if consideration would also want to be given to how many times a retry happened to not get stuck in this state.  All these thoughts made me think simpler code was better code so deleted it!
2. If all `Scanner`s complete for a manifest then the index report should be persisted in the `IndexFinished` state with all the required information in it for querying.  If it does not end up in this state it will not be possible to query the manifest but the current implementation of `checkManifest` would also not allow for a rescan.  I have updated `checkManifest` to make sure that the manifest is in `IndexFinished` state and if not it will reprocess the whole image from scratch.  My main question with this part of the change is whether `FetchLayers` is the right state to go into.  By the time we have check the index report we know all the `Scanner`s have database information for them so would we be better skipping to `Coalesce` so that we do not need to pull the image again?  I opted for `FetchLayers` for two reasons, both of which I suspect someone more familiar with the code may disagree with:
   1. We don't necessarily know when the index report did not end up in the right state, I felt FetchLayers was the safest option for handling an unexpected state but this only holds true if there was any situation where `ManifestScanned` returns true but not everything is in the database.
   2. I was worried there may be information required in `Coalesce` that is only set in previous states rather than being in the database.

Would love the thoughts of someone familiar with this code!


paulaldridge:
Would really love to see this improvement. Today after an upgrade problem we have a load of manifests with a committed index report with `State: IndexError`, and `success: false`. But resubmitting the manifest for another scan attempt just goes down the `Manifest already scanned` path, so they are stuck with bad index reports. 

@hdonnay @crozzy any chance of getting this prioritised at some point? Thanks :)

PR #1137, opened 2023-11-07
fix(libindex): use different temp perms per OS

RTann:
ClairCore 1.5.20 did not build on macOS due to non-existent file perms. This PR aims to correct that.


RTann:
This is not complete, as the fetcher reads from `proc/self/fd/FD`, which doesn't exist on Darwin. Instead it should be `/dev/fd/FD` (I think), but I keep getting a permission error, so I'm thinking this strategy may not work for darwin

RTann:
closing this in favor of https://github.com/quay/claircore/pull/1140

PR #1140, opened 2023-11-07
fix(libindex): temp files on non-linux

RTann:
`O_TMPFILE` does not exist on non-linux unix-based OSes, so I was unable to build ClairCore on my macOS (Darwin) machine. Similarly, I don't think it exists on Windows, but `FILE_FLAG_DELETE_ON_CLOSE` does, which sounds to be about the same.

The implemented solution for Darwin runs into the issue `O_TMPFILE` aimed to fix: ClairCore crashing can still leave files on the system. Since this was previous behavior, I don't see this as a regression, necessarily.

### DIFF
--- a/libindex/tempfile_unix.go
+++ b/libindex/tempfile_unix.go
@@ -1,0 +1,30 @@
+//go:build unix && !linux
+
+package libindex
+
+import (
+	"errors"
+	"os"
+)
+
+type tempFile struct {
+	*os.File
+}
+
+func openTemp(dir string) (*tempFile, error) {
+	f, err := os.CreateTemp(dir, "*.fetch")
+	if err != nil {
+		return nil, err
+	}
+	return &tempFile{
+		File: f,
+	}, nil
+}
+
+func (t *tempFile) Reopen() (*os.File, error) {
+	return os.OpenFile(t.Name(), os.O_RDONLY, 0644)
+}
+
+func (t *tempFile) Close() error {
+	return errors.Join(os.Remove(t.Name()), t.File.Close())
+}
--- a/libindex/fetcher.go
+++ b/libindex/fetcher.go
@@ -21,7 +21,6 @@ import (
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/singleflight"
-	"golang.org/x/sys/unix"
 
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/indexer"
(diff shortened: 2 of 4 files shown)


BradLugo:
Would it be worth adding macOS and Windows builds to CI? e.g., `GOOS=darwin go build ...`

hdonnay:
It seems fine as long as they don't slow down the CI.
Might be better off as a nightly CI.
None of the server components are really supported on windows or darwin, they just need to build for Clair's cli.

RTann:
> None of the server components are really supported on windows or darwin, they just need to build for Clair's cli.

I develop on a mac and can't compile at the moment 😢

hdonnay (changes requested):
I think `(*tempfile).read` would be better named as `Reopen` or `Clone`. `(*ref).Val` should just return a `*os.File` instead of a `*tempFile`, because none of the file handles pulled out of the `rc` should need special handling.

RTann:
> `(*ref).Val` should just return a `*os.File` instead of a `*tempFile`, because none of the file handles pulled out of the `rc` should need special handling.

We need to make sure the file is removed on unix, so we need to call `(*tempFile).Close` instead of `(*os.File).Close`

RTann:
> > `(*ref).Val` should just return a `*os.File` instead of a `*tempFile`, because none of the file handles pulled out of the `rc` should need special handling.
> 
> We need to make sure the file is removed on unix, so we need to call `(*tempFile).Close` instead of `(*os.File).Close`

Oh I see now. I'll change it

libindex/tempfile_unix.go
```diff
@@ -0,0 +1,30 @@
...
+func (t *tempFile) Reopen() (*os.File, error) {
+	return os.OpenFile(t.Name(), os.O_RDONLY, 0644)
+}
+
+func (t *tempFile) Close() error {
+	if err := os.Remove(t.Name()); err != nil {
+		return err
+	}
+	return t.File.Close()
+}
```
hdonnay, line 30:
```suggestion
func (t *tempFile) Close() error {
	return errors.Join(os.Remove(t.Name()), t.File.Close())
}
```

Should probably always remove and close, I think.

hdonnay (changes requested):
I think one last issue and it's good to go

libindex/tempfile_unix.go
RTann: makes sense to me

hdonnay (approved):

PR #1142, opened 2023-11-09
osv: parse severity from database_specific

celek:
If there is no available severity from the advisory, attempt to read the database_specific stanza and
extract the severity
Signed-off-by: Christophe Elek (celek@ca.ibm.com)

### DIFF
--- a/updater/osv/severity.go
+++ b/updater/osv/severity.go
@@ -1,0 +1,49 @@
+// Package osv is an updater for OSV-formatted advisories.
+package osv
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/quay/claircore"
+)
+
+// check if severity (lower case) exists in the Database object if so, use it
+func extractSeverityFromDatabase(defaultSev claircore.Severity, a advisory) (sev claircore.Severity) {
+	advisoryDatabase := a.Database
+	var databaseJSON map[string]json.RawMessage
+	if err := json.Unmarshal([]byte(advisoryDatabase), &databaseJSON); err == nil {
+		lowerDatabaseJSON := make(map[string]json.RawMessage, len(databaseJSON))
+		for key, value := range databaseJSON {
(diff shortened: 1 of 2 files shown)


crozzy:
Interesting, thanks for the PR!

I have a few questions:
* Do you have some examples of this that could be used in testcases?
* The [spec](https://ossf.github.io/osv-schema/#database_specific-field) says that the `database_specific` object is "beyond the scope of the spec", is there a specific vuln DB that has the `database_specific` field spec'ed that we can refer back to?
* Was the severity mapping derived from somewhere?

updater/osv/severity.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/quay/claircore"
+)
+
+// check if severity (lower case) exists in the Database object if so, use it
+func extractSeverityFromDatabase(defaultSev claircore.Severity, a advisory) (sev claircore.Severity) {
```
crozzy, line 12: I'd say don't pass in the default severity, either let the `Insert` method handle it or return `claircore.Unknown` as the default case in `severityFromString()` like the other `NormalizeSeverity` functions do so it's easier to follow. Also, I'd probably pass in the advisory.Database as the function is called `extractSeverityFromDatabase()`.

AFAICT the named return param is not used.

updater/osv/severity.go
```diff
@@ -0,0 +1,49 @@
...
+package osv
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/quay/claircore"
+)
+
+// check if severity (lower case) exists in the Database object if so, use it
```
crozzy, line 11: nit: we follow the go [function comments](https://tip.golang.org/doc/comment) guide of starting `ExtractSeverityFromDatabase takes an advisory.Database object ...etc`

crozzy:
Closing in favour of: https://github.com/quay/claircore/pull/1294

PR #1143, opened 2023-11-09
cvss: add package for dealing with CVSS scores

hdonnay:
This adds a package for handling CVSS versions 2, 3.0, 3.1, and 4. The envisioned used-case is for calculating scores and canonicalizing vector strings.


hdonnay:
#1144 came in shortly after I sent this.

Here's the criticism of the quick-n-dirty CVSS support in the `osv` package, so I can make sure they're addressed:

- [ ] CVSS v2.0 metric ordering is not respected, invalid vectors goes through without error
- [ ] mandatory metrics must be part of the vector, at least once at most once
- [ ] non-mandatory metrics could be part of the vector, at most once
- [x] obvious invalid CVSS v2.0 score computation

In my defence for the last one, there's no specified qualitative mapping for v2 so I felt OK getting in the ballpark.

* * *

@pandatix If I can harass you over here, I'm unclear on how/if the v3/v4 environmental scores are combined with the base scores. Are the base metrics just ignored if the environmental metrics are present?

hdonnay:
I think this implementation is going to be generous with metric order on ingesting, seeing as the v4 spec is self-contradictory on this point.

pandatix:
Hey, no problem, if I can help you :)

Yes the idea of an Environmental metric that finds its equivalent in the Base metrics group (for instance Attack Vector / AV and Modified Attack Vector / MAV), if the Modified is not specified in a vector string representation or if it is set to Undefined / X and its Base equivalent value is used for further calculations. If not, it is the value of the Base equivalent that is used.

So if MAV != X, take MAV, else take AV :tada:

hdonnay:
Great, that makes sense. :+1:

hdonnay:
Going to make the decision to allow invalid orderings for vector parsing.

Not doing this would mean not being able to implement v4 to spec while using the published examples. I also think the erratum that makes the spec internally consistent will have to allow for parsing in both the currently specified order and the examples order.

pandatix:
Hey, do you saw an invalid-ordered vector in the specification/examples/guide ?

We won't come back on our decision of fixing the CVSS v4 metric order as it is justified by the fact that an unfixed ordering would not enable providing a valid regular expression for use (for front-end/back-end validation, as achieved in the CVE Schema). If you try you will end up with n! combinations and with 32 metrics the regular expression would be terrabytes long.
This is a known problem with the CVSS v3 spec that we adressed through v4.
To properly implement CVSS v4 you must validate ordering.

hdonnay:
Yeah, as I pointed out on RedHatProductSecurity/cvss-v4-calculator#44, all the examples are invalid and the calculator is implemented with the invalid order.

I think having the order specified is the correct choice, but unless all the examples are changed and all existing implementations are changed, any real implementation is going to need to accept both the specified order (`VC/SC/VI/SI/VA/SA`) and the observed order (`VC/VI/VA/SC/SI/SA`). Implementing alternative ordering for *just* these metrics is a lot of work.

Given that this package isn't meant for validation and only emits valid vectors, I think being liberal in accepted input is fine.

pandatix:
The RedHat calculator is the official calculator, and you can easily check that [here](https://github.com/FIRSTdotorg/cvss-v4-calculator). As I already answered on [your issue](https://github.com/RedHatProductSecurity/cvss-v4-calculator/issues/44), the ordering problem lies in the specification and will be fixed soon. 
This official calculator **is valid**.
An implementation that validates `VC / SC / VI / SI / VA / SA` **is not and won't be compliant** as of CVSS v4.0 Section 7.

The examples in the [CVSS v4.0 Examples](https://www.first.org/cvss/v4.0/examples) are valid, so could you please highlight why you don't think so ?

I don't recommend validating **invalid vectors** i.e. the ones that does not fully comply with the CVSS v4.0 specification (Section 7 requires you to not do it too) as it may propagate invalid values through the supply-chain, mostly for safety reasons, and for compliance sake.
As long as I have no power on this PR, I can't decide on anything, I can't tell you to have a Zero-Trust approach on your data feeds, but I don't recommend you to begin implementing the standard so poorly... That is the reason why I wrote a Go module for CVSS :grinning:

hdonnay:
I see there's a new revision of the spec, so I've updated the code. I think it's pretty impolite to describe implementing the specification as written as "implementing the standard so poorly."

I stand by my comment about the javascript implementation, as the weights and specific ordering of "subvectors" can only be read out of the javascript implementation; they're not in the specification proper. Getting the same "highest-score" vector requires looking at the javascript implementation because it's not specified at all.

I don't understand what you're saying about "upward provider pollution." We're always consuming data from vendors so if they're adding additional metrics, we want to make use of them.

toolkit/types/cvss/cvss_v2.go
```diff
@@ -0,0 +1,247 @@
...
+	case V2Exploitability:
+		switch c {
+		case 'P':
+			return "POC"
+		case 'N':
+			return "ND"
+		}
+	case V2RemediationLevel:
+		switch c {
+		case 'O', 'T':
```
crozzy, line 59: Shouldn't this be:
```go
	case V2RemediationLevel:
		switch c {
		case 'O':
			return "OF"
		case 'T':
			return "TF"
		case 'N':
			return "ND"
		}
```
To allow proper roundtripping of vectors such as `AV:N/AC:H/Au:N/C:N/I:N/A:C/E:F/RL:OF`?

hdonnay: Hm, I think so.

crozzy (approved):

PR #1144, opened 2023-11-09
Fix multiple CVSS implementation issues, prepare CVSS v4.0 support in the OSV schema

pandatix:
Hi, FIRST.ORG CVSS SIG member and maintainer of [github.com/pandatix/go-cvss](https://github.com/pandatix/go-cvss) here :wave:
While working on the Golang ecosystem and its adoption of CVSS and our recent v4.0, I stumbled upon quay's _unexpected_ CVSS implementation.

## What motivates this PR

Through rapid review, I discovered it is not compliant with the specifications in multiple ways:
- CVSS v2.0 metric ordering is not respected, invalid vectors goes through without error
- mandatory metrics must be part of the vector, at least once at most once
- non-mandatory metrics could be part of the vector, at most once
- obvious **invalid CVSS v2.0 score computation**
- ... probably some others, the previous are already enough to motivate a refactor

For traceability reasons I could open an issue with those information if wanted.

Additionally, CVSS v4.0 has been released, the OSV schema will be updated so we need to be prepared ! :wink:

## Under the hood

Under the hood, this PR provides multiple things:
- compliance with the CVSS specifications
- reduce technical dept by defering maintenance to external Go module
- add support of CVSS v4.0 to prepare the OSV schema update and its support
- provide all this in only one alloc :tada:

The compiled code under `github.com/quay/claircore/updater/osv/cvss.go` is covered at **88.5%**.

### Comment in code

To provide an answer to the comment in code "As far as I can tell, the equations for calculating v3.0 and v3.1 "base" scores are the same.", it differs.

From the Base metrics perspective only, great observation because it did not change: only the Environmental metrics equations changed.
As long as the OSV only provides the security advisories (i.e. Base group metrics, not the Environmental group metrics) it does not affect your computation.
But in the general case, the Environmental metrics concerned will overload their Base equivalent, so pre-computation becomes a bit tricky.

**TL.DR.:** in the OSV case you are handling, you were right !

### Invalid test case

While migrating the test cases to this new code, I discovered a CVSS v2.0 test case that computed invalid scores and transitively invalid severities (big problem for prioritization).
To cite it, the vector `AV:N/AC:L/Au:N/C:C/I:C/A:C` gave `High` but should be `Critical` (score = 10).

This further motivates this PR and the fixes it provides.

### Changes to `go.mod`

This PR introduces a Go version update, which could be blocking for a while before being accepted.
This is due to github.com/pandatix/go-cvss requiring go 1.21.

As long as it is not necessary due to go 1.18 sticking for ClusterFuzz, we may find a solution to this thanks to forward compatibility.


hdonnay:
Thanks for sending this, but I was planning to use the implementations I built in #1143.

When I get a moment, I'll cross reference your comments over there.

pandatix:
Oh nice, did not saw that coming hours ago when I started working on it.

I'll add it to the benchmark/differential fuzzing corpus then !

I let you close this PR if you don't find it relevant as duplicated :wink:

hdonnay:
Closed in favor of #1144.

PR #1145, opened 2023-11-10
gobin: take into account package replacements

hdonnay:


gobin/exe.go
```diff
@@ -132,6 +132,9 @@ func toPackages(ctx context.Context, out *[]*claircore.Package, p string, r io.R
 		vs[bi.Main.Path] = bi.Main.Version
 	}
 	for _, d := range bi.Deps {
+		if r := d.Replace; r != nil {
```
crozzy, line 135: Is it possible the tree is bigger than 2 levels deep? Like can you replace the right-hand side of a replace with another replace? (I know this a little like a play stupid games win stupid prizes issue)

```
replace (
    golang.org/x/net v1.2.3 => example.com/fork/net v1.4.5
    example.com/fork/net v1.4.5 => example.com/another_fork v1.4.6
)
```

crozzy (changes requested):
One query, also, you think it merits a version increment?

gobin/exe.go
hdonnay: I'll look into that. I assume the whole chain is resolved, but don't know that

dominikholler:
@hdonnay thanks for the PR!

crozzy (approved):
LGTM, thanks for checking

PR #1147, opened 2023-11-17
integration: make `PGVERSION` a pattern

hdonnay:
#1140 got me thinking about how the database harness works for not-Linux, so I took a look.
Turns out the project that builds the binaries has added more architectures in the meantime,
so I reworked things to not pin versions as aggressively and support the current suite of OS + architecture.

This change makes it so that specifying a pattern like `latest` or `15` looks at the bill of materials for the embedded PostgreSQL binaries and fetches a matching version. Previously, this environment variable needed to be a full `<major>.<minor>.<patch>` version. That's still allowed if a specific version is needed.


hdonnay:
TODO:
- [ ] Use the BOM "lastUpdated" instead of a week.

crozzy (dismissed):

hdonnay:
Needed to make changes after doing another bout of testing from a clean cache, so re-requesting review.

crozzy (approved):
LGTM

PR #1148, opened 2023-11-17
tarfs: check a potential interger overflow

hdonnay:
If the value is being loaded this way, it should only be up to 11 digits long and thus impossible to overflow an int64, but doing this way "proves" that.

### DIFF
--- a/pkg/tarfs/parse.go
+++ b/pkg/tarfs/parse.go
@@ -185,7 +185,7 @@ func parseNumber(b []byte) (int64, error) {
 	if len(b) == 0 {
 		return 0, nil
 	}
-	n, err := strconv.ParseUint(cstring(b), 8, 64)
+	n, err := strconv.ParseUint(cstring(b), 8, 63) // Only positive int64 values allowed.
 	if err != nil {
 		return 0, err
 	}


pkg/tarfs/parse.go
```diff
@@ -185,7 +185,7 @@ func parseNumber(b []byte) (int64, error) {
 	if len(b) == 0 {
 		return 0, nil
 	}
-	n, err := strconv.ParseUint(cstring(b), 8, 64)
+	n, err := strconv.ParseUint(cstring(b), 8, 63) // Only positive int64 values allowed.
```
crozzy, line 188: What is the point in Parsing into a `uint64` and then returning a `int64` here? A quick way to check for a negative number?

crozzy (approved):

pkg/tarfs/parse.go
hdonnay: The `strconv.ParseInt` function allows for a `+` or `-`, which `strconv.ParseUint` doesn't. The return type being an `int64` is dictated by other parts of the API, it just happens that they cause friction here.

PR #1149, opened 2023-11-17
postgres: remove internal timeouts

hdonnay:
This removes all the internal timeouts for this package. The idea was to keep execution time bounded, but it's becoming increasingly clear this is just a source of problems.


crozzy (approved):
LGTM, sorry, might have gone a little stale but I don't think there has been too much churn in the postgres package

RTann (approved):

PR #1158, opened 2023-11-21
java: allow deactivation of Maven API fetching

BradLugo:
## Description

There's logic in the Java package scanner to disable reaching out to the internet, but we currently set a default API endpoint during `Configure` to prevent that condition. These changes introduce a `DisableAPI` flag to allow the deactivation of the fetching from the public Maven API.

### DIFF
--- a/java/packagescanner_test.go
+++ b/java/packagescanner_test.go
@@ -1,7 +1,12 @@
 package java_test
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"github.com/quay/zlog"
+	"net/http"
+	"net/http/httptest"
 	"path"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


BradLugo:
I think I found something interesting when working on the unit tests for this PR; it seems that `TestScan` for the java package scanner tests instances where `Configure` is never called, which means we've always tested the case where the API is disabled anyway. However, maybe this analysis is incorrect (or maybe even irrelevant if the java package scanner's `Configure` is never called)

java/packagescanner_test.go
```diff
@@ -10,6 +15,35 @@ import (
 	"github.com/quay/claircore/test"
 )
 
+// TestDisableAPI runs the java scanner over some layers known to have java
+// packages installed.
+func TestDisableAPI(t *testing.T) {
```
RTann, line 20: what information did we find with this test?

RTann (commented):
looks good to me just one question

java/packagescanner_test.go
BradLugo: Sorry, can you rephrase the question?

(I elaborated a bit here but I'm not sure it answers your question: https://github.com/quay/claircore/pull/1158#issuecomment-1920237990)

java/packagescanner_test.go
```diff
@@ -10,6 +15,35 @@ import (
...
+		t.Error("http request called")
+	}))
+
+	ctx := context.Background()
+	ctx = zlog.Test(ctx, t)
+	scanner := &java.Scanner{}
+	var buf bytes.Buffer
+	cfg := &java.ScannerConfig{
+		DisableAPI:        true,
+		API:               "",
```
crozzy, line 31: Should this be `API: srv.URL`? Otherwise we never trigger the expected test failure (we just trigger the results parity failure).

java/packagescanner_test.go
```diff
@@ -10,6 +15,35 @@ import (
 	"github.com/quay/claircore/test"
 )
 
+// TestDisableAPI runs the java scanner over some layers known to have java
+// packages installed.
+func TestDisableAPI(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		t.Error("http request called")
```
crozzy, line 22: nit: maybe be kind to a future operator by explaining why this is an error

crozzy: Also, this should be `t.Fatal()` if the API is ever called, we don't care about results parity after this error.

crozzy (changes requested):
Just a couple of comments on the test, the changes look fine to me

java/packagescanner_test.go
BradLugo: 🤦 Thank you

BradLugo:
Thanks for the feedback @crozzy!

java/packagescanner_test.go
BradLugo: re: explaining the error, lmk if the new changes are still unclear. Definitely want to value clearer messages in the tests

crozzy (approved):
LGTM, thanks @BradLugo

RTann (approved):

PR #1163, opened 2023-11-28
rhel(oval): support unfixed OpenShift 4 vulnerabilities

RTann:
The OVAL v2 feeds specifies the next, unreleased version of OpenShift 4 for unfixed OpenShift 4 vulnerabilities. It is meant to indicate each version, y, of OpenShift 4, 4.0 <= y <= 4.x, is affected, where x is the next, unreleased version of OpenShift 4.

ClairCore expects the CPEs to only specify the exact version(s) which is/are affected, as it performs exact matches on the CPEs when matching vulnerabilities. This PR creates clones of the vulnerability for each OpenShift 4 CPE, 4.0 <= y <= 4.x, which is affected by the vulnerability so ClairCore may continue to perform its exact matching on CPEs.

### DIFF
--- a/pkg/ovalutil/rpm.go
+++ b/pkg/ovalutil/rpm.go
@@ -23,12 +23,10 @@ const (
 	NoneDefinition       DefinitionType = "none"
 )
 
-var moduleCommentRegex, definitionTypeRegex *regexp.Regexp
-
-func init() {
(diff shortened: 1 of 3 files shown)


RTann:
This will be closed in favor of supporting this with VEX files in the future

PR #1164, opened 2023-11-28
cicd: add `/fast-forward` command

hdonnay:
This should allow us to merge correctly from the web UI.


hdonnay:
I don't think we can prevent authorized users from pressing the button, but we can at least make it possible.

crozzy (approved):
LGTM

PR #1165, opened 2023-11-29
rhel: add csaf/vex updater

crozzy:
TODO

- [x] Use specific product CVSS scores and normalized them when https://github.com/quay/claircore/pull/1143 is merged
- [x] Delta updating needs to be implemented before this will be mergable
- [x] Implement ingestion of compressed VEX data (once available)
- [x] Finalize fetcher logic
- [x] Write fetcher tests once logic is finalized
- [x] Extend parser tests
- [x] Revisit repo matching logic
- [x] Add CSAF-VEX updater to defaults
- [x] Remove rhel updater from defaults
- [x] Delete rhel vulnerabilities


rhel/vex/fetcher.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+		if err != nil {
+			return nil, hint, err
+		}
+		if res.StatusCode != http.StatusOK {
+			return nil, hint, fmt.Errorf("unexpected response from latest compressed file: %s", res.Status)
+		}
+
+		lm := res.Header.Get("last-modified")
+		fp.requestTime, err = time.Parse(time.RFC1123, lm)
```
hdonnay, line 118:
```suggestion
		fp.requestTime, err = time.Parse(http.TimeFormat, lm)
```

rhel/vex/parser.go
```diff
@@ -0,0 +1,394 @@
...
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/pkg/cpe"
+	"github.com/quay/claircore/toolkit/types/csaf"
+	"github.com/quay/claircore/toolkit/types/cvss"
+)
+
+// Parse implements [driver.Updater].
+func (u *VEXUpdater) Parse(ctx context.Context, contents io.ReadCloser) ([]*claircore.Vulnerability, error) {
+	// NOOP
+	return nil, nil
```
hdonnay, line 23:
```suggestion
	return nil, errors.ErrUnsupported
```

Returning `(nil, nil)` is a valid result, is that a desired signal?

crozzy: I think better to return an error, if someone calls this they are probably expecting it to do something.

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,279 @@
...
+			}
+			if year < lookBackToYear {
+				continue
+			}
+			buf.Grow(int(h.Size))
+			if _, err := buf.ReadFrom(r); err != nil {
+				return nil, hint, err
+			}
+
+			bc := &bytes.Buffer{}
```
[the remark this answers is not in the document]
crozzy: Thanks, also a couple more to correct I noticed

crozzy:
I figure this is probably ready for a review even if it can't be fully merged yet

rhel/matcher.go
```diff
@@ -28,12 +29,32 @@ func (*Matcher) Filter(record *claircore.IndexRecord) bool {
 func (*Matcher) Query() []driver.MatchConstraint {
 	return []driver.MatchConstraint{
 		driver.PackageModule,
-		driver.RepositoryName,
```
RTann, line 31: Do you think we will suddenly match against a lot more rows than previously?

updater/defaults/defaults.go
```diff
@@ -63,6 +64,7 @@ func inner(ctx context.Context) error {
 	updater.Register("debian", df)
 
 	updater.Register("osv", new(osv.Factory))
+	updater.Register("rhel-vex", new(vex.Factory))
```
RTann, line 67: Is the idea to support both OVAL and VEX at the same time for a period, or will this PR remove `"rhel"`, too, later?

crozzy: Probably remove the OVAL in the same PR

rhel/matcher.go
crozzy: The subset we get back from the DB will be bigger, but the repository matching logic is just moving the the `Vulnerable()` method in order to be able to do more complex comparisons than string matching.

updater/defaults/defaults.go
crozzy: @RTann check out the last commit

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+	if err != nil {
+		return nil, hint, err
+	}
+
+	f, err := tmp.NewFile("", "rhel-vex.")
+	if err != nil {
+		return nil, hint, err
+	}
+
+	cw := snappy.NewBufferedWriter(f)
```
RTann, line 48: just curious what made you choose snappy?

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+			if err != nil {
+				return nil, hint, fmt.Errorf("error compressing JSON %s: %w", h.Name, err)
+			}
+			bc.WriteByte('\n')
+			cw.Write(bc.Bytes())
+			buf.Reset()
+			bc.Reset()
+			entriesWritten++
+		}
+		if err != io.EOF {
```
RTann, line 166: similar to another question I had: I've seen a lot of `errors.Is(err, io.EOF)` in the codebase

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+
+		zlog.Debug(ctx).
+			Str("updater", u.Name()).
+			Int("entries written", entriesWritten).
+			Msg("finished writing compressed data to spool")
+
+	}
+
+	uri, err := u.url.Parse("changes.csv")
```
RTann, line 177: since the other files are global, should we do the same here?

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+	if res != nil {
+		defer res.Body.Close()
+	}
+	if err != nil {
+		return nil, hint, err
+	}
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		if t := fp.changesEtag; t == "" || t != res.Header.Get("etag") {
```
RTann, line 198: does this just protect against the case the server incorrectly sends `http.StatusOK` when it should be sending `http.StatusNotModified`?

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+	default:
+		return nil, hint, fmt.Errorf("unexpected response from changes.csv: %s", res.Status)
+	}
+	fp.changesEtag = res.Header.Get("etag")
+
+	rd := csv.NewReader(res.Body)
+	rd.FieldsPerRecord = 2
+	rd.ReuseRecord = true
+	var (
+		l       = 0
```
RTann, line 215: curious why not just `int`?

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+		buf, bc bytes.Buffer
+	)
+	rec, err := rd.Read()
+	for ; err == nil; rec, err = rd.Read() {
+		if len(rec) != 2 {
+			return nil, hint, fmt.Errorf("could not parse changes.csv file")
+		}
+		var err error
+		e := Entry{}
+		e.Path += rec[0] // This += should result in us getting a copy.
```
RTann, line 225: why do we want a copy?

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+			req, err := http.NewRequestWithContext(ctx, http.MethodGet, advisoryURI.String(), nil)
+			if err != nil {
+				return nil, hint, fmt.Errorf("error creating advisory request %w", err)
+			}
+
+			res, err := u.client.Do(req)
+			if err != nil {
+				return nil, hint, fmt.Errorf("error making advisory request %w", err)
+			}
+			if res.StatusCode != http.StatusOK {
```
RTann, line 252: missing a `Body.Close()` when we reach this

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+			err = json.Compact(&bc, buf.Bytes())
+			if err != nil {
+				return nil, hint, fmt.Errorf("error compressing JSON: %w", err)
+			}
+			bc.WriteByte('\n')
+			cw.Write(bc.Bytes())
+			buf.Reset()
+			bc.Reset()
+			l++
+			res.Body.Close()
```
RTann, line 276: another option is to put this into a separate func so we can `defer` without worrying about the loop

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+				var b strings.Builder
+				if _, err := io.Copy(&b, res.Body); err != nil {
+					zlog.Warn(ctx).Err(err).Msg("additional error while reading error response")
+				} else {
+					zlog.Warn(ctx).Str("response", b.String()).Msg("received error response")
+				}
+				return nil, hint, fmt.Errorf("unexpected response from advisary URL: %s %s", res.Status, req.URL)
+			}
+
+			_, err = buf.ReadFrom(res.Body)
```
RTann, line 262: so does this cover the case where this is the first time we read the VEX data and there is a newer version of a few files? This will overwrite data for those files?

rhel/vex/parser.go
```diff
@@ -0,0 +1,410 @@
...
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/toolkit/types/cpe"
+	"github.com/quay/claircore/toolkit/types/csaf"
+	"github.com/quay/claircore/toolkit/types/cvss"
+)
+
+// Parse implements [driver.Updater].
+func (u *VEXUpdater) Parse(ctx context.Context, contents io.ReadCloser) ([]*claircore.Vulnerability, error) {
```
RTann, line 23: For StackRox's usage of Claircore at the moment, we pretty much start fresh every time. Does not implementing `Parse` mean our use case may not work? Related: will the jsonblob stuff work OOTB?

rhel/vex/parser.go
RTann, line 1: Will need to look at this another time

rhel/matcher.go
```diff
@@ -25,15 +28,37 @@ func (*Matcher) Filter(record *claircore.IndexRecord) bool {
...
+		return false, nil
+	}
+	var err error
+	// This conversion has to be done because our current datastructure doesn't
+	// support the claircore.Vulnerability.Repo.CPE field.
+	vuln.Repo.CPE, err = cpe.Unbind(vuln.Repo.Name)
+	if err != nil {
+		return false, nil
+	}
+	if !cpe.Compare(record.Repository.CPE, vuln.Repo.CPE).IsSubset() {
```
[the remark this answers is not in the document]
hdonnay: All CPEs are unbound as 2.3 CPEs. There are tests for matching across URI and formatted-string bindings: https://github.com/quay/claircore/commit/9bb03a13ffebdb5b885f4a84198a0dc76e5dd7c9

rhel/matcher_test.go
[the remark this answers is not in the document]
hdonnay: If the provided pattern doesn't match what it's "meant" to (and there's not a bug in the match implementation), it's a bug in the data.

rhel/vex/updater.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return err
+	}
+	u.url = url
+	zlog.Info(ctx).
+		Str("updater", u.Name()).
+		Msg("configured url")
+
+	u.client = c
+	u.client.Timeout = 5 * time.Second
```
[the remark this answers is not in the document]
hdonnay: This is a bug, you shouldn't be modifying the client. Also, this timeout is almost certainly not what you want, as it's the timeout for _all_ use of a request & response.

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+		uri, err := u.url.Parse(compressedFilename)
+		if err != nil {
+			return nil, hint, err
+		}
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, uri.String(), nil)
+		if err != nil {
+			return nil, hint, err
+		}
+
+		u.client.Timeout = time.Minute * 2 // This is a large request.
```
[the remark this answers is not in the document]
hdonnay: Same bug as the other client -- this should be assumed shared and not modified.

internal/matcher/controller.go
```diff
@@ -142,7 +142,7 @@ func (mc *Controller) filter(ctx context.Context, interested []*claircore.IndexR
 		if err != nil {
 			return nil, err
 		}
-		filtered[record.Package.ID] = match
+		filtered[record.Package.ID] = append(filtered[record.Package.ID], match...)
```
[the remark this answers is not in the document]
crozzy: For context:
* We get the relevant indexing records by leveraging the matcher's `Filter()` method
* We pass those records to the query function to get back a map of potential vulns (`map[record.Package.ID][]vuln`).
* When checking if `Vulnerable()` we iterate through the records and check if it (the record) is vulnerable to any of it's potential vulns.
* The matched vulns are save in the `filtered` map we're referring to
* Index records are create as one per package-repository combination

So we have a situation where previously the `record.Package.ID` was specific enough because there was only one `record` with that `record.Package.ID`  *because* we were Querying on the `RepoName`. Now there are mutiple `records` with the same `record.Package.ID` but different `record.Repo`s. Hence, the last `record` with a random `Repo` would overwrite any potential legitimate vulnerabilities if the matches weren't appended

rhel/matcher.go
```diff
@@ -25,15 +28,37 @@ func (*Matcher) Filter(record *claircore.IndexRecord) bool {
 }
 
 // Query implements driver.Matcher.
-func (*Matcher) Query() []driver.MatchConstraint {
-	return []driver.MatchConstraint{
-		driver.PackageModule,
-		driver.RepositoryName,
+func (m *Matcher) Query() []driver.MatchConstraint {
+	mcs := []driver.MatchConstraint{driver.PackageModule}
```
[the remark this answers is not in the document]
crozzy: We are not deriving any distribution info from the VEX data. The OVAL updater didn't derive distribution from the data either, it parses the name of the file, and this info isn't used for matching.

rhel/matcher.go
```diff
@@ -25,15 +28,37 @@ func (*Matcher) Filter(record *claircore.IndexRecord) bool {
 }
 
 // Query implements driver.Matcher.
-func (*Matcher) Query() []driver.MatchConstraint {
-	return []driver.MatchConstraint{
-		driver.PackageModule,
-		driver.RepositoryName,
```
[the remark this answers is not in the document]
crozzy: As a result of SECDATA-496 (not linking as it's internal) it was decided that unpatched advisories would not be related to all potential repos but they would be related to a subset (using CPE's URI Binding matching notation). This means that comparing the repo is more involved than `==` is it was previously. Hence, we need to move the comparison logic out of the query layer and expand it in the application code.

rhel/matcherfactory.go
```diff
@@ -0,0 +1,35 @@
+package rhel
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+type MatcherFactory struct {
```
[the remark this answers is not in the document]
crozzy: I don't believe that individual Matchers have the ability to be configured, the factory is a bit of middleman between the Clair config and the Matcher (correct me if you see something to the contrary):
```yaml
matchers:
  config:
    rhel:
      ignore_unpatched: true
```

rhel/vex/fetcher.go
crozzy: It satisfied the need for:
> very high speeds and reasonable compression
i.e. the actual compression doesn't save us much apart from the spool being smaller and it being structured JSON there is most likely a lot of low hanging fruit anyway for compression so spending extra effort to compress it the smallest possible didn't seem worth the trade-off.

datastore/postgres/migrations/matcher/13-delete-rhel-oval.sql
```diff
@@ -0,0 +1,4 @@
+-- The rhel-vex updater will now be responsible for RHEL advisories so we have
+-- to delete the existing rhel vulnerabilities.
+DELETE FROM update_operation WHERE updater ~ 'RHEL[5-9]-*';
+DELETE FROM vuln v2 where v2.updater ~ 'RHEL[5-9]-*';
```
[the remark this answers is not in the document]
crozzy: A hangover from a previous query, will remove

datastore/postgres/querybuilder.go
```diff
@@ -70,6 +70,8 @@ func buildGetQuery(record *claircore.IndexRecord, opts *datastore.GetOpts) (stri
 			ex = goqu.Ex{"dist_arch": record.Distribution.Arch}
 		case driver.RepositoryName:
 			ex = goqu.Ex{"repo_name": record.Repository.Name}
+		case driver.HasFixedInVersion:
+			ex = goqu.Ex{"fixed_in_version": goqu.Op{"neq": ""}}
```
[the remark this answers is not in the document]
crozzy: :+1:

rhel/matcher.go
```diff
@@ -25,15 +28,37 @@ func (*Matcher) Filter(record *claircore.IndexRecord) bool {
...
+	if vuln.Repo == nil || record.Repository == nil || vuln.Repo.Key != repositoryKey {
+		return false, nil
+	}
+	var err error
+	// This conversion has to be done because our current datastructure doesn't
+	// support the claircore.Vulnerability.Repo.CPE field.
+	vuln.Repo.CPE, err = cpe.Unbind(vuln.Repo.Name)
+	if err != nil {
+		return false, nil
+	}
```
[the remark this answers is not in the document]
crozzy: Probably yes, we're already filtering by `record.Repository.Key == repositoryKey` so we should be expecting something that can be unbound

rhel/matcher_test.go
```diff
@@ -118,43 +119,97 @@ func TestVulnerable(t *testing.T) {
...
 		{ir: record, v: fixedVulnFuture, want: true, name: "outdated package"},
 		{ir: record, v: unfixedVuln, want: true, name: "unfixed vuln"},
+		{ir: record, v: unfixedVulnBadCPE, want: false, name: "unfixed vuln, invalid CPE"},
+		{ir: record, v: unfixedVulnRepoIsSubset, want: true, name: "unfixed vuln, Repo is a subset"},
+		{ir: record, v: unfixedVulnRepoNotSubset, want: false, name: "unfixed vuln, Repo not a subset"},
 	}
 
 	m := &Matcher{}
-
+	ctx := context.TODO()
```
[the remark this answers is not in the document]
crozzy: Meh probably background

rhel/matcherfactory.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+type MatcherFactoryConfig struct {
+	IgnoreUnpatched bool `json:"ignore_unpatched" yaml:"ignore_unpatched"`
+}
+
+// MatcherFactory implements driver.MatcherConfigurable.
+func (f *MatcherFactory) Configure(ctx context.Context, cfg driver.MatcherConfigUnmarshaler, c *http.Client) error {
+	var fc MatcherFactoryConfig
+	// TODO: add log
```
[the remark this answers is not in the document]
crozzy: No, here. Will update

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+// Entry is an entry in a changes.csv file.
+type Entry struct {
+	// This path should be parsed in the context of the manifest's URL.
+	Path        string
+	updatedTime time.Time
+}
+
+func (u *VEXUpdater) Fetch(ctx context.Context, hint driver.Fingerprint) (io.ReadCloser, driver.Fingerprint, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "rhel/VEXUpdater.Fetch")
+	fp, err := ParseFingerprint(string(hint))
```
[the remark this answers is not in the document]
crozzy: Yeah, probably nicer to accept a `driver.Fingerprint()`, :+1:

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+		if err != nil {
+			return nil, hint, err
+		}
+
+		compressedFilename := string(body)
+		zlog.Debug(ctx).
+			Str("filename", compressedFilename).
+			Msg("requesting latest compressed file")
+
+		// Decided against parsing the filename to allow the implementation
```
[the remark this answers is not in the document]
crozzy: I think I added this as the compressed filename is when it was generated but I think it's kind of moot since prodsec added the `archive_latest.txt` endpoint, will remove comment.

rhel/vex/fetcher.go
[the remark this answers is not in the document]
crozzy: :+1:

rhel/vex/updater.go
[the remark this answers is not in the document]
crozzy: :+1:

updater/defaults/defaults.go
crozzy: yep, arbitrary afaik

rhel/vex/updater.go
```diff
@@ -0,0 +1,151 @@
...
+var (
+	_ driver.Updater           = (*VEXUpdater)(nil)
+	_ driver.Configurable      = (*VEXUpdater)(nil)
+	_ driver.DeltaUpdater      = (*VEXUpdater)(nil)
+	_ driver.UpdaterSetFactory = (*Factory)(nil)
+	_ driver.Configurable      = (*Factory)(nil)
+)
+
+const (
+	baseURL        = "https://access.redhat.com/security/data/csaf/beta/vex/"
```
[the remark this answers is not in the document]
crozzy: This is one of the reasons the PR is still in draft but can also add a TODO

rhel/vex/testdata/example_vex.jsonl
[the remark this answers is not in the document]
crozzy: yes, json newline delimited

rhel/vex/fetcher.go
crozzy: At the moment we'll spool the compressed file to disk then append the updated files, the parser takes care of only persisting the latest version.

rhel/matcherfactory.go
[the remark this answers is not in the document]
crozzy: No, now. Adding

datastore/postgres/querybuilder.go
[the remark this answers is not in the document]
crozzy: Can do

rhel/vex/fetcher.go
crozzy: package variable, yes probably

rhel/vex/fetcher.go
crozzy: Can you expand of this a little?

rhel/vex/fetcher.go
crozzy: for me it's marginally clearer this way, can change, what is the argument for the other way?

rhel/vex/testdata/1.json
crozzy, line 1: Need to delete

rhel/vex/updater.go
```diff
@@ -0,0 +1,151 @@
...
+	latestFile     = "archive_latest.txt"
+	lookBackToYear = 2005
```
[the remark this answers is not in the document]
crozzy: Honestly it was kind of arbitrary, I figure it well beyond what other sources offer and there is toss up between knowing as much as possible and having data that is never queried hanging around in the DB, 20 years seemed like a cautious balance (although the exponential pattern of vulns / year is such that I'm not sure how much space we actually save).

rhel/vex/updater.go
```diff
@@ -0,0 +1,151 @@
...
+
+func NewFingerprint(etag string, requestTime time.Time) *fingerprint {
+	return &fingerprint{
+		changesEtag: etag,
+		requestTime: requestTime,
+	}
+}
+
+func ParseFingerprint(fp string) (*fingerprint, error) {
+	// On the fence with this or adding a specialist error
```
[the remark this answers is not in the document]
crozzy: usually the fingerprint simple string and doesn't need to be interpreted. I decided not to

rhel/vex/updater.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Configure implements driver.Configurable.
+func (u *VEXUpdater) Configure(ctx context.Context, f driver.ConfigUnmarshaler, c *http.Client) error {
+	ctx = zlog.ContextWithValues(ctx, "component", "rhel/VEXUpdater.Configure")
+
+	var cfg UpdaterConfig
+	if err := f(&cfg); err != nil {
+		return err
+	}
+	bu := baseURL
```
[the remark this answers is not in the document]
crozzy: It just means the base URL is configurable at both levels, I think this is maybe overkill. For Clair's config I think it makes sense at the Factory level and maybe the `updater.Configure` goes away. How does it work from an ACS POV? I guess you instantiate the factories too.

rhel/vex/fetcher.go
crozzy: I've refactored this out, I don't think creating an `Entry` is necessary

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,288 @@
...
+			defer latestRes.Body.Close()
+		}
+		if err != nil {
+			return nil, hint, err
+		}
+		if latestRes.StatusCode != http.StatusOK {
+			return nil, hint, fmt.Errorf("unexpected response from archive_latest.txt: %s", latestRes.Status)
+		}
+
+		body, err := io.ReadAll(latestRes.Body) // Fine to use as expecting small number of bytes.
```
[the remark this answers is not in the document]
crozzy: I'm not sure it makes sense considering we're downloading over HTTPS and presumably if someone somehow had access to the server they could change the checksums anyway. The signature could be more useful but probably more trouble than it's worth to set up PGP verification in go as well as suffering the additional transfer. What are your thoughts?

rhel/vex/parser.go
crozzy: When starting from fresh, `DeltaParse` will consider the whole RH VEX corpus, so it should be fine there. I think for the jsonblob store to work the `DeltaUpdateVulnerabilities` would need to be implemented.

crozzy: https://github.com/quay/claircore/pull/1345 I added this do basically make sure any delta updaters will persist all their data.

rhel/matcher.go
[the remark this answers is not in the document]
RTann: This is used for the `WHERE` clause, right? I think I'm still confused why we don't also filter out anything under a different OS

internal/matcher/controller.go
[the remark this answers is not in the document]
RTann: ah ok, so this is called prior to `Vulnerable()` and we need all potential CPEs, as `Vulnerable()` will do the CPE matching from now on. If this is correct, feel free to resolve

rhel/vex/updater.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return nil
+}
+
+type VEXUpdater struct {
+	url    *url.URL
+	client *http.Client
+}
+
+// fingerprint is used to track the state of the changes.csv endpoint.
```
RTann, line 85: and deletions.csv

rhel/vex/updater_test.go
```diff
@@ -0,0 +1,49 @@
...
+			err:  false,
+		},
+		{
+			name: "date error",
+			val:  `one\two\2006-01-02T15:04:05ZMore`,
+			err:  true,
+		},
+		{
+			name: "etag error",
+			val:  `one\tw\o\2006-01-02T15:04:05ZMore`,
```
RTann, line 28: can you make the time valid to confirm we are failing due to the extra fields and not the time?

rhel/vex/testdata/example_vex.jsonl
RTann, line 1: what is jsonl?

rhel/vex/parser.go
```diff
@@ -0,0 +1,450 @@
...
+	for _, pc := range v.ProductStatus["known_affected"] {
+		rel := c.c.ProductTree.Relationships.FindRelationship(pc, "default_component_of")
+		if rel == nil {
+			// Should never get here, error in data
+			zlog.Debug(ctx).
+				Str("product:package", pc).
+				Msg("could not find a relationship for product:package")
+			continue
+		}
+		if strings.HasPrefix(rel.ProductRef, "kernel") {
```
RTann, line 183: Ask Prod Sec about this once, and I see we wrote this note:

> For Scanner Certification purposes, it’s ok to show kernel header vulns, but you can also feel free to ignore them since they aren’t used in images

so LGTM

rhel/vex/parser.go
```diff
@@ -0,0 +1,450 @@
...
+			default:
+				pkgName = purl.Name
+			}
+			if purl.Type != packageurl.TypeRPM || purl.Namespace != "redhat" {
+				// Just ingest advisories that are Red Hat RPMs, this will
+				// probably change down the line when we consolidate updaters.
+				continue
+			}
+		}
+		if strings.HasPrefix(pkgName, "kernel") {
```
RTann, line 234: why check again?

rhel/vex/parser.go
```diff
@@ -0,0 +1,450 @@
...
+			Kind: claircore.SOURCE,
+		}
+
+		wfn, err := cpe.Unbind(cpeHelper)
+		if err != nil {
+			return nil, fmt.Errorf("could not unbind cpe: %s %w", cpeHelper, err)
+		}
+		vuln.Repo = c.rc.Get(wfn)
+		if sc := c.c.FindScore(pc); sc != nil {
+			vuln.NormalizedSeverity, vuln.Severity, err = CVSSVectorFromScore(sc)
```
RTann, line 254: This should be related to the "threats" field for the CPE instead of the CVSS score

rhel/vex/parser.go
```diff
@@ -0,0 +1,450 @@
...
+		return
+	default:
+		err = errors.New("could not find a valid CVSS object")
+	}
+	return
+}
+
+// CVSS2ClaircoreSeverity returns a claircore.Severity given a cvss.Qualitative
+// It will panic if asked to reason with an unknown Qualitative score.
+func CVSS2ClaircoreSeverity(q cvss.Qualitative) claircore.Severity {
```
RTann, line 419: why the 2? I thought this was specifically for CVSS 2.0 at first

rhel/vex/parser.go
RTann, line 1: I looked through this a bit, but this PR is rather massive, so I'll have to look in more detail next round

datastore/postgres/migrations/matcher/13-delete-rhel-oval.sql
```diff
@@ -0,0 +1,4 @@
+-- The rhel-vex updater will now be responsible for RHEL advisories so we have
+-- to delete the existing rhel vulnerabilities.
+DELETE FROM update_operation WHERE updater ~ 'RHEL[5-9]-*';
```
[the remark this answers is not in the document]
crozzy: I was keying off of https://access.redhat.com/security/data/oval/v2/PULP_MANIFEST so I _think_ it should be ok

datastore/postgres/migrations/matcher/13-delete-rhel-oval.sql
```diff
@@ -0,0 +1,4 @@
...
+DELETE FROM vuln where updater ~ 'RHEL[5-9]-*';
```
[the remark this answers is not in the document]
crozzy: They should be doing the same thing. The osv deletions is derived from the garbage collection logic there the update_operation is initially deleted (which deletes the uo_vuln entry in the CASCADE), then deletes any vulns that don't have a uo_vuln entry (and are from the the `osv` updater).

This migration will delete the update_operation given the updater, then the vulns given the same updater.

datastore/postgres/querybuilder.go
```diff
@@ -70,6 +71,8 @@ func buildGetQuery(record *claircore.IndexRecord, opts *datastore.GetOpts) (stri
 			ex = goqu.Ex{"dist_arch": record.Distribution.Arch}
 		case driver.RepositoryName:
 			ex = goqu.Ex{"repo_name": record.Repository.Name}
+		case driver.HasFixedInVersion:
+			ex = goqu.Ex{"fixed_in_version": goqu.Op{exp.NeqOp.String(): ""}}
```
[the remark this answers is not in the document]
crozzy: It's currently always `""` when there is no fixed in version (due to the fact that vulnerability.FixedInVersion is just a string and there is no special logic in the insertion that will convert `""`s to NULL/nil pre-insertion).

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,388 @@
...
+
+		uri, err := u.url.Parse(compressedFilename)
+		if err != nil {
+			return nil, hint, err
+		}
+
+		rctx, cancel := context.WithTimeout(ctx, compressedFileTimeout)
+		defer cancel()
+
+		req, err := http.NewRequestWithContext(rctx, http.MethodGet, uri.String(), nil)
```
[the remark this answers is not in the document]
crozzy: Right, we also had that information from prodsec, I think we're ok to not use them

rhel/vex/fetcher.go
crozzy: Right, if it StatusOK but nothing actually changed, fallthrough to the `case http.StatusNotModified`

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,388 @@
...
+
+		cvePath, uTime := rec[0], rec[1]
+		year, err := strconv.ParseInt(path.Dir(cvePath), 10, 64)
+		if err != nil {
+			return fmt.Errorf("error parsing year %w", err)
+		}
+		if year < lookBackToYear {
+			continue
+		}
+		updatedTime, err := time.Parse(time.RFC3339, uTime)
```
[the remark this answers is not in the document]
crozzy: Yes, `time.RFC3339` should encompass Zulu and TZ offsetted strings

rhel/matcher.go
```diff
@@ -2,15 +2,21 @@ package rhel
 
 import (
 	"context"
+	"strings"
 
 	version "github.com/knqyf263/go-rpm-version"
 
+	"github.com/quay/zlog"
```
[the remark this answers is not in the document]
hdonnay: Yeah, this should be in normal form.

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,388 @@
...
+					}
+					zlog.Debug(ctx).Str("url", advisoryURI.String()).Msg("copying body to file")
+					err = json.Compact(&bc, buf.Bytes())
+					if err != nil {
+						return fmt.Errorf("error compressing JSON: %w", err)
+					}
+				default:
+					// Something went wrong, try and read some body and return an error.
+					var b strings.Builder
+					if _, err := io.Copy(&b, res.Body); err != nil {
```
hdonnay, line 282: limit how much of the body is read, so that the logs don't end up with entire quoted HTML documents in them.

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,388 @@
...
+					} else {
+						zlog.Warn(ctx).Str("response", b.String()).Msg("received error response")
+					}
+					return fmt.Errorf("unexpected response from advisary URL: %s %s", res.Status, req.URL)
+				}
+
+				bc.WriteByte('\n')
+				wtr.Write(bc.Bytes())
+				buf.Reset()
+				bc.Reset()
```
hdonnay, line 293: Same concerns with this as the `Fetch` use.

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,388 @@
...
+				l++
+				return nil
+			}()
+			if !errors.Is(err, nil) {
+				return err
+			}
+		}
+	}
+	switch err {
+	case io.EOF:
```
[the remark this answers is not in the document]
hdonnay: I rather have the default case empty and the `return` immediately after, but that's just a style peccadillo.

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,388 @@
...
+	default:
+		return fmt.Errorf("error parsing the changes.csv file: %w", err)
+	}
+	return nil
+}
+
+// processDeletions deals with the published deletions.csv, adding records
+// to the wtr is they are deemed to have changed since the compressed
+// file was last processed. wtr and fp can be modified.
+func (u *VEXUpdater) processDeletions(ctx context.Context, wtr io.Writer, fp *fingerprint) error {
```
hdonnay, line 313: Same comment on an idiomatic name

hdonnay (changes requested):
Submitting review on first chunk of files

rhel/matcher_test.go
```diff
@@ -163,3 +240,36 @@ func TestVulnerable(t *testing.T) {
...
+
+	testcases := []struct {
+		name               string
+		recordCPE, vulnCPE cpe.WFN
+		match              bool
+	}{
+		{
+			name:      "simple_case",
+			recordCPE: cpe.MustUnbind("cpe:/a:redhat:openshift:4.13::el8"),
+			vulnCPE:   cpe.MustUnbind("cpe:/a:redhat:openshift:4"),
```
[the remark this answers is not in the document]
crozzy: No, not from what I've observed, it'll either have the minor and the prefix or neither

rhel/vex/updater.go
```diff
@@ -0,0 +1,160 @@
...
+	_ driver.Updater           = (*VEXUpdater)(nil)
+	_ driver.Configurable      = (*VEXUpdater)(nil)
+	_ driver.DeltaUpdater      = (*VEXUpdater)(nil)
+	_ driver.UpdaterSetFactory = (*Factory)(nil)
+	_ driver.Configurable      = (*Factory)(nil)
+)
+
+const (
+	// TODO: VEX data is currently in beta, update URL when VEX data is GA.
+	baseURL        = "https://access.redhat.com/security/data/csaf/beta/vex/"
```
RTann, line 27: Looks like it's available now: https://access.redhat.com/security/data/csaf/v2/vex/

rhel/matcher.go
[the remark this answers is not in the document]
crozzy: It would be used in the where clause if we added a constraint here, but we currently don't have a good way to derive the distro from the VEX data (AFAICT) so we can't filter on it. Currently with OVAL data we derive the distro from the OVAL file's name but it isn't used in the matching.

rhel/vex/parser.go
crozzy: We want to keep the `vuln.Severity` as the CVSS score though right (so we are able to surface it to the user)? Is there some doc that says we should use the `threats` field? (not saying we shouldn't, it'd just be nice to reference something and explain why the `vuln.NormalizedSeverity` and `vuln.Severity` might equate to different normalized values.

rhel/vex/fetcher.go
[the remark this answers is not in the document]
crozzy: @RTann this means nil will satisfy the first case no?

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,388 @@
...
+	if err != nil {
+		return nil, hint, err
+	}
+
+	err = u.processDeletions(ctx, cw, fp)
+	if err != nil {
+		return nil, hint, err
+	}
+
+	fp.requestTime = time.Now()
```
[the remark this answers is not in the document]
crozzy: If the compressed file is processed (i.e. the `changesEtag` is empty) we want to process all the changes that happened since the compressed file was modified (i.e. `fp.requestTime, err = time.Parse(http.TimeFormat, lm`). If the compressed file was not read (i.e. this wasn't the first time the updater was run) We want to process all the changes since the last time we processed the changes i.e. setting `fp.requestTime = time.Now()` so the next run knows where it left off. Let me know if this makes sense.

rhel/vex/parser.go
```diff
@@ -0,0 +1,450 @@
...
+
+	r := bufio.NewReader(snappy.NewReader(contents))
+	for b, err := r.ReadBytes('\n'); err == nil; b, err = r.ReadBytes('\n') {
+		c, err := csaf.Parse(bytes.NewReader(b))
+		if err != nil {
+			return nil, nil, fmt.Errorf("error parsing CSAF: %w", err)
+		}
+		name := c.Document.Tracking.ID
+		if c.Document.Tracking.Status == "deleted" {
+			deleted = append(deleted, name)
```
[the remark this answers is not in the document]
crozzy: Nice catch, we wouldn't actually do any work because the object wouldn't have vulnerabilities, but best to early out

rhel/vex/parser.go
crozzy: This is checking the `pkgName` which could be distinct if it's an `oci` purl (although that probably negates it from being a kernel advisory. I'll remove this check

rhel/vex/updater.go
```diff
@@ -0,0 +1,181 @@
...
+	f.base, err = url.Parse(u)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// VEXUpdater is responsible from reading VEX data served at the URL
+// and creating vulnerabilities.
+type VEXUpdater struct {
```
hdonnay, line 95: `vex.VEXUpdater` stutters. Any particular reason for this name choice?

rhel/vex/updater.go
```diff
@@ -0,0 +1,181 @@
...
+	client *http.Client
+}
+
+// fingerprint is used to track the state of the changes.csv and deletions.csv endpoints.
+//
+// The spec (https://www.rfc-editor.org/rfc/rfc9110.html#name-etag) mentions
+// that there is no need for the client to be aware of how each entity tag
+// is constructed, however, it mentions that servers should avoid backslashes.
+// Hence, the `\` character is used as a separator when stringifying.
+type fingerprint struct {
```
hdonnay, line 106: Export this type if `NewFingerprint` or `ParseFingerprint` are needed outside this package, or unexport those functions.

rhel/vex/parser.go
```diff
@@ -0,0 +1,463 @@
...
+}
+
+// productCache keeps a cache of all seen csaf.Products
+type productCache struct {
+	cache map[string]*csaf.Product
+}
+
+// NewProductCache returns a productCache with the backing
+// map instanciated.
+func NewProductCache() *productCache {
```
hdonnay, line 143: Ditto for being exported.

rhel/vex/parser.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	p := c.ProductTree.FindProductByID(productID)
+	pc.cache[productID] = p
+	return p
+
+}
+
+// NewCreator returns a creator object used for processing parts of a VEX file
+// and returning claircore.Vulnerabilities.
+func NewCreator(vulnName, vulnLink string, c *csaf.CSAF, pc *productCache, rc *repoCache) *creator {
```
hdonnay, line 164: Ditto for being exported.

rhel/vex/parser.go
hdonnay: Ditto, I think this a bad name

rhel/vex/parser.go
```diff
@@ -0,0 +1,463 @@
...
+		}
+	}
+	out := make([]*claircore.Vulnerability, len(c.fixedVulns))
+	for i := range c.fixedVulns {
+		out[i] = &c.fixedVulns[i]
+	}
+	return out, nil
+}
+
+func CVSSVectorFromScore(sc *csaf.Score) (sev claircore.Severity, vec string, err error) {
```
hdonnay, line 392: Exported?

rhel/vex/parser.go
```diff
@@ -0,0 +1,463 @@
...
+		return
+	default:
+		err = errors.New("could not find a valid CVSS object")
+	}
+	return
+}
+
+// CVSS2ClaircoreSeverity returns a claircore.Severity given a cvss.Qualitative
+// It will panic if asked to reason with an unknown Qualitative score.
+func CVSS2ClaircoreSeverity(q cvss.Qualitative) claircore.Severity {
```
hdonnay, line 432: Exported?

rhel/vex/parser.go
```diff
@@ -0,0 +1,463 @@
...
+	s, ok := CVSS2ClaircoreSeverityMap[q]
+	if !ok {
+		panic("unmappable cvss.Qualitative score")
+	}
+	return s
+}
+
+var CVSS2ClaircoreSeverityMap = map[cvss.Qualitative]claircore.Severity{
```
hdonnay, line 440: Exported?

hdonnay (changes requested):
Okay, went over all of it.

Don't have any structural/correctness questions.

rhel/vex/updater.go
crozzy: Just a hangover for it originally starting out-of-tree, changing

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,394 @@
...
+				}
+			default:
+				// Something went wrong, try and read some body and return an error.
+				bodyStart, err := io.ReadAll(io.LimitReader(res.Body, 256))
+				if err != nil {
+					zlog.Warn(ctx).Err(err).Msg("additional error while reading error response")
+				} else {
+					zlog.Warn(ctx).Bytes("response", bodyStart).Msg("received error response")
+				}
+				return fmt.Errorf("unexpected response from advisory URL %q: %s", req.URL, res.Status)
```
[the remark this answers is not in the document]
crozzy: https://github.com/quay/claircore/issues/1366

hdonnay (approved):
LGTM

rhel/vex/parser.go
```diff
@@ -0,0 +1,454 @@
...
+			return nil, fmt.Errorf("could not unbind cpe: %s %w", cpeHelper, err)
+		}
+		vuln.Repo = c.rc.Get(wfn)
+		if sc := c.c.FindScore(pc); sc != nil {
+			vuln.Severity, err = cvssVectorFromScore(sc)
+			if err != nil {
+				return nil, fmt.Errorf("could not parse CVSS score: %w, file: %s", err, c.vulnLink)
+			}
+		}
+		if t := c.c.FindThreat(pc, "impact"); t != nil {
```
crozzy, line 271: @RTann @hdonnay Here we are getting the threat and converting to a `claircore.Severity` as discussed. There are currently a few files that are known to have no `threat` object, we are waiting on RH prodsec for instructions on how to deal with those / if that is a bug. Right now the severity will just be `Unknown` and the logs will warn.

rhel/vex/updater.go
```diff
@@ -99,6 +101,7 @@ func (f *Factory) Configure(ctx context.Context, cf driver.ConfigUnmarshaler, c
 type Updater struct {
 	url             *url.URL
 	client          *http.Client
+	version         string
```
hdonnay: Why is this a runtime value and not a package constant?

crozzy: It doesn't need to be, I just thought it made sense to tie the version to the thing it's describing, but same same, I can just read the const.

hdonnay: I'd rather it be a constant (like the indexers) because it can't vary at runtime.

hdonnay (dismissed):

rhel/vex/parser.go
```diff
@@ -0,0 +1,471 @@
...
+		}
+		ctx = zlog.ContextWithValues(ctx, "link", selfLink)
+		creator := newCreator(name, selfLink, c, pc, rc)
+		for _, v := range c.Vulnerabilities {
+			// Create vuln here, there should always be one vulnerability
+			// here in the case of RH VEX but the spec allows multiple.
+			links := []string{}
+			for _, r := range v.References {
+				links = append(links, r.URL)
+			}
```
RTann, line 66: It'd be nice to ensure "self" is first (this one. Not the one from `c.Document.References`) so implementations who just look at the first link will always get this one. I know StackRox does this, but I'm not sure if Quay does, too

rhel/vex/parser.go
```diff
@@ -0,0 +1,471 @@
...
+					desc = n.Text
+				}
+			}
+
+			protoVuln := func() *claircore.Vulnerability {
+				v := &claircore.Vulnerability{
+					Updater:            u.Name(),
+					Name:               name,
+					Description:        desc,
+					Issued:             v.ReleaseDate,
```
RTann, line 81: just curious, we prefer this over `discovery_date`? Guess this was published on `release_date`

rhel/vex/parser.go
```diff
@@ -0,0 +1,471 @@
...
+		c.uniqueVulnsIdx[vulnKey] = idx
+	}
+	return &c.fixedVulns[idx], ok
+}
+
+// FixedVulnerabilities processes the "fixed" array of products in the
+// VEX object.
+func (c *creator) fixedVulnerabilities(ctx context.Context, v csaf.Vulnerability, protoVulnFunc func() *claircore.Vulnerability) ([]*claircore.Vulnerability, error) {
+	for _, pc := range v.ProductStatus["fixed"] {
+		rel := c.c.FindRelationship(pc, "default_component_of")
```
RTann, line 305: I'm looking at https://access.redhat.com/security/data/csaf/v2/vex/2021/cve-2021-44228.json, and it looks like we will skip "EAP 7.4 log4j async", for example, which has an affected/fixed component, "log4j-core". We should not skip this, right?

crozzy: We need a relationship to discern the repository and the RPM package, what would the RPM package affected be in this case, what is it's purl?

rhel/vex/parser.go
crozzy: yep

rhel/vex/parser.go
crozzy: You think a user would prefer to parse through a big blob of JSON rather than the RH CVE page?

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,390 @@
...
+		}
+		deletedJSON, err := createDeletedJSON(cvePath)
+		if err != nil {
+			zlog.Warn(ctx).Err(err).Msg("error creating JSON object denoting deletion")
+		}
+		bc.Write(deletedJSON)
+		bc.WriteByte('\n')
+		w.Write(bc.Bytes())
+	}
+	switch {
```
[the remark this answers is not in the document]
crozzy: I don't think so, I went down this path originally and the logic became overly complex.

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,390 @@
...
+			return nil
+		}()
+		if !errors.Is(err, nil) {
+			return err
+		}
+
+	}
+
+	switch {
+	case errors.Is(err, nil):
```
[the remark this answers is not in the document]
crozzy: In this case it wouldn't but in the changes function it would as the `csv.Read()` err is only scoped to the for loop in that function. My preference is to revert the change that moves the initial `Read()` to the for loop declaration and then just check if it's EOF or not.

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,390 @@
...
+
+	cw := snappy.NewBufferedWriter(f)
+
+	var success bool
+	defer func() {
+		if err := cw.Close(); err != nil {
+			zlog.Warn(ctx).Err(err).Msg("unable to close snappy writer")
+		}
+		if success {
+			if _, err := f.Seek(0, io.SeekStart); err != nil {
```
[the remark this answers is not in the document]
crozzy: If we weren't successful we won't be reading the file, I think I could probably move this to not be bound by success being true, but I kind of like the fact that the seeking is setup after the file is created, there's no chance something else could be added to the file.

rhel/vex/parser.go
```diff
@@ -0,0 +1,454 @@
...
+func newProductCache() *productCache {
+	return &productCache{
+		cache: make(map[string]*csaf.Product),
+	}
+}
+
+// Get is a wrapper around the FindProductByID method that
+// attempts to return from the cache before traversing the
+// CSAF object.
+func (pc *productCache) Get(c *csaf.CSAF, productID string) *csaf.Product {
```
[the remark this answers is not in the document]
crozzy: sure

rhel/vex/updater.go
```diff
@@ -0,0 +1,185 @@
...
+type Updater struct {
+	url    *url.URL
+	client *http.Client
+}
+
+// fingerprint is used to track the state of the changes.csv and deletions.csv endpoints.
+//
+// The spec (https://www.rfc-editor.org/rfc/rfc9110.html#name-etag) mentions
+// that there is no need for the client to be aware of how each entity tag
+// is constructed, however, it mentions that servers should avoid backslashes.
```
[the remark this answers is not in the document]
crozzy: I added the comment in case it wasn't obvious why backslashes were used, haven't seen then used at any point in the etags coming from RH security data servers.

rhel/matcher.go
```diff
@@ -25,15 +30,55 @@ func (*Matcher) Filter(record *claircore.IndexRecord) bool {
...
+	// support the claircore.Vulnerability.Repo.CPE field.
+	vuln.Repo.CPE, err = cpe.Unbind(vuln.Repo.Name)
+	if err != nil {
+		zlog.Warn(ctx).
+			Str("vulnerability name", vuln.Name).
+			Err(err).
+			Msg("unable to unbind repo CPE")
+		return false, nil
+	}
+	if !cpe.Compare(record.Repository.CPE, vuln.Repo.CPE).IsSubset() && !isCPESubstringMatch(record.Repository.CPE, vuln.Repo.CPE) {
```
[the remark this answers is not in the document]
crozzy: IIUC `cpe:/a:redhat:product:1.0` is not a subset of `cpe:/a:redhat:product:1.1`

rhel/matcher.go
```diff
@@ -25,15 +30,55 @@ func (*Matcher) Filter(record *claircore.IndexRecord) bool {
...
+// doesn't use the syntax defined in the Matching Expression spec.
+// For example, "cpe:/a:redhat:openshift:4" is expected to match "cpe:/a:redhat:openshift:4.13::el8".
+//
+// This is defined (citation needed) to be a substring match on the "pattern" and "target" CPEs.
+// Since we always normalize CPEs into v2.3 "Formatted String" form, we need to trim the
+// added "ANY" attributes from the pattern.
+//
+// TODO(crozzy) Remove once RH VEX data updates CPEs with standard matching expressions.
+func isCPESubstringMatch(recordCPE cpe.WFN, vulnCPE cpe.WFN) bool {
+	return strings.HasPrefix(recordCPE.String(), strings.TrimRight(vulnCPE.String(), ":*"))
```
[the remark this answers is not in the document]
crozzy: The `*` is a literal, when we retrieve the value from the DB it has been formatted (i.e. `cpe:/a:redhat:openshift:4:*:*:*:*:*:*`, this will just remove the `:*`s not any actual information.

rhel/matcherfactory.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	return []driver.Matcher{m}, nil
+}
+
+type MatcherFactoryConfig struct {
+	IgnoreUnpatched bool `json:"ignore_unpatched" yaml:"ignore_unpatched"`
+}
+
+// MatcherFactory implements driver.MatcherConfigurable.
+func (f *MatcherFactory) Configure(ctx context.Context, cfg driver.MatcherConfigUnmarshaler, c *http.Client) error {
```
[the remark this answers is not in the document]
crozzy: We don't need to the client as part of the matching process,  I can replace with an `_` if that'd make it more obvious.

crozzy:
@RTann https://github.com/quay/claircore/pull/1165/commits/1d2ccd3bcb0e45bab7f6b6e7ebe0de1958ea77bc comments are addressed in this fixup or in-line

rhel/matcher.go
```diff
@@ -25,15 +30,55 @@ func (*Matcher) Filter(record *claircore.IndexRecord) bool {
...
 }
 
 // Vulnerable implements driver.Matcher.
-func (m *Matcher) Vulnerable(_ context.Context, record *claircore.IndexRecord, vuln *claircore.Vulnerability) (bool, error) {
+//
+// Vulnerable will interpret the claircore.Vulnerability.Repo.CPE
+// as a CPE match expression, and to be considered vulnerable,
+// the relationship between claircore.IndexRecord.Repository.CPE and
+// the claircore.Vulnerability.Repo.CPE needs to be a CPE Name Comparison
+// Relation of SUBSET(⊂)(source is a subset of, or equal to the target).
```
hdonnay, line 61: Should be ⊇ (or the spec `SUPERSET`) -- "Source is a superset or equal to the target"

rhel/matcher.go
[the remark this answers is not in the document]
hdonnay: Incoming CPEs will always be in 2.3 Formatted String binding, so removing trailing `:*` characters turns it into the populated substring.

rhel/vex/fetcher.go
```diff
@@ -0,0 +1,384 @@
...
+
+		// Use a func here as we're in a loop and want to make sure the
+		// body is closed in all events.
+		err = func() error {
+			res, err := u.client.Do(req)
+			if err != nil {
+				return fmt.Errorf("error making advisory request %w", err)
+			}
+			defer res.Body.Close()
+			err = httputil.CheckResponse(res, http.StatusOK)
```
RTann, line 275: just curious why this is used here but not earlier in the file where similar status checks were used

rhel/vex/parser.go
```diff
@@ -0,0 +1,521 @@
...
+// Parse implements [driver.Updater].
+func (u *Updater) Parse(ctx context.Context, contents io.ReadCloser) ([]*claircore.Vulnerability, error) {
+	// NOOP
+	return nil, errors.ErrUnsupported
+}
+
+// DeltaParse implements [driver.DeltaUpdater].
+func (u *Updater) DeltaParse(ctx context.Context, contents io.ReadCloser) ([]*claircore.Vulnerability, []string, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "rhel/vex/Updater.DeltaParse")
+	// This map is needed for deduplication purposes, the compressed CSAF data maybe include
```
RTann, line 32:
```suggestion
	// This map is needed for deduplication purposes, the compressed CSAF data may include
```

rhel/vex/parser.go
RTann, line 1: @crozzy confirming this file is ready for review? I haven't looked at it this time around, yet

crozzy: Yes, ready for review

rhel/vex/fetcher.go
crozzy: Added for the latest request and the request of the compressed data, didn't add it for the others (changes.csv and deletions.csv) as status informs the logic going forward.

rhel/vex/parser.go
crozzy: Updated

rhel/vex/parser.go
RTann: I'm referring to this "self" under `vulnerabilities.references`:

```
"references": [
        {
          "category": "self",
          "summary": "Canonical URL",
          "url": "https://access.redhat.com/security/cve/CVE-2024-3727"
        },
        {
          "category": "external",
          "summary": "RHBZ#2274767",
          "url": "https://bugzilla.redhat.com/show_bug.cgi?id=2274767"
        },
```

not sure if it's guaranteed to be first, but it's definitely the one users will care about most. Though, that can just be the client's problem to fetch it, if that's the one they want (for example StackRox can just search for the link prefixed with `https://access.redhat.com/security/cve/CVE-`

RTann (approved):

hdonnay (approved):
LGTM

PR #1166, opened 2023-12-01
enrichers: Load enrichments from jsonblob storage

jvdm:
A patch to fix the following BUG:

https://github.com/quay/claircore/blob/faffb8e880263171ca9b54dc2f5609547e53cbb7/libvuln/jsonblob/jsonblob.go#L94-L96

### DIFF
--- a/libvuln/updates.go
+++ b/libvuln/updates.go
@@ -2,8 +2,10 @@ package libvuln
 
 import (
 	"context"
+	"fmt"
 	"io"
 
+	"github.com/google/uuid"
 	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/quay/zlog"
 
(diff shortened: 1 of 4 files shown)


libvuln/updates.go
```diff
@@ -47,9 +49,16 @@
...
+		var ref uuid.UUID
+		if e.Enrichment != nil {
+			if ref, err = s.UpdateEnrichments(ctx, e.Updater, e.Fingerprint, e.Enrichment); err != nil {
+				return fmt.Errorf("updating enrichements: %w", err)
+			}
+		}
+		if e.Vuln != nil {
+			if ref, err = s.UpdateVulnerabilities(ctx, e.Updater, e.Fingerprint, e.Vuln); err != nil {
+				return fmt.Errorf("updating vulnerabilities: %w", err)
+			}
```
hdonnay, line 61: I think this makes it technically possible to have both Vulnerabilites and Enrichments in the same UpdateOperation. Nothing in-tree would produce output that way, so I don't think this is a big deal.

hdonnay (approved):
LGTM pending rebase

PR #1167, opened 2023-12-01
updater: Support delta updates

crozzy:
TODO:

- [x] Doc strings
- [x] Move added SQL to embedded .sql file? (not going to tackle this in the same PR as we're now touching an existing method)

### DIFF
--- a/datastore/postgres/updatevulnerabilities.go
+++ b/datastore/postgres/updatevulnerabilities.go
@@ -32,7 +32,7 @@ var (
 			Name:      "updatevulnerabilities_total",
 			Help:      "Total number of database queries issued in the updateVulnerabilities method.",
 		},
-		[]string{"query"},
+		[]string{"query", "is_delta"},
 	)
 	updateVulnerabilitiesDuration = promauto.NewHistogramVec(
 		prometheus.HistogramOpts{
@@ -41,7 +41,7 @@ var (
 			Name:      "updatevulnerabilities_duration_seconds",
 			Help:      "The duration of all queries issued in the updateVulnerabilities method",
 		},
-		[]string{"query"},
+		[]string{"query", "is_delta"},
 	)
 )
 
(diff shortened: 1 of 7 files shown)


datastore/postgres/matcher_store.go
```diff
@@ -83,3 +86,102 @@ func (s *MatcherStore) RecordUpdaterStatus(ctx context.Context, updaterName stri
 func (s *MatcherStore) RecordUpdaterSetStatus(ctx context.Context, updaterSet string, updateTime time.Time) error {
 	return recordUpdaterSetStatus(ctx, s.pool, updaterSet, updateTime)
 }
+
+func (s *MatcherStore) GetLatestVulnerabilities(ctx context.Context, updater string) ([]*claircore.Vulnerability, error) {
```
hdonnay, line 90: Does this need to load all the data eagerly? An iterator+map façade might get us a reduced memory footprint?

crozzy: That'd probably be an improvement :thinking: I'd have to think about a map that lazily fetches it's values though... I think we really need to nail-down what these DeltaUpdaters are going to do with the information. I tried to capture some of that in the test `testUpdater.DeltaParse()`.

hdonnay: I think the choice of using a client-calculated digest is going to bite us here, as we would need to expose it in order to do an equality check.

crozzy: I have updated the PR and feel a lot better about the flow now. The updater just needs to signal that the data it's fetching are deltas and the manager will choose the `DeltaUpdatevulnerabilities()`. All the updating is captured in this function. It's job is to:
* Get existing vulnerabilities (just ID and name will suffice)
* Delete any existing vulnerabilities that appear in the new batch
* Create the uo (as is tradition)
* Add the existing vulns to the new uo
* Persist the new vulns to the DB (as is tradition)

datastore/postgres/updatevulnerabilities.go
```diff
@@ -51,9 +51,48 @@ var (
...
+// DeltaUpdateVulnerabilities implements vulnstore.Updater.
+//
+// It shares functionality with UpdateVulnerabilities but with delta
+// support, so the process goes:
+// * Create a new UpdateOperation
+// * Query existing vulnerabilities for the updater
+// * Discount and vulnerabilities with newer updates and deleted vulnerabilities
+// * Update the associated updateOperation for the remaining existing vulnerabilities
+// * Insert the new vulnerabilities
+// * Associate new vulnerabilities with new updateOperation
```
hdonnay, line 66:
```suggestion
//   - Create a new UpdateOperation
//   - Query existing vulnerabilities for the updater
//   - Discount and vulnerabilities with newer updates and deleted vulnerabilities
//   - Update the associated updateOperation for the remaining existing vulnerabilities
//   - Insert the new vulnerabilities
//   - Associate new vulnerabilities with new updateOperation
```
See the "Lists" section in https://go.dev/doc/comment

datastore/postgres/updatevulnerabilities.go
```diff
@@ -51,9 +51,48 @@ var (
 // provided vulnerabilities and computes a diff comprising the removed
 // and added vulnerabilities for this UpdateOperation.
 func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string, fingerprint driver.Fingerprint, vulns []*claircore.Vulnerability) (uuid.UUID, error) {
+	return s.updateVulnerabilities(ctx, updater, fingerprint, vulns, nil, false)
+}
+
+// DeltaUpdateVulnerabilities implements vulnstore.Updater.
+//
+// It shares functionality with UpdateVulnerabilities but with delta
+// support, so the process goes:
```
hdonnay, line 60: I don't understand this, please reword.

datastore/postgres/updatevulnerabilities.go
```diff
@@ -84,18 +123,18 @@ func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string
 	)
 	ctx = zlog.ContextWithValues(ctx, "component", "internal/vulnstore/postgres/updateVulnerabilities")
```
hdonnay, line 124: I'd like the logs to indicate if this is in the `delta` flow or not.

datastore/postgres/updatevulnerabilities.go
```diff
@@ -117,6 +156,64 @@ func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string
 		Str("ref", ref.String()).
 		Msg("update_operation created")
 
+	if delta {
+		// Get existing vulns
+		// The reason this still works even though the new update_operation
+		// is already created is because the latest_update_operation view isn't updated until
+		// the end of this function.
```
hdonnay, line 163: 🧠

datastore/postgres/updatevulnerabilities.go
```diff
@@ -117,6 +156,64 @@ func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string
...
+		start = time.Now()
+		rows, err := s.pool.Query(ctx, selectExisting, updater)
+		if err != nil {
+			return uuid.Nil, fmt.Errorf("failed to get existing vulns: %w", err)
+		}
+		updateVulnerabilitiesCounter.WithLabelValues("selectExisting").Add(1)
+		updateVulnerabilitiesDuration.WithLabelValues("selectExisting").Observe(time.Since(start).Seconds())
+
+		oldVulns := make(map[string][]string)
+		for rows.Next() {
```
hdonnay, line 173: Need to call `rows.Close` and check `rows.Err`

datastore/postgres/updatevulnerabilities.go
```diff
@@ -117,6 +156,64 @@ func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string
...
+			}
+			for _, delName := range deletedVulns {
+				// If we have an existing vuln that has been signaled
+				// as deleted by the updater then delete it so it doesn't
+				// get associated with the new update_operation.
+				delete(oldVulns, delName)
+			}
+		}
+		start = time.Now()
+		// Update oldVulns' update_operations
```
hdonnay, line 203: This comment seems misleading. My read of what should/does happen is that the current update operation has the set of retained vulnerabilites associated with it.

datastore/postgres/updatevulnerabilities.go
```diff
@@ -117,6 +156,64 @@ func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string
...
+		for _, vs := range oldVulns {
+			for _, vID := range vs {
+				_, err := tx.Exec(ctx, assocExisting, uoID, vID)
+				if err != nil {
+					return uuid.Nil, fmt.Errorf("could not update old vulnerability with new UO: %w", err)
+				}
+			}
+		}
```
hdonnay, line 211: Two loops here feels kind of bad, but I guess it's okay. The alternative I hit on was building an indexed slice data structure, which is probably not worth it in a first implementation.

libvuln/updates/manager.go
hdonnay, line 1: LGTM

libvuln/jsonblob/jsonblob.go
hdonnay, line 1: LGTM

libvuln/driver/updater.go
hdonnay, line 1: LGTM

datastore/updater.go
hdonnay, line 1: LGTM

datastore/postgres/updatevulnerabilities.go
hdonnay, line 1: The metrics collected in the `updateVulnerabilites` method should probably also have a `is_delta` or similar label so that we don't end up with a histogram that's bimodal in a way that's not particularly useful.

hdonnay (changes requested):
Needs a few changes, then the metrics addition.

hdonnay (approved):

PR #1179, opened 2023-12-13
updater/driver: add the ability to signal "removal" in an update

hdonnay:


updater/driver/v1/types.go
```diff
@@ -26,11 +26,18 @@ type EnrichmentRecord struct {
...
-	Package       []Package
-	Distribution  []Distribution
-	Repository    []Repository
+	// The following are indexed into via Vulnerability structs.
+	Package      []Package
+	Distribution []Distribution
+	Repository   []Repository
+	// Removed is a list of Vulnerabilities (the "Name" member) removed in the
+	// current set of Vulnerabilities.
+	Removed []string
```
crozzy, line 40: Just to mention: I had the realization that for us the `Name` is generic i.e. for some VEX CVE files we generate 100s of vulnerabilities with the same name (CVE-2023-001 for samba, CVE-2023-001 for samba-libs etc). This didn't actually change much except that the `UpdateVulnerabilities()` method now parses existing vulnerabilities into a `map[string][]string`, where the key is the name (CVE-2023-001) and the value is a list of all vuln ID that share that name.

crozzy: All that to say, I think this is still valid, it's just worth pointing out the nuanced difference

hdonnay: Right -- this isn't in use yet, but I imagine this is in the current Updater's namespace

PR #1181, opened 2023-12-14
fix(jsonblob): check for error in Next

RTann:
After I thought #1180 would fix the panic I ran into, it turns out it's still there. I believe this is the culprit.

`Next` does the following loop:

```
if l.err != nil {
	return false
}

var vs []claircore.Vulnerability
for l.err = l.dec.Decode(&l.de); l.err == nil; l.err = l.dec.Decode(&l.de) { ... }
l.e = l.next
return true
```

After the loop, it is assumed `l.err == nil`, but that's not necessarily true. It's possible `Decode` returns a non-nil error, which should make `Next` return `false`.


libvuln/jsonblob/jsonblob.go
```diff
@@ -103,6 +103,9 @@ func (l *Loader) Next() bool {
 			}
 		}
 	}
+	if l.Err() != nil {
```
RTann, line 106: this may not be completely correct. This allows EOF to still return `true` which is probably ok, but only if `l.next != nil`, I think

hdonnay: By my thinking, the error check on the first entry should catch this?

PR #1187, opened 2023-12-18
claircore: fix `Layer` finalizer

hdonnay:
@RTann reported some unexpected memory usage, which led to figuring out that this finalizer doesn't work as intended.

This fixes the finalizer, then fixes the issues reported by the subsequent panics.

### DIFF
--- a/layer.go
+++ b/layer.go
@@ -46,18 +46,13 @@ type LayerDescription struct {
 // called on an uninitialized Layer will report errors and may panic.
 type Layer struct {
 	noCopy noCopy
-	// NoFun is used to both implement the panicing Finalizer and to track
-	// initialization.
+	// Final is used to implement the panicking Finalizer.
 	//
 	// Without a unique allocation, we cannot set a Finalizer (so, when filling
 	// in a Layer in a slice). A pointer to a zero-sized type (like a *struct{})
-	// is not unique. So this camps on a heap allocation that's
-	// self-referential.
-	//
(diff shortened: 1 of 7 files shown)


layer.go
```diff
@@ -138,14 +132,7 @@ func (l *Layer) Close() error {
 		_, file, line, _ := runtime.Caller(1)
 		panic(fmt.Sprintf("%s:%d: Layer closed twice", file, line))
 	}
-	if l.noFun == nil {
-		return errors.New("claircore: Close: uninitialized Layer")
```
RTann, line 142: Do we want to allow users to close uninitialized layers?

libindex/fetcher.go
```diff
@@ -193,7 +193,7 @@ func (a *RemoteFetchArena) fetchInto(ctx context.Context, l *claircore.Layer, cl
 			return err
 		}
 		*cl = closeFunc(func() error {
-			return errors.Join(f.Close(), r.Close())
+			return errors.Join(f.Close(), r.Close(), l.Close())
```
RTann, line 196: 💯

layer.go
hdonnay: Good question. Previously, we needed to de-reference a pointer to do the close (well, in addition to the receiver) and now we don't. Leaving it as-is might make failure paths easier?

After looking at callsites, I don't think it actually will, so I'll go back to the previous behavior.

crozzy (approved):
LGTM

PR #1193, opened 2024-01-04
java: grow buffers to the right capacity before use

frostmar:
Avoids reallocations when buffers grow repeatedly for large files, and oversizing (as buffer capacity doubles when capacity is reached).

When indexing an unusually large image we've seen, that has deeply nested jars, this reduces peak process size from 18.9Gb to 11.1Gb.

---

The jar spec says nested jars must be "stored" (not compressed) in the outer jar, so in principal the parsing of nested jars could re-use sections of the outer file's buffer, greatly reducing memory use ... but those changes would be much more intrusive so I don't know if you'd want to go that way.

### DIFF
--- a/java/jar/jar.go
+++ b/java/jar/jar.go
@@ -275,6 +275,7 @@ func extractInner(ctx context.Context, p srcPath, z *zip.Reader) ([]Info, error)
 		}
 		defer rc.Close()
 		buf.Reset()
+		buf.Grow(int(fi.Size()))
 		h.Reset()
 		sz, err := buf.ReadFrom(io.TeeReader(rc, h))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


hdonnay (approved):
Good catch!

hdonnay:
> The jar spec says nested jars must be "stored" (not compressed) in the outer jar, so in principal the parsing of nested jars could re-use sections of the outer file's buffer, greatly reducing memory use ... but those changes would be much more intrusive so I don't know if you'd want to go that way.

This is interesting. Do you have any more information on this?

frostmar:
Perhaps this is only a definitive statement for SpringBoot, but seems plausible for any nest-jar classloader that's aiming to be efficient
https://docs.spring.io/spring-boot/docs/current/reference/html/executable-jar.html#appendix.executable-jar.jarfile-class
https://docs.spring.io/spring-boot/docs/current/reference/html/executable-jar.html#appendix.executable-jar.restrictions

PR #1201, opened 2024-01-09
ruby: enable matcher by default

RTann:
Just realized this isn't enabled by default. After https://github.com/quay/claircore/pull/1125 was merged, I think (hope) the matcher is ready to be turned on by default.


crozzy (approved):
LGTM

PR #1202, opened 2024-01-10
libindex: add some guess at layer preallocation sizing

hdonnay:
This is an idea I had that I threw together.


hdonnay:
This uses the mean, but probably shouldn't. Something like a running median would be better, but would require an indexed skip list implementation.

PR #1204, opened 2024-01-10
java: update matching error with vuln name

RTann:
It's hard to debug this when I don't know the vulnerability's name

### DIFF
--- a/java/matcher.go
+++ b/java/matcher.go
@@ -40,7 +40,7 @@ func (*Matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, v
 
 	// Check for missing upper version
 	if !decodedVersions.Has("fixed") && !decodedVersions.Has("lastAffected") {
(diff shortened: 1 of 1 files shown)


crozzy (approved):

RTann:
Another thing to consider: should we error here or just say it's not vulnerable? Is it worth getting no results vs a false negative for one vuln?

@hdonnay @crozzy

crozzy:
> Another thing to consider: should we error here or just say it's not vulnerable? Is it worth getting no results vs a false negative for one vuln?
> 
> @hdonnay @crozzy

I think this is a far point but in theory I think you would get results just none from that particular matcher (not sure if that's better or worse). Currently, it looks like there is mixture of patterns, I'm leaning towards log and return nil.

hdonnay (approved):

crozzy:
Is he going to try :fast_forward: ?

hdonnay:
.jpg)

PR #1207, opened 2024-01-12
rhel: explicit RHEL Atomic Host support

RTann:
Without the updated regex, ClairCore did correctly detect `rhel:7` for this scenario, but it seemed more due to luck from thei line in the `os-release`: `REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"`

This change make the support explicit

### DIFF
--- a/rhel/distributionscanner.go
+++ b/rhel/distributionscanner.go
@@ -19,7 +19,7 @@ var (
 	_ indexer.DistributionScanner = (*DistributionScanner)(nil)
 	_ indexer.VersionedScanner    = (*DistributionScanner)(nil)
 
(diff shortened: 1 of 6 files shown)


crozzy:
> This LGTM, should we update the scanner version?

Traditionally I'd say yes, but if the result of the index report is the same maybe not?

RTann:
> > This LGTM, should we update the scanner version?
> 
> Traditionally I'd say yes, but if the result of the index report is the same maybe not?

From what I can tell, the results are the same, as ClairCore pretty coincidentally correctly found the distribution, so I don't think a Scanner version bump is necessary.

crozzy (approved):

PR #1210, opened 2024-01-15
fix OSV matching algorithms

RTann:
GHSA-qw4h-3xjj-84cc showed that `fixed` and `last_affected` are not required. See https://ossf.github.io/osv-schema/#requirements and https://ossf.github.io/osv-schema/#unfixed-vulnerability-example to see it is valid for neither one to exist in an entry.

This change accounts for this and updates the 'vulnerable' algorithm as follows:

1. Parse the `IndexRecord`'s version. If we cannot parse this, we cannot correctly know if it's vulnerable to anything
2. Compare the `IndexRecord` to the `introduced` version, if it exists
3. Compare the `IndexRecord` to the `fixed` version, if it exists
4. Compare the `IndexRecord` to the `last_affected` version, if it exists
5. Default to claim the package is vulnerable. This may introduce false positives/noise, but this may be unlikely enough to be ok so users can at least verify this on their end. This is the chosen default, as we cannot prove the package is not vulnerable.

### DIFF
--- a/java/matcher.go
+++ b/java/matcher.go
@@ -2,9 +2,9 @@ package java
 
 import (
 	"context"
-	"fmt"
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/libvuln/driver"
+	"github.com/quay/zlog"
 	"net/url"
 )
 
@@ -29,8 +29,17 @@ func (*Matcher) Query() []driver.MatchConstraint {
 }
 
 func (*Matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, vuln *claircore.Vulnerability) (bool, error) {
+	// TODO(ross): This is a common pattern for OSV vulnerabilities. This should be moved into
+	// a common place for all OSV vulnerability matchers.
+
 	if vuln.FixedInVersion == "" {
 		return true, nil
+	}
+
+	// Parse the package first. If it cannot be parsed, it cannot properly be analyzed for vulnerabilities.
+	rv, err := parseMavenVersion(record.Package.Version)
+	if err != nil {
+		return false, err
 	}
 
 	decodedVersions, err := url.ParseQuery(vuln.FixedInVersion)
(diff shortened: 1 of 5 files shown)


crozzy:
@RTann is it worth trying to encapsulate this logic in one place (the osv package probably), given that we're parsing the same spec for all of these?

RTann:
> @RTann is it worth trying to encapsulate this logic in one place (the osv package probably), given that we're parsing the same spec for all of these?

Perhaps. The one concern would be that the version types are all different, so I'd have to look into making the function generic enough to accept all the different types. Due to time constraints, I don't know if I'd be able to look into that. Would a TODO in each file suffice?

java/matcher.go
```diff
@@ -33,46 +33,61 @@ func (*Matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, v
...
+	decodedVersions, err := url.ParseQuery(vuln.FixedInVersion)
 	if err != nil {
 		return false, err
 	}
 
-	v2, err := parseMavenVersion(upperVersion)
-	if err != nil {
-		return false, err
+	introduced := decodedVersions.Get("introduced")
+	// If there is an introduced version, check is the package's version is lower.
```
crozzy, line 48: Nit: spelling

python/matcher.go
```diff
@@ -49,50 +50,51 @@ func (*Matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, v
...
 		return false, err
 	}
-	upperVersion := decodedVersions.Get("fixed")
-	lastAffected := decodedVersions.Get("lastAffected")
-	switch {
-	case upperVersion != "":
-		uv, err := pep440.Parse(upperVersion)
+
+	introduced := decodedVersions.Get("introduced")
+	// If there is an introduced version, check is the package's version is lower.
```
crozzy, line 55: nit: spelling

ruby/matcher.go
```diff
@@ -36,61 +35,65 @@ func (*Matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, v
...
 
-	switch {
-	case decodedVersions.Has("fixed") && rv.Compare(uv) >= 0:
-		return false, nil
-	case decodedVersions.Has("lastAffected") && rv.Compare(uv) > 0:
-		return false, nil
-	case decodedVersions.Has("introduced"):
-		introduced := decodedVersions.Get("introduced")
+	introduced := decodedVersions.Get("introduced")
+	// If there is an introduced version, check is the package's version is lower.
```
crozzy, line 54: nit: spelling

crozzy (approved):
Two things:
1) As mentioned, I think it'd be nice to have this logic be generic to avoid the separate implementations getting out of sync (as has kind of happened here with the python matcher).
2) It's worth calling out that point 5 is not changing the current logic, in fact, it's more like; we can't make any case for why this package isn't vulnerable.

RTann:
@crozzy I added the TODO to each matcher implementation and fixed the spelling issue

RTann:
> As mentioned, I think it'd be nice to have this logic be generic to avoid the separate implementations getting out of sync (as has kind of happened here with the python matcher).

Added TODOs to each matcher implementation

> It's worth calling out that point 5 is not changing the current logic, in fact, it's more like; we can't make any case for why this package isn't vulnerable.

updated PR description to reflect this

crozzy (approved):

PR #1213, opened 2024-01-16
nodejs: vuln matching support

RTann:
This requires https://github.com/quay/claircore/pull/1214 to be merged first

### DIFF
--- a/nodejs/testdata/indexreport-splunk-8.2.6.json
+++ b/nodejs/testdata/indexreport-splunk-8.2.6.json
@@ -1,0 +1,392 @@
+{
+  "manifest_hash": "sha256:a2f4094b86baa321e3555c4e225ebf0905eae7146fa4f59382403d2b0f15868c",
+  "state": "IndexFinished",
+  "packages": {
+    "4": {
+      "id": "4",
+      "name": "cssom",
+      "version": "0.3.0",
+      "kind": "binary",
+      "source": {
+        "id": "1",
+        "name": "",
+        "version": ""
+      },
+      "normalized_version": "semver:0.0.3.0.0.0.0.0.0.0"
+    },
+    "20": {
+      "id": "20",
+      "name": "parse5",
+      "version": "1.1.6",
+      "kind": "binary",
+      "source": {
+        "id": "1",
+        "name": "",
+        "version": ""
+      },
+      "normalized_version": "semver:0.1.1.6.0.0.0.0.0.0"
+    },
+    "22": {
+      "id": "22",
+      "name": "pdfkit",
+      "version": "0.1.5",
+      "kind": "binary",
+      "source": {
+        "id": "1",
+        "name": "",
+        "version": ""
+      },
+      "normalized_version": "semver:0.0.1.5.0.0.0.0.0.0"
+    },
+    "28": {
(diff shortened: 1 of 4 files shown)


nodejs/testdata/indexreport-splunk-8.2.6.json
```diff
@@ -0,0 +1,392 @@
...
+      "id": "4",
+      "name": "cssom",
+      "version": "0.3.0",
+      "kind": "binary",
+      "source": {
+        "id": "1",
+        "name": "",
+        "version": ""
+      },
+      "normalized_version": "semver:0.0.3.0.0.0.0.0.0.0"
```
RTann, line 15: This is populated via https://github.com/quay/claircore/pull/1214

nodejs/testdata/indexreport-splunk-8.2.6.json
RTann, line 1: This was generated by Clair using a version of ClairCore based on this branch.

updater/osv/osv.go
crozzy, line 1: Looking at:
```sql
select * from vuln where updater = 'osv/npm' and name = 'GHSA-6c3j-c64m-qhgq'
```
presumably for the empty ranges we're just never going to match anything for them

crozzy: Doing some analysis on this:
osv/go:
empty range = 155
non-empty = 2530

osv/npm:
empty range = 179
non-empty range = 12936

So I suppose this isn't a new thing, just can't think of the value of having these vulns, not an issue for this PR

crozzy: Interesting, before this PR the ratio of empty to non-empty for `osv/go` was different:
osv/go:
empty range = 1594
non-empty = 1096
:thinking: Is this expected from [this change?](https://github.com/quay/claircore/pull/1213/files#diff-f4ccdc13b708304dcc4583c99a6f56de609a15dab040a5c620fdde190c8fdf82R550)

RTann: What do you mean by empty ranges? The referenced change just improves a range's lower bound. Without that change, there are a good number of Go and Node.js vulnerabilities we ignore due to https://github.com/quay/claircore/blob/v1.5.20/updater/osv/osv.go#L627

https://github.com/quay/claircore/blob/v1.5.20/updater/osv/osv.go#L623 only tries to set a blank Upper.Kind to a (hopefully?) set Lower.Kind. It doesn't account for when Upper.Kind is set, but Lower.Kind is not, which happens when `Introduced == 0`

For Go, we missed vulnerabilities such as https://osv.dev/vulnerability/GHSA-4374-p667-p6c8

RTann: ok I played around with this, and I was looking into `GHSA-4374-p667-p6c8` (Go). I can't seem to figure out why the range is `empty`. Any ideas how that could be?

RTann: We looked into this offline, and it looks like `GHSA-4374-p667-p6c8` is populated in the DB properly.

To answer the original question, yes, we expect this change to affect Go as well. However, I didn't expect the number of empty ranges to increase. In fact, I still don't understand why there are any empty ranges

crozzy (approved):

PR #1214, opened 2024-01-17
nodejs: set package NormalizedVersion

RTann:


crozzy (approved):
SGTM, we have an issue about [overhauling the version type](https://issues.redhat.com/browse/PROJQUAY-6111) going forward, also we really should have tests for `FromSemver` but can be done separately IMO

crozzy (approved):

PR #1215, opened 2024-01-17
all: various whiteout file support updates

RTann:


RTann:
can someone remind me how these test cases were written because there is no chance i'm manually updating these

crozzy (commented):
The package scanner versions should probably be update here

RTann:
> The package scanner versions should probably be update here

I always forget that, thanks

crozzy (approved):
LGTM

RTann:
sorry @crozzy I will need one more review. I realized I forgot the log discard stuff in the python packagescanner, so I fixed that, and I also updated it to match what Ruby does to avoid future mishaps

crozzy (approved):

PR #1217, opened 2024-01-20
rhel-updater: improve error messages

RTann:
I was testing something and saw the following:

```
{"level":"error","host":"scanner-v4-indexer-69c646f78f-n9fvh","component":"rhel/internal/common/Updater.Get","error":"received status code 'ƥ' querying mapping url","time":"2024-01-20T22:51:22Z","message":"error updating mapping file"}
```

I think it'd be more useful to see the number rather than the decoded value.

Also, I noticed a usage of `%v` with `fmt.Errorf`, and I believe `%w` is preferred, so I included that, too


hdonnay (approved):
LGTM -- I suspect someone meant `resp.Status`, but just the code should be fine.

PR #1223, opened 2024-01-23
zreader: remove zlib support

hdonnay:
Noticed on #1213 that the zreader test was failing. I was able to reproduce reliably by just upping the number of test runs.

Ultimately, the zlib header is just 2 bytes and not sufficiently unique enough to get the autodetection to work. Currently we don't consume anything that's "bare" zlib (outside of HTTP requests, which are handled by `net/http` transparently), so I think it's fine to just drop it.

We never hit a problem in the fetcher because gzip and uncompressed tars have a sufficiently different header from the zlib header that it doesn't trip.

### DIFF
--- a/internal/zreader/compression_string.go
+++ b/internal/zreader/compression_string.go
@@ -11,13 +11,12 @@ func _() {
 	_ = x[KindGzip-0]
 	_ = x[KindZstd-1]
 	_ = x[KindBzip2-2]
-	_ = x[KindZlib-3]
-	_ = x[KindNone-4]
+	_ = x[KindNone-3]
 }
 
-const _Compression_name = "KindGzipKindZstdKindBzip2KindZlibKindNone"
+const _Compression_name = "KindGzipKindZstdKindBzip2KindNone"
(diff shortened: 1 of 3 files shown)


RTann (approved):
LGTM. I was comparing Clair v4 and Clair v4, and I noticed Clair v2 supported xz. Any plans for supporting that in v4, or nah?

RTann:
> Supported xz for what?

It was to account for a potential `tar.xz` layer, though I checked https://github.com/opencontainers/image-spec/blob/main/media-types.md, and it looks like only `tar`, `tar+gzip` and `tar+zstd` are supported.

Guess bzip2 isn't hurting anyone, but is that a possibility?

hdonnay:
I don't think so. In the case of the fetcher specifically, it might be worth to use the `Detect` API and make sure what it reports is what's expected. This package is meant to just be a general way to handle it, to head off issues like when some Red Hat XML data just started coming back in bz2 despite the endpoint or Content-Type not changing.

PR #1225, opened 2024-01-25
osv: set FixedInVersion for SEMVER ecosystems

RTann:
For example, when looking at Go vulnerabilities, ClairCore does not specify any vulnerabilities are fixed. Only affected or not, essentially. This change includes `FixedByVersion`, when available.

### DIFF
--- a/updater/osv/osv.go
+++ b/updater/osv/osv.go
@@ -557,6 +557,7 @@ func (e *ecs) Insert(ctx context.Context, skipped *stats, name string, a *adviso
 						ver, err = semver.NewVersion(ev.Fixed)
 						if err == nil {
 							v.Range.Upper = claircore.FromSemver(ver)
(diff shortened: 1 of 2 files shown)


updater/osv/osv.go
```diff
@@ -570,6 +571,7 @@ func (e *ecs) Insert(ctx context.Context, skipped *stats, name string, a *adviso
 						if err == nil {
 							nv := ver.IncPatch()
 							v.Range.Upper = claircore.FromSemver(&nv)
+							v.FixedInVersion = nv.Original()
```
RTann, line 574: I'd be ok with leaving this out. We don't actually know the name of the version in which this is fixed

crozzy (approved):

crozzy (approved):

PR #1231, opened 2024-01-29
debug: add some debugging bits

hdonnay:
Add some scripts I cooked up while investigating some Alpine stuff.

May need more documentation in the prose documentation?


crozzy (approved):
LGTM, hopefully we'll be able to add to these as time goes on

PR #1240, opened 2024-02-07
rhel: allow deactivation of ContainerAPI fetching

BradLugo:
## Description

There's logic in the RHEL repository scanner to disable reaching out to the internet, but we currently set a default API endpoint during Configure to prevent that condition. These changes introduce a DisableAPI flag to allow the deactivation of the fetching from the Red Hat Container API.

### DIFF
--- a/rhel/repositoryscanner_test.go
+++ b/rhel/repositoryscanner_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"github.com/quay/claircore/pkg/tmp"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -31,6 +32,10 @@ func TestRepositoryScanner(t *testing.T) {
 		"rh-pkg-1-1": strings.NewReader(`{"data":[{"cpe_ids":["cpe:/o:redhat:enterprise_linux:8::computenode","cpe:/o:redhat:enterprise_linux:8::baseos"],"parsed_data":{"architecture":"x86_64","labels":[{"name":"architecture","value":"x86_64"}]}}]}`),
 	}
 	mappingData := strings.NewReader(`{"data":{"content-set-1":{"cpes":["cpe:/o:redhat:enterprise_linux:6::server","cpe:/o:redhat:enterprise_linux:7::server"]},"content-set-2":{"cpes":["cpe:/o:redhat:enterprise_linux:7::server","cpe:/o:redhat:enterprise_linux:8::server"]}}}`)
+	var mappingDataBytes bytes.Buffer
+	if _, err := io.Copy(&mappingDataBytes, mappingData); err != nil {
+		t.Fatal(err)
+	}
 
 	mux := http.NewServeMux()
 	mux.HandleFunc("/repository-2-cpe.json", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


rhel/repositoryscanner_test.go
```diff
@@ -100,6 +122,50 @@ func TestRepositoryScanner(t *testing.T) {
...
+					Key:  repositoryKey,
+					CPE:  cpe.MustUnbind("cpe:/o:redhat:enterprise_linux:7::server"),
+				},
+				{
+					Name: "cpe:/o:redhat:enterprise_linux:8::server",
+					Key:  repositoryKey,
+					CPE:  cpe.MustUnbind("cpe:/o:redhat:enterprise_linux:8::server"),
+				},
+			},
+			cfg:       &RepositoryScannerConfig{DisableAPI: true, API: esrv.URL, Repo2CPEMappingURL: "/", Repo2CPEMappingFile: f.Name()},
```
RTann, line 166: Can you add a test where we can't get any repo data because there is no file and `DisableAPI: true`?

BradLugo: I'm not entirely following, but I tried to add what I thought you were getting at. If the new changes don't seem quite right, can you rephrase your comment?

RTann: yeah this is what I meant, thanks

rhel/repositoryscanner_test.go
```diff
@@ -31,6 +32,10 @@ func TestRepositoryScanner(t *testing.T) {
 		"rh-pkg-1-1": strings.NewReader(`{"data":[{"cpe_ids":["cpe:/o:redhat:enterprise_linux:8::computenode","cpe:/o:redhat:enterprise_linux:8::baseos"],"parsed_data":{"architecture":"x86_64","labels":[{"name":"architecture","value":"x86_64"}]}}]}`),
 	}
 	mappingData := strings.NewReader(`{"data":{"content-set-1":{"cpes":["cpe:/o:redhat:enterprise_linux:6::server","cpe:/o:redhat:enterprise_linux:7::server"]},"content-set-2":{"cpes":["cpe:/o:redhat:enterprise_linux:7::server","cpe:/o:redhat:enterprise_linux:8::server"]}}}`)
+	mappingDataBytes := new(bytes.Buffer)
```
RTann, line 35: up to you, but you could also just do `var mappingDataBytes bytes.Buffer`

RTann (dismissed):

rhel/repositoryscanner_test.go
```diff
@@ -55,6 +60,23 @@ func TestRepositoryScanner(t *testing.T) {
 	srv := httptest.NewServer(mux)
 	defer srv.Close()
 
+	esrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		t.Fatal("external http request invoked when none was expected")
+	}))
+
+	f, err := os.CreateTemp("", "repository-2-cpe.json")
```
crozzy, line 67: nit but we have `pkg/tmp.NewFile()` and `.Close()` that will close the file and remove it.

rhel/repositoryscanner_test.go
```diff
@@ -55,6 +60,23 @@ func TestRepositoryScanner(t *testing.T) {
...
+	if err != nil {
+		t.Fatal("trying to create repository-2-cpe.json for FromMappingFile test", err)
+	}
+	defer os.Remove(f.Name())
+
+	if _, err := f.Write(mappingDataBytes.Bytes()); err != nil {
+		t.Fatal(fmt.Sprintf("trying to write %s for FromMappingFile test", f.Name()), err)
```
crozzy, line 74: `t.Fatalf` is useful here (and in the next block)

crozzy (changes requested):
LGTM, just a couple of nits

rhel/repositoryscanner_test.go
RTann: another option is using `t.TempDir()` which will auto-close once the test is done

BradLugo: A combination of the two seems pretty nice

rhel/repositoryscanner_test.go
BradLugo: 🤦 I swear I've written Go before

BradLugo:
Thanks for the feedback, yall!

PR #1241, opened 2024-02-07
dockerfile: additional expansions

hdonnay:
Saw some errors on quayio's instance related to variable expansions.
Checking the buildkit dockerfile frontend, it supports the whole compliment of POSIX expansions and bash-style replace (`${var/pat/word/}`).

This PR implements all the POSIX expansions and reports a dedicated error for the slash pattern replacement.

* * *
- [ ] Check changes after finding an exemplar problematic `Dockerfile`


hdonnay:
I've run this through the fuzzer for a while, with no crashers being reported.

    fuzz: elapsed: 30m1s, execs: 121110401 (0/sec), new interesting: 51 (total: 388)

PR #1242, opened 2024-02-09
rhel: check for syntax error when parsing content_manifests

crozzy:
Some layers in the wild will have invalid json in the content_manifests files, this change checks for json.Syntax errors and logs.


hdonnay (dismissed):
LGTM -- maybe we should think through what should happen if there are multiple results from the glob, and some aren't actually JSON

crozzy:
> LGTM -- maybe we should think through what should happen if there are multiple results from the glob, and some aren't actually JSON

Yeah, it looks like we only take into account the first entry in the glob now so if it's invalid I suppose it's technically possible that subsequent ones would be valid. I thought the expectation was that there should be one per layer though, not sure in what situation there would be more, maybe squashed layers?

hdonnay (approved):

hdonnay:
Yeah, squashed layers, maybe. But that brings up the question of what the "right" one would be, then

PR #1250, opened 2024-02-12
rpm: validate native DBs before querying them

crozzy:
In production there are a large number of layers that cannot be
scanned because they contain native DBs that do not adhear to the
expected schema. This adds a Validate() function to the nativeDB
interface and calls it before attempting to extract packages.


rpm/sqlite/sql/validate.sql
```diff
@@ -0,0 +1 @@
+SELECT name FROM pragma_table_list() where name='Packages';
```
hdonnay, line 1:
```suggestion
SELECT 1 FROM pragma_table_list() WHERE name = 'Packages';
```

rpm/sqlite/sqlite.go
```diff
@@ -88,5 +88,13 @@ func (db *RPMDB) AllHeaders(ctx context.Context) ([]io.ReaderAt, error) {
 	return r, nil
 }
 
+func (db *RPMDB) Validate(ctx context.Context) error {
+	var table string
+	return db.db.QueryRow(validate).Scan(&table)
```
hdonnay, line 93:
```suggestion
	var ignore int64
	err := db.db.QueryRow(validate).Scan(&ignore)
	switch {
	case errors.Is(err, nil):
	case errors.Is(err, sql.ErrNoRows):
	    return errors.New("not an rpm database")
	default:
	    return err
	}
	return nil
```

hdonnay (changes requested):
I have some opinions on the query; I think just returning a static value and then "rewriting" the error in the expected failure mode makes more sense.

hdonnay (approved):
LGTM w/ rebase

PR #1251, opened 2024-02-12
cicd: add some dependabot grouping

hdonnay:
Added some groups for modules that tend to move in unison.


crozzy (approved):
Oh cool

hdonnay:
Will merge after CI runs.

PR #1252, opened 2024-02-12
rhel: skip distribution when oracle-release file exists

RTann:


hdonnay:
Having not thought very hard about this: IMO, this logic should be in the coalesce step. The `rhel` code shouldn't need to know about every "superset" distribution, the coalescer should have that sort of knowledge to be able to pick the least-wrong one.

RTann:
> Having not thought very hard about this: IMO, this logic should be in the coalesce step. The `rhel` code shouldn't need to know about every "superset" distribution, the coalescer should have that sort of knowledge to be able to pick the least-wrong one.

I thought about that, but I think the problem is that we are dealing with two ecosystems each with its own coalescer: `rhel` and `rpm`.

The `rhel` ecosystem uses the `rhel` distribution scanner and `rhel` coalescer while the `rpm` ecosystem uses the `aws`, `oracle`, `suse`, and `photon` distribution scanners and the `linux` coalescer.

ClairCore is essentially claiming Oracle Linux under both the `rhel` and `rpm` ecosystems. I'm not sure if the `rhel` coalescer will know about the Oracle distro findings since that was found in a different ecosystem

RTann:
> > Having not thought very hard about this: IMO, this logic should be in the coalesce step. The `rhel` code shouldn't need to know about every "superset" distribution, the coalescer should have that sort of knowledge to be able to pick the least-wrong one.
> 
> I thought about that, but I think the problem is that we are dealing with two ecosystems each with its own coalescer: `rhel` and `rpm`.
> 
> The `rhel` ecosystem uses the `rhel` distribution scanner and `rhel` coalescer while the `rpm` ecosystem uses the `aws`, `oracle`, `suse`, and `photon` distribution scanners and the `linux` coalescer.
> 
> ClairCore is essentially claiming Oracle Linux under both the `rhel` and `rpm` ecosystems. I'm not sure if the `rhel` coalescer will know about the Oracle distro findings since that was found in a different ecosystem

unless you meant to update `MergeSR` https://github.com/quay/claircore/blob/v1.5.21/indexer/controller/coalesce.go#L113 to check for the case when `RHEL` and `Oracle` are both distros? Though if we were to just detect that and then claim it's only `Oracle`, then that means we can't handle the theoretical case where the distro truly changes from RHEL to Oracle or vice versa.

I figured the best way to handle this was to check if `etc/redhat-release` and `etc/oracle-release` exist in the same layer which I guess a coalescer could do, but I'm not sure if I'd see the `oracle` distro scanner results in the `rpm` ecosystem can be seen by the `rhel` coalescer or vice versa

RTann:
> > > Having not thought very hard about this: IMO, this logic should be in the coalesce step. The `rhel` code shouldn't need to know about every "superset" distribution, the coalescer should have that sort of knowledge to be able to pick the least-wrong one.
> > 
> > 
> > I thought about that, but I think the problem is that we are dealing with two ecosystems each with its own coalescer: `rhel` and `rpm`.
> > The `rhel` ecosystem uses the `rhel` distribution scanner and `rhel` coalescer while the `rpm` ecosystem uses the `aws`, `oracle`, `suse`, and `photon` distribution scanners and the `linux` coalescer.
> > ClairCore is essentially claiming Oracle Linux under both the `rhel` and `rpm` ecosystems. I'm not sure if the `rhel` coalescer will know about the Oracle distro findings since that was found in a different ecosystem
> 
> unless you meant to update `MergeSR` https://github.com/quay/claircore/blob/v1.5.21/indexer/controller/coalesce.go#L113 to check for the case when `RHEL` and `Oracle` are both distros? Though if we were to just detect that and then claim it's only `Oracle`, then that means we can't handle the theoretical case where the distro truly changes from RHEL to Oracle or vice versa.
> 
> I figured the best way to handle this was to check if `etc/redhat-release` and `etc/oracle-release` exist in the same layer which I guess a coalescer could do, but I'm not sure if I'd see the `oracle` distro scanner results in the `rpm` ecosystem can be seen by the `rhel` coalescer or vice versa

looked into it more, and it looks like each ecosystem's coalesce would only know about its own distributions due to https://github.com/quay/claircore/blob/v1.5.21/indexer/controller/coalesce.go#L58 which only returns dists which match the ecosystem's distro scanners

hdonnay (approved):
Makes sense in light of those explanations, thanks.

Good to merge with rebase and green CI.

PR #1253, opened 2024-02-13
gobin: parse semver and ensure segments fit in claircore.Version

crozzy:
Some gobin versions are nanosecond timestamps and cannot be converted to a semver using the library. This change brings the semver parsing in-house and truncates any version segments to ensure they'll fit in int32s.

### DIFF
--- a/gobin/exe.go
+++ b/gobin/exe.go
@@ -6,10 +6,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"regexp"
+	"strconv"
 	"strings"
 	_ "unsafe" // for error linkname tricks
 
-	"github.com/Masterminds/semver"
 	"github.com/quay/zlog"
 
 	"github.com/quay/claircore"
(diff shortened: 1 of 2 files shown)


gobin/exe_test.go
hdonnay, line 1: Fix grouping

gobin/exe.go
```diff
@@ -174,3 +169,42 @@ func toPackages(ctx context.Context, out *[]*claircore.Package, p string, r io.R
...
 	return nil
 }
+
+var versionRegex = regexp.MustCompile(`^v?([0-9]+)(\.[0-9]+)?(\.[0-9]+)?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?$`)
+var ErrInvalidSemVer = errors.New("invalid semantic version")
+
+// ParseVersion will return a claircore.Version of type semver given
+// a valid semantic version. If the string is not a valid semver it
+// will return an ErrInvalidSemVer.
+func ParseVersion(ver string) (c claircore.Version, err error) {
```
hdonnay, line 179: In the case of a go pseudo version, should it discard the date? I think I could go either way.

hdonnay (changes requested):

gobin/exe.go
crozzy: `ParseVersion()` behaves the same as `semver.NewVersion()` + `version.FromSemver()` in the current code so I believe anything that is not in the Major, Minor, Patch triplet is discarded. If I'm not mistaken we're talking about versions like this: 
```
v0.0.0-20170426191122-ca1404ee6e83
```
In which case I'm not sure how we fit that into the current `claircore.Version` so it's possibly moot. In future world, maybe it makes sense to cut off the hash and save the date (presumably there is some general acceptance that `v0.0.0-20170426191122-aaaaaaaaaaaa > v0.0.0-20170426191121-zzzzzzzzzzzz`)

crozzy: Just for any observers, the problem versions are currently versions like 
```
v1.2022081210521689159162.2023071210561689159372
```
For "valid" semvers like:
```
v0.0.0-20170426191122-ca1404ee6e83
``` 
they are currently parsed as just `v0.0.0` (or `0` to be more precise)

hdonnay (approved):

PR #1255, opened 2024-02-14
nodejs: skip invalid package.json

RTann:
When testing, we ran into empty package.json files. The ruby package scanner skips invalid gems, so I figure we should skip invalid package.json files

### DIFF
--- a/nodejs/packagescanner.go
+++ b/nodejs/packagescanner.go
@@ -90,16 +90,18 @@ func (s *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircor
 	}
 
 	ret := make([]*claircore.Package, 0, len(pkgs))
+	var invalidPkgs []string
 	for _, p := range pkgs {
 		f, err := sys.Open(p)
 		if err != nil {
-			return nil, fmt.Errorf("nodejs: unable to open file: %w", err)
+			return nil, fmt.Errorf("nodejs: unable to open file %q: %w", p, err)
 		}
 
 		var pkgJSON packageJSON
 		err = json.NewDecoder(bufio.NewReader(f)).Decode(&pkgJSON)
 		if err != nil {
-			return nil, fmt.Errorf("nodejs: unable to decode package.json file: %w", err)
+			invalidPkgs = append(invalidPkgs, p)
+			continue
 		}
 
 		pkg := &claircore.Package{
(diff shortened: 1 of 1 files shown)


nodejs/packagescanner.go
```diff
@@ -93,13 +93,17 @@ func (s *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircor
...
 		if err != nil {
-			return nil, fmt.Errorf("nodejs: unable to open file: %w", err)
+			return nil, fmt.Errorf("nodejs: unable to open file %q: %w", p, err)
 		}
 
 		var pkgJSON packageJSON
 		err = json.NewDecoder(bufio.NewReader(f)).Decode(&pkgJSON)
 		if err != nil {
-			return nil, fmt.Errorf("nodejs: unable to decode package.json file: %w", err)
+			zlog.Warn(ctx).
```
hdonnay, line 102:
```suggestion
			zlog.Debug(ctx).
```
I don't think this warrants a warning.

If you want some logging that appears by default, this could collect up the paths of weird `package.json` files and emit an info log after this loop.

hdonnay (approved):

hdonnay (changes requested):
Sorry, wrong radio button: see previous comment.

nodejs/packagescanner.go
RTann: I opted to collect the invalid paths and emit a warning log. If you think warning is wrong let me know

hdonnay:
I'm unsure that "warn" is the right level, as there's nothing an operator can do when a layer that triggers the lines is processed

RTann:
> I'm unsure that "warn" is the right level, as there's nothing an operator can do when a layer that triggers the lines is processed

I interpret warn as something that's not expected and unusual, yet we will continue to make progress regardless. I can switch to info if that will make it consistent with the rest of claircore

RTann:
I'll make it debug

nodejs/packagescanner.go
```diff
@@ -113,7 +115,7 @@ func (s *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircor
 		if sv, err := semver.NewVersion(pkgJSON.Version); err == nil {
 			pkg.NormalizedVersion = claircore.FromSemver(sv)
 		} else {
-			zlog.Info(ctx).
+			zlog.Warn(ctx).
```
RTann, line 118: @hdonnay would you prefer I revert this?

hdonnay: Yeah. As a rule of thumb, we shouldn't warn if the request will continue. At least that's what I wrote down previously: https://github.com/quay/claircore/blob/main/docs/contributor/logging.md#levels

RTann: reverted. should be ready for (hopefully) one last pass

hdonnay (approved):
LGTM

PR #1256, opened 2024-02-14
ndb: bound slots data read to avoid over-reading

crozzy:
We error on some ndb files due to reading too many bytes and bleeding into the blob data (BlbS). This change bounds the data read to num slots * slots per page.

### DIFF
--- a/rpm/ndb/package.go
+++ b/rpm/ndb/package.go
@@ -42,7 +42,7 @@ func (db *PackageDB) Parse(r io.ReaderAt) error {
 	db.slot = make([]pkgSlot, 0, lastIndex)
 	b = b[:slotSize]
 	// Read every populated slot (these should be contiguous) and populate the lookup table.
-	for off := int64(slotStart * slotSize); ; off += slotSize {
+	for off := int64(slotStart * slotSize); off < int64(db.NPages*pageSz); off += slotSize {
 		if _, err := r.ReadAt(b, off); err != nil {
 			return fmt.Errorf("ndb: package: unable to read slot @%d: %w", off, err)
 		}
--- a/rpm/ndb/package_test.go
+++ b/rpm/ndb/package_test.go
@@ -28,13 +28,14 @@ func TestLoadPackage(t *testing.T) {
 				t.Fatal(err)
 			}
 			pkgf := bytes.NewReader(b)
+
 			var pkg PackageDB
 			if err := pkg.Parse(pkgf); err != nil {
-				t.Fatal(err)
+				t.Fatal("error parsing Packages file", err)
 			}
 			rds, err := pkg.AllHeaders(ctx)
 			if err != nil {
-				t.Fatal(err)
+				t.Fatal("error getting AllHeaders", err)
 			}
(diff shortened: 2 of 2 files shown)


rpm/ndb/package_test.go
```diff
@@ -16,7 +16,7 @@ func TestLoadPackage(t *testing.T) {
 	ctx := context.Background()
 
 	dir := os.DirFS("testdata")
-	ms, err := fs.Glob(dir, "Packages*.db")
+	ms, err := fs.Glob(dir, "Packages_unreadable.db")
```
hdonnay, line 19: Inadvertent change?

hdonnay (changes requested):

hdonnay:
There's also something wrong, as it reports way fewer packages on the same database before and after this change.

rpm/ndb/ndb.go
```diff
@@ -14,8 +14,9 @@ var le = binary.LittleEndian
 
 // Used throughout the various DBs.
 const (
-	slotSize  = 4 * 4
-	slotStart = 2
+	slotSize           = 4 * 4
+	slotStart          = 2
+	slotEntriesPerPage = 4096 / 16
```
hdonnay, line 19: Where does this come from?

crozzy: https://github.com/rpm-software-management/rpm/blob/master/lib/backend/ndb/rpmpkg.c#L175
https://github.com/rpm-software-management/rpm/blob/master/lib/backend/ndb/rpmpkg.c#L177

crozzy:
> There's also something wrong, as it reports way fewer packages on the same database before and after this change.

Oh I know the issue, updating

rpm/ndb/ndb.go
hdonnay, line 19:
```suggestion
	slotEntriesPerPage = pageSz / slotSize
```

Please forgive my inconsistent naming.

rpm/ndb/package.go
```diff
@@ -42,7 +42,7 @@ func (db *PackageDB) Parse(r io.ReaderAt) error {
 	db.slot = make([]pkgSlot, 0, lastIndex)
 	b = b[:slotSize]
 	// Read every populated slot (these should be contiguous) and populate the lookup table.
-	for off := int64(slotStart * slotSize); ; off += slotSize {
+	for off := int64(slotStart * slotSize); off < int64(db.NPages*pageSz); off += slotSize {
```
hdonnay, line 45: So this works by virtue of reading the whole page and hitting the end. I suppose the extra time to read a few K isn't the end of the world.

hdonnay (changes requested):

rpm/ndb/ndb.go
hdonnay: Although this is unused, so just remove it

crozzy: :facepalm:

hdonnay (approved):
LGTM. I'll make a note to add fixtures for comparing the databases to a desired list.

PR #1259, opened 2024-02-15
cicd: add go1.22 to CI

hdonnay:


crozzy (approved):
LGTM Weird that code coverage would have gone down, it seems like 1.21 and 1.22 would deal with the same code footprint

hdonnay:
Yeah, I'm unsure if it's the go instrumentation changing? I've looked at this a little locally and can't work out why the coverage fluctuates.

hdonnay:
arg, this is merged. GitHub managed to process the push to my fork that happened after the push to origin in the wrong order.

PR #1260, opened 2024-02-15
rpm: replace error with log when HOFFPAGE length is incorrect

crozzy:
We see DBs in the wild whose offset pages evidently report the expected length of the item incorrectly. No longer returning the error allows these incorrectly formatted DB files to be read at the expense of correctness.

### DIFF
--- a/rpm/bdb/bdb.go
+++ b/rpm/bdb/bdb.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+
+	"github.com/quay/zlog"
 )
 
 // PackageDB is the "pkgdb" a.k.a. "Packages", the raw package data.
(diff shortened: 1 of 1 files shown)


crozzy:
I'm not sure this is the best solution but currently don't have any others.

crozzy:
Forgot the tests were implicit

rpm/bdb/bdb.go
```diff
@@ -233,9 +235,13 @@ func (db *PackageDB) AllHeaders(_ context.Context) ([]io.ReaderAt, error) {
...
 			// Double-check we'll read the intended amount.
 			if got, want := r.Size(), int64(offpg.Length); got != want {
-				return nil, fmt.Errorf("bdb: expected data length botch: %d != %d", got, want)
+				zlog.Info(ctx).
+					Int64("got", got).
+					Int64("want", want).
+					Msg("bdb: expected data length botch")
 			}
 			ret = append(ret, &r)
+
```
hdonnay, line 244:
```suggestion
```
Any reason for this?

crozzy: nope, removed

hdonnay (approved):

PR #1264, opened 2024-02-19
gobin: allow for empty version segment strings

crozzy:
This error gets surfaced when the version is in the form 1.18. We see this in production due to the runtime version being sometimes reported as go1.18 (`go` is trimmed before the version conversion). Addition error context added for easier debugging.


hdonnay (approved):
lgtm

PR #1269, opened 2024-02-22
bug: os-release whitespace

hdonnay:
Found some files in the wild that had weird whitespace.

Systemd and friends handled it fine, so this fixes things to match.

I also updated the tests and docs while in here.


crozzy (approved):
LGTM

PR #1275, opened 2024-02-27
all: account for language package overwrites

RTann:
This was originally discovered in StackRox Scanner V2: https://github.com/stackrox/stackrox/issues/7033

StackRox now offers a Scanner based on ClairCore, which also has this same problem. The issue is that ClairCore does not consider the fact that the image build system may decide to overwrite the language package instead of deleting and recreating it.

This was demonstrated in the OCI image `namloc2001/nodesem:a`.

Each language's package scanner implements `DefaultRepoScanner`, which means there is only a single (related) repository per layer, and it only exists if the layer has (related) packages. Each language's coalescer is also more-or-less the same (Go's was unique, but the unique parts of it are unnecessary, as those checks will definitely hold true based on how to Ecosystem is set up), so I decided to make a single, shared coalescer for the languages.


RTann:
Though this does not explicitly touch package scanners, this still may merit a reindex. Thoughts?

language/coalescer.go
hdonnay, line 1: Needs documentation on package and exports.

gobin/coalescer.go
hdonnay, line 1: This seems to be copy-pasted; why?

language/coalescer.go
```diff
@@ -21,9 +21,14 @@ func (c *coalescer) Coalesce(_ context.Context, ls []*indexer.LayerArtifacts) (*
 		Packages:     map[string]*claircore.Package{},
 		Repositories: map[string]*claircore.Repository{},
 	}
-
-	for _, l := range ls {
-		// If we didn't find at least one npm repo in this layer
+	// Similar to ir.Packages, except instead of mapping
+	// id -> package, it maps packageDB -> package.
+	// For langauge packages, it is possible the
+	// packageDB is overwritten.
```
hdonnay, line 27: Yeah this package needs to document this assumption, and we need to check that the uses actually work that way.

- [x] gobin
- [ ] java
- [ ] nodejs
- [ ] python
- [ ] ruby

hdonnay (changes requested):

hdonnay:
> Though this does not explicitly touch package scanners, this still may merit a reindex. Thoughts?

Yeah, needs the version changed in every indexer that's moving to it.

gobin/coalescer.go
RTann: The rest were definite copy/pastes of each other. This one is slightly unique, so I copied over the related change, but kept the fact this is still mildly different from the rest. Do you think this coalescer should just match the rest?

language/coalescer.go
RTann: just curious why `gobin` is already checked

hdonnay: I read through it to see what it did.

gobin/coalescer.go
RTann: I updated it to also move this into the language-agnostic coalescer. The parts of this which were unique were not very necessary (I think)

language/coalescer_test.go
```diff
@@ -0,0 +1,155 @@
...
+		for _, id := range e.RepositoryIDs {
+			r := ir.Repositories[id]
+			if got, want := r.Name, "npm"; got != want {
+				t.Errorf("got: %q, want: %q", got, want)
+			}
+		}
+	}
+}
+
+func TestCoalescer_package_overwrite(t *testing.T) {
```
crozzy, line 76: It's kind of an annoying thing to test, I'm thinking maybe it could be table driven so we can add cases if needed but might take some thinking.

language/coalescer.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		for _, pkg := range l.Pkgs {
+			if seen, exists := packages[pkg.PackageDB]; exists {
+				// If the package was renamed or has a different version in a high layer,
+				// then we consider this a different package and ignore the
+				// original in the lower layer.
+				if pkg.Name != seen.Name || pkg.Version != seen.Version {
+					continue
+				}
+				// The name and version is the same, so delete the entry related to the higher
```
crozzy, line 53: nit `The name and version are the same`

language/coalescer.go
```diff
@@ -0,0 +1,70 @@
...
+var _ indexer.Coalescer = (*coalescer)(nil)
+
+type coalescer struct{}
+
+// NewCoalescer returns a new common programming language coalescer.
+func NewCoalescer(_ context.Context) (indexer.Coalescer, error) {
+	return &coalescer{}, nil
+}
+
+// Coalesce implements [indexer.Coalescer].
```
crozzy, line 21: The logic here is actually not trivial w/r/t which packages end up in the final index report, I think this docstring could do with more explanation.

crozzy (commented):
Looks like it's in good shape, a few comments.

language/coalescer_test.go
RTann: This does test a very specific scenario, and I'm not sure if I can think of other similar tests which would be useful to put here. Is it ok to just leave as-is?

language/coalescer.go
crozzy: @RTann was this checked?

RTann: I think I'd need more clarity on the exact ask. Depending on the image builder, the file may either be updated or deleted and recreated. The changes I made attempts to account for both. Not sure if that's what the question is asking about or something else

RTann:
@hdonnay @crozzy can you take a look at this when you get the chance? Thanks

java/packagescanner.go
```diff
@@ -72,7 +72,7 @@ type Scanner struct {
 func (*Scanner) Name() string { return "java" }
 
 // Version implements scanner.VersionedScanner.
-func (*Scanner) Version() string { return "7" }
+func (*Scanner) Version() string { return "8" }
```
RTann, line 75: I just updated this from 6 to 7 in https://github.com/quay/claircore/pull/1467. Depending on when this is merged, I can revert this

RTann:
> LGTM. I think there are improvements we can make to the code we're moving over (clarify comments, add obvious error in test), but we can save it for a future cleanup. lmk if you want to post those reviews (I ended up deleting them after realizing I was commenting on code that was being moved)

@BradLugo please leave any comments you have!

crozzy (approved):
LGTM

hdonnay (approved):
SGTM

PR #1278, opened 2024-03-06
toolkit: convert WFN.Value to value receiver

crozzy:
When using the go sql driver WFN will not be recognized as a driver.Valuer (only *WFN). As WFN is direct value in claircore's structs this makes it fiddly, the change allows both pointer and value to benefit from Value().


hdonnay:
I think this was the case in the initial migration into the toolkit module and that broke things -- double-check the git log

PR #1282, opened 2024-03-08
chore: don't fail changelog label checker when there's no data

crozzy:
Adding the -r command to xargs (--no-run-if-empty) to avoid the command failing if the gh command does not yield results.


crozzy:
https://github.com/quay/claircore/actions/runs/8194677643

hdonnay (approved):
🙏 been meaning to fix this, thanks for taking it up

PR #1287, opened 2024-03-15
[WIP] ROX-22258: Add RHCOS Ecosystem

Maddosaurus:
Apologies for the notification spam, GitHub defaulted the PR to upstream instead of my fork. 
I'm leaving it open as we might as well use it once it is ready :sweat_smile:

### DIFF
--- a/rhel/rhcos/distributionscanner.go
+++ b/rhel/rhcos/distributionscanner.go
@@ -1,0 +1,79 @@
+package rhcos
+
+import (
+	"context"
+	"fmt"
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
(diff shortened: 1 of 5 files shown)


hdonnay:
This doesn't look right -- I don't think any vulnerabilities are reported in terms of CoreOS, only in terms of RHEL.

This is maybe different with the VEX data, but I don't think so.

Maddosaurus:
I'll definitely check in the old Scanner codebase how/what we matched.
IIRC, it is very similar, yet slightly different to RHEL. 
Right now, the goal of this PR is to introduce the RHCOS ecosystem enough so I can run it against an RHCOS system and see results.
Apologies again for the premature review notification

PR #1292, opened 2024-03-18
libindex: add fallback path for O_TMPFILE not being supported

hdonnay:


libindex/tempfile_linux.go
```diff
@@ -1,24 +1,75 @@
...
+	"sync"
 
 	"golang.org/x/sys/unix"
 )
 
+// Holds: dir → error
+var fsChecks sync.Map
+
+// CheckFS returns the memoized result of trying to open a tempfile in the
+// specified directory.
```
crozzy, line 16: What is the reason for attempting to use `O_TMPFILE` (and falling back to attempting to use `CreateTempDir()`) and then based on the results actually doing one or the other as opposed to just keeping all the logic in `openTemp` (including the "cached" results). Not sure if that's a good explanation, like why check when the actual opening attempts can serve as the check?

crozzy (commented):
Nice, just a couple of clarifying comments

libindex/tempdir_unix.go
```diff
@@ -0,0 +1,19 @@
+package libindex
+
+import (
+	"os"
+)
+
+// FixTemp modifies "dir" according to the documented defaults.
+//
+// See [NewRemoteFetchArena].
+func fixTemp(dir *string) {
```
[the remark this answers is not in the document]
hdonnay: no real reason, that's probably a better way to go about it

libindex/tempfile_linux.go
hdonnay: This grew out of trying to use `statfs`, so that's probably why. I'll stream-line it.

libindex/tempfile_linux.go
```diff
@@ -1,24 +1,63 @@
...
+			break
+		}
+		f, err = os.CreateTemp(dir, "fetcher.*")
+	default:
+		panic("unreachable")
+	}
+	if !loaded {
+		setTmp(dir, ok)
+	}
+	if !ok && err == nil {
```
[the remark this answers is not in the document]
crozzy: I managed to hack up a local reproducer (and it is hacky), and got: 
```write /tmp/fetcher.1385048962: file already closed```
hence why I got to this logic. It's also possible something is bauked in my local setup.

File: libindex/fetcher.go [hunk 4]
```diff
@@ -54,11 +48,36 @@ type RemoteFetchArena struct {
...
+// does not support O_TMPFILE, files may linger in the event of a process
+// crashing or an unclean shutdown. Operators should either use a different
+// filesystem or arrange for periodic cleanup via [systemd-tmpfiles(8)] or
+// similar.
+//
+// In a containerized environment, operators may need to mount a directory or
+// filesystem on "/var/tmp".
+//
+// On OSX, temporary files are not unlinked from the filesystem upon creation,
+// because an equivalent to Linux's "/proc/self/fd" doesn't seem to exist.
```
BradLugo, line 68: Hey @hdonnay, do you mind elaborating on this or rephrasing the comment? I'm having a hard time grokking it.

> ... temporary files are not unlinked from the filesystem upon creation,
> because an equivalent to Linux's "/proc/self/fd" doesn't seem to exist.

I couldn't find a reason `"/proc/self/fd"` was required for "unlinking" temporary files. I'm also not sure what unlinking in this context means; my assumption based on the word choice would be that the implementation does some sort of symlinking and we need to unlink them at some point, but I don't think that's the case.

> temporary files are not unlinked from the filesystem upon creation

I'm particularly confused by this. If we're on Linux, do other temporary files get unlinked when we create our temporary files?

---

fwiw, I'm not sure if others do this, but I know Ross and I both run the ACS Scanner on macOS devices, so sorting this out would help us out a lot (in other words, I'm not trying to be pedantic with this comment 🙂)

crozzy (commented):
LGTM passing baton to @BradLugo

crozzy (approved):

libindex/fetcher.go
BradLugo: (also happy to move this discussion to another PR or elsewhere if we want to unblock merging this)

hdonnay: No worries.

"Unlink" in this context is a filesystem/syscall term-of-art: it means making a name no longer visible. This isn't _exactly_ deleting a file because the semantics change based on the the type of object, the underlying filesystem, and outstanding file descriptions. Out of a combination of battle scars and habit, I tend to say "unlink" when I'm specifically thinking of [`unlink(2)`].

With regards to `/proc/self/fd`, that's a directory containing magic symlinks (actual terminology, see [`symlink(7)`]) representing open file **descriptors**. Opening one of them creates a new file descriptor backed by a new file **description** (see the NOTES section of [`open(2)`]). This means that unlike `dup(2)`, the uses of the two file descriptors are completely independent (e.g. seeking one will not change the offset on the other).

The Linux code uses this to open files that aren't linked into the filesystem (meaning they'll never outlast the process, because only the process has them open) but can still be reopened as many times as needed to create independent file descriptors.

AFAICT, there's not a way to do this on OSX. Looking at FreeBSD man pages, [`fdescfs(5)`] acts the same way but only if mounted with `nodup`, which I don't think is the default. There is mention on that manpage about a `O_EMPTY_PATH` to `openat(2)`, but I can't find any information on whether that exists on OSX and don't have a machine to test with. To work around this, the code simply doesn't unlink the file and re-opens it by name as needed. The downside is that the files are only removed as long as the program shuts down cleanly.

[`unlink(2)`]: https://man7.org/linux/man-pages/man2/unlink.2.html
[`symlink(7)`]: https://man7.org/linux/man-pages/man7/symlink.7.html
[`open(2)`]: https://man7.org/linux/man-pages/man2/open.2.html
[`fdescfs(5)`]: https://man.freebsd.org/cgi/man.cgi?query=fdescfs&sektion=5&apropos=0&manpath=FreeBSD+14.0-RELEASE+and+Ports

BradLugo: Thank you for the detailed explanation! I need to dive deeper into all this info, so please feel free to consider this discussion non-blocking for the PR

hdonnay:
will merge with green CI

Issue #1291, opened 2024-03-18
False positive vulnerability match for postgresql-client version


cathyyoung:
We've seen instances of false positive vulnerability matching for the package `postgresql-client` which appears to be due to the package's non-standard versioning scheme. 

We are currently using claircore v1.5.19.

```bash
# apt show postgresql-client
Package: postgresql-client
Version: 13+225
```

Clair finds the package in the vulnerability report:
```json
"7884339": {
      "id": "7884339",
      "name": "postgresql-client",
      "version": "13+225",
      "kind": "binary",
      "source": {
        "id": "7884338",
        "name": "postgresql-common (225)",
        "version": "13+225",
        "kind": "source"
      },
      "arch": "all"
    },
```

But then incorrectly matches three vulnerabilities to it, e.g.:

```json
  "1844653": {
      "id": "1844653",
      "updater": "debian/updater/bullseye",
      "name": "CVE-2019-3466 postgresql-common",
      "description": "The pg_ctlcluster script in postgresql-common in versions prior to 210 didn't drop privileges when creating socket/statistics temporary directories, which could result in local privilege escalation.",
      "issued": "0001-01-01T00:00:00Z",
      "links": "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-3466",
      "severity": "",
      "normalized_severity": "Unknown",
      "package": {
        "id": "",
        "name": "postgresql-client",
        "version": "",
        "kind": "binary"
      },
      "distribution": {
        "id": "",
        "did": "debian",
        "name": "Debian GNU/Linux",
        "version": "11 (bullseye)",
        "version_code_name": "bullseye",
        "version_id": "11",
        "arch": "",
        "cpe": "",
        "pretty_name": "Debian GNU/Linux 11 (bullseye)"
      },
      "repository": {},
      "fixed_in_version": "0:210"
    },
```

It seems likely this is due to the non-standard versioning scheme being used but shouldn't return false positives.

hdonnay:
Seems like the dpkg `Source` information may contain a version component? Need to look at some documentation to see if that's a convention or load bearing.

hdonnay:
I couldn't find the status files spec, but it looks like [`dpkg-query(1)`](https://man7.org/linux/man-pages/man1/dpkg-query.1.html) parses it out, so we should probably do similar.

PR #1293, opened 2024-03-18
dpkg: improve `Source` key handling

hdonnay:
This is how tools seem to interpret this. I couldn't find official documentation on a quick search.

Closes: #1291


crozzy:
I don't think this solves the problem because the debian vulns are always inserted as `binary` and the index record is still going to look something like this:
```json
    "378": {
      "id": "378",
      "name": "postgresql-client",
      "version": "15+248",
      "kind": "binary",
      "source": {
        "id": "377",
        "name": "postgresql-common",
        "version": "248",
        "kind": "source"
      },
      "arch": "all"
    },
```
So the version `15+248` will still be incorrectly matched to. Maybe something about the sourcemapper is creating an incorrect assumption about the versions.

hdonnay:
I think it addresses the source/binary issue now.

debian/updater.go
```diff
@@ -125,26 +126,10 @@ var (
 func (f *Factory) UpdaterSet(ctx context.Context) (driver.UpdaterSet, error) {
 	s := driver.NewUpdaterSet()
 
-	ds, err := f.findReleases(ctx, f.mirror)
```
crozzy, line 128: I think we need to keep `findReleases()` or at least call `mkDist()` on the found distributions, otherwise when we're parsing the JSON tracker data and trying to `getDist()` it will return an error (which is ignored currently).

hdonnay: ah, good catch

crozzy (approved):
LGTM

PR #1294, opened 2024-03-19
osv: parse database_specific severity when no CVSS severity is defined

crozzy:
Occasionally there are OSV advisories that don't include any severity information in the `.severity` object but they do contain a severity in the `.database_specific` object. This change attempts to parse that severity if we don't get a severity from the native `.severity` object.


crozzy:
original idea from @RTann and @celek https://github.com/quay/claircore/pull/1142

hdonnay (approved):
LGTM

PR #1120, opened 2023-10-23
jsonblob: add a disk buffering step

hdonnay:
This eagerly encodes Vulnerabilities and EnrichmentRecords to JSON and writes them to a disk-backed buffer. This should reduce memory consumption as long as the system actually allows for flushing disk cache.

See-also: PROJQUAY-5915

PR #1295, opened 2024-03-20
Fix jsonblob memory usage

ludydoo:
Trying to fix large memory usage when loading vulnerabilities

### DIFF
--- a/libvuln/jsonblob/jsonblob.go
+++ b/libvuln/jsonblob/jsonblob.go
@@ -73,7 +73,6 @@ func (l *Loader) Next() bool {
 		return false
 	}
 
-	var vs []claircore.Vulnerability
 	for l.err = l.dec.Decode(&l.de); l.err == nil; l.err = l.dec.Decode(&l.de) {
 		id := l.de.Ref
 		// If we just hit a new Entry, promote the current one.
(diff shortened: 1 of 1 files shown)


ludydoo:
@hdonnay

crozzy:
Interesting, thanks for the PR, can I trouble you for some memory usage metrics? On the surface LGTM unless I'm missing something about `vs`'s usage.

crozzy:
https://github.com/quay/clair/blob/main/Documentation/contribution/commit_style.md also, the CI will be happy if you reformat the commit message

RTann:
I asked about this before: https://github.com/quay/claircore/pull/1120#discussion_r1396327236:

> It's the backing slice for the returned `[]*Vulnerability`.

So I guess the question is: is that actually true, or do we have 2 separate slices?

hdonnay:
I think this makes sense; it increases GC pressure in exchange for not having a backing slice that needs copies while growing. I generally dislike this because it ends up making a *lot* of objects if not coalesced in a slice. We can probably rationalize doing individual allocations because this shouldn't be used in a long-lived process.

RTann:
I am curious what the cost savings may be. From what I can tell, we are between:

* Current - two slices: 1 with `claircore.Vulnerability` structs and the other with pointers to these structs. Keep everything in contiguous memory
* Proposed - Single slice of `claircoire.Vulnerability` pointers pointing to structs allocated randomly in the heap (probably? May even be stack allocated and then jumps to the heap. Not sure)

Current method may technically utilize more space via slice struct headers, but otherwise, they are almost equal in terms of memory. May have capacity differences, too, but not sure. I think this may be rather minimal anyway

Does this sound right?

hdonnay:
Right, a backing slice means all the structs are one object to the GC and are pre-allocated based on the capacity. Not doing that means all the allocations are individual objects.

hdonnay:
@ludydoo This is probably good to go with a rebase and the commit message rewritten (something like `jsonblob: improve memory usage` for the first line).

<details>
<summary>rebase steps</summary>

Just adding a note here in case you're unfamiliar with a rebase workflow.
Conceptually, you're just taking your commits and applying the diff to the current `main` instead of where they were originally committed.
Assuming `origin` is the `quay/claircore` repo, and `fork` is your fork:

1. `git checkout BRANCH_NAME`
1. `git fetch origin`
1. `git rebase origin/main`
1. `git push --force-with-lease fork`

If you get a merge conflict in the `rebase` step, just replace the conflict block with the code you think should be there. If something's wrong, we'll catch it. If it's for an autogenerated file like `go.sum`, just recreate it.

If we asked you to change a commit message, add the `-i` flag to the `rebase` command and use the `reword` command.
</details>

hdonnay (approved):

PR #1300, opened 2024-03-29
rpm: extract filename information

hdonnay:
This should allow us to extract a list of known file patterns for use later in the Indexer pipeline.


rpm/native_db.go
```diff
@@ -159,14 +164,65 @@ func (i *Info) Load(ctx context.Context, h *rpm.Header) error {
...
 		}
 	}
 	return nil
 }
 
+// FilePatterns is a regular expression for *any* file that may need to be
+// recorded alongside a package.
+//
+// The tested strings are absolute paths.
+var filePatterns = regexp.MustCompile(`^.*/[^/]+\.jar$|^/usr/lib/python[23]\.[0-9]+/site-packages/[^/]+\.egg-info/PKG-INFO$`)
```
jvdm, line 217: Should we add Node.js (`package.json`), Ruby (`.gemspec`), and potential binaries (`/bin`, `/usr/bin`, etc) to match against the Go parser?

hdonnay: I can add those patterns, sure.

hdonnay:
Added tests

rpm/native_db.go
```diff
@@ -159,14 +164,84 @@ func (i *Info) Load(ctx context.Context, h *rpm.Header) error {
...
+	// [fs.WalkDirFunc] while this package (and dpkg, etc) can tell that another
+	// indexer will find those files relevant.
+	//
+	// [yara]: https://github.com/VirusTotal/yara
+	pat := []string{
+		`^.*/[^/]+\.jar$`, // Jar files
+		`^.*/site-packages/[^/]+\.egg-info/PKG-INFO$`, // Python packages
+		`^.*/package.json$`,                           // npm packages
+		`^.*/[^/]+\.gemspec$`,                         // ruby gems
+		`^/usr/bin/[^/]+$`,                            // any executable
```
crozzy, line 233: Adding the binary pattern adds an extra 1100 files for a vanilla UBI8 layer (562 -> 1641)

hdonnay: is that too many?

If we were to blue-sky redesign the detection API, it should probably be based around registering intents, so that something like this could only cross-check if another component cared about one of these files.

crozzy: I'm not really sure, they should only exist once in the DB with junction table (`file_scanartifact`) providing the link from file-> to layer, but the same DB can be in a lot of layers. If an layer installing an RPM package is built on top of a base layer containing an RPM db does the RPM DB get counted twice and deduped in the coalesce process?

> If we were to blue-sky redesign the detection API, it should probably be based around registering intents, so that something like this could only cross-check if another component cared about one of these files.

yeah agreed

hdonnay: So which ones do we include, then?

crozzy: Any pattern that is specific to a language ecosystem?

hdonnay: So we never catch any go executables, because they're named just like C programs?

crozzy: I think it's a tradeoff, there are a few packages that provide hundreds of binaries that we will never consult but then again it is a file-detector and there will be interesting files in there, so I'm fine keeping them in.

crozzy (approved):

PR #1303, opened 2024-04-03
datastore: avoid update_operation timing window by creating in txn

frostmar:
Previously the `update_operation` row was created outside the transaction which created vulns associated with that update_operation. This meant other updaters could refresh the `latest_update_operation` materialized view, making the new `update_operation` used for reporting before any of it's vulns were committed.

The upshot was a window where **vulnerability reports could give false-negatives** for a short period (they couldn't see any vulns associated with the most recent `update_operation`, because those vulns hadn't been committed yet).

We've seen these false-negative reports for Debian images (updater: `debian/updater`), possibly because that's an updater for a source with a large number of vulns. The window from _`update_operation` row create_ --> _vulns committed_ would be larger than for many other updaters.

### DIFF
--- a/datastore/postgres/updatevulnerabilities.go
+++ b/datastore/postgres/updatevulnerabilities.go
@@ -121,7 +121,6 @@ func (s *MatcherStore) updateVulnerabilities(ctx context.Context, updater string
 			$3,
 			(SELECT id FROM vuln WHERE hash_kind = $1 AND hash = $2))
 		ON CONFLICT DO NOTHING;`
-		undo        = `DELETE FROM update_operation WHERE id = $1;`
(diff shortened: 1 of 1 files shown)


crozzy (dismissed):
This looks good, it seems to make more sense to create the update operation inside the transaction, the first time loading won't make a difference but during every subsequent update this should ensure that no update operations are queried before their vulns are commited.

crozzy:
I rebased and fixed a conflict, @frostmar please take another look

frostmar:
@crozzy thanks for the rebase, looks good to me

crozzy (approved):

PR #1305, opened 2024-04-07
fix(tarfs): follow hardlinks in  `ReadFile`

BradLugo:
## Description

In `ReadFile`, we either follow symlinks or attempt to read the amount of data from the tar as specified by the `Size` field in `tar.Header`. When we come across a hardlink, however, we essentially tell the caller that the file is empty since the `Size` is `0`.

These changes to `ReadFile` are largely based on the changes @hdonnay made to `Open` for handling hardlinks: https://github.com/quay/claircore/commit/48317f8b4266d1e0f1478bcadef470d9865c61c2

### DIFF
--- a/pkg/tarfs/tarfs.go
+++ b/pkg/tarfs/tarfs.go
@@ -457,10 +457,23 @@ func (f *FS) ReadFile(name string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	if i.h.FileInfo().Mode().Type()&fs.ModeSymlink != 0 {
+
+	dataSize := i.h.Size
+	typ := i.h.FileInfo().Mode().Type()
+	var r *tar.Reader
+	switch {
+	case typ.IsRegular() && i.h.Typeflag != tar.TypeLink:
+		r = tar.NewReader(io.NewSectionReader(f.r, i.off, i.sz))
+	case typ.IsRegular() && i.h.Typeflag == tar.TypeLink || typ&fs.ModeSymlink != 0: // is hardlink or symlink
 		return f.ReadFile(i.h.Linkname)
+	default:
+		// Pretend all other kinds of files don't exist.
+		return nil, &fs.PathError{
+			Op:   op,
+			Path: name,
+			Err:  fs.ErrExist,
+		}
 	}
-	r := tar.NewReader(io.NewSectionReader(f.r, i.off, i.sz))
 	if _, err := r.Next(); err != nil {
 		return nil, &fs.PathError{
 			Op:   op,
@@ -468,7 +481,7 @@ func (f *FS) ReadFile(name string) ([]byte, error) {
 			Err:  err,
 		}
 	}
-	ret := make([]byte, i.h.Size)
+	ret := make([]byte, dataSize)
 	if _, err := io.ReadFull(r, ret); err != nil {
 		return nil, &fs.PathError{
 			Op:   op,
--- a/pkg/tarfs/tarfs_test.go
+++ b/pkg/tarfs/tarfs_test.go
@@ -460,6 +460,83 @@ func TestSymlinks(t *testing.T) {
 	}))
 }
 
+func TestReadFile(t *testing.T) {
+	type tarfsFile struct {
+		Header tar.Header
+		Data   []byte
+	}
+
+	tmp := t.TempDir()
+	setupAndRun := func(openErr bool, tarfsFiles []tarfsFile, chk func(*testing.T, *FS)) {
+		t.Helper()
+		// This is a perfect candidate for using test.GenerateFixture, but
+		// creates an import cycle.
+		f, err := os.Create(filepath.Join(tmp, filepath.Base(t.Name())))
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer f.Close()
+
+		tw := tar.NewWriter(f)
+		for _, tarfsFile := range tarfsFiles {
+			h := tarfsFile.Header
+			h.Size = int64(len(tarfsFile.Data))
+			h.Format = tar.FormatGNU
+			if err := tw.WriteHeader(&h); err != nil {
+				t.Error(err)
+			}
+			_, err := tw.Write(tarfsFile.Data)
+			if err != nil {
+				t.Error(err)
+			}
+		}
+		if err := tw.Close(); err != nil {
+			t.Error(err)
+		}
+
+		sys, err := New(f)
+		t.Log(err)
+		if (err != nil) != openErr {
+			t.Fail()
+		}
+
+		if chk != nil {
+			chk(t, sys)
+		}
+	}
+
+	t.Run("Hardlink", func(t *testing.T) {
+		originalData := []byte(`Hello, World!`)
+		abData := make([]byte, len(originalData))
+		copy(abData, originalData)
+
+		setupAndRun(false, []tarfsFile{
+			{
+				Header: tar.Header{Name: `a/`},
+			},
+			{
+				Header: tar.Header{Name: `a/b`},
+				Data:   abData,
+			},
+			{
+				Header: tar.Header{
+					Name:     `a/c`,
+					Typeflag: tar.TypeLink,
+					Linkname: `a/b`,
+				},
+			},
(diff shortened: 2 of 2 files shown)


BradLugo:
> When we come across a hardlink, however, we essentially tell the caller that the file is empty since the Size is 0.

I wanted to note that I don't know if this behavior is consistent. For example, perhaps there's a flag you can pass to `tar` when creating the archive that sets the size header for hardlinks.

BradLugo:
Another point of discussion: instead of reimplementing a bunch of logic from `Open` in `ReadFile`, should we instead call `Open` in `ReadFile`? Looks like that's what the `os` package does as well

BradLugo:
The more I think about this PR, the more I wonder if this is the right approach. `Open` and `ReadFile` share a lot of the same logic, and the `os` library's `ReadFile` calls its `Open` (https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/os/file.go;l=768).

It makes me wonder if we should move the `datasize` logic to `Open`  then call `Open` in `ReadFile`. The main problem with `Open` (IIUC) is that hardlinks still aren't *technically* implemented correctly, i.e., they return an inode with a `tar.Header` that has a `Size` of 0. That's not hugely problematic since the caller can follow the link themselves, but I think it goes against the idea of an inode

pkg/tarfs/tarfs.go
```diff
@@ -457,18 +457,38 @@ func (f *FS) ReadFile(name string) ([]byte, error) {
...
+	case typ.IsRegular() && i.h.Typeflag != tar.TypeLink:
+		r = tar.NewReader(io.NewSectionReader(f.r, i.off, i.sz))
+	case typ.IsRegular() && i.h.Typeflag == tar.TypeLink:
+		tgt, err := f.getInode(op, i.h.Linkname)
+		if err != nil {
+			return nil, err
+		}
+		r = tar.NewReader(io.NewSectionReader(f.r, tgt.off, tgt.sz))
+		dataSize = tgt.h.Size
+	case typ&fs.ModeSymlink != 0: // typ.IsSymlink()
```
RTann, line 474: is it possible `typ.IsRegular() && i.h.Typeflag != tar.TypeLink` and `typ&fs.ModeSymlink != 0` are both true? If so, we should make this the first case, otherwise symlinks won't be handled correctly

pkg/tarfs/tarfs.go
```diff
@@ -457,18 +457,38 @@ func (f *FS) ReadFile(name string) ([]byte, error) {
...
 	}
-	if i.h.FileInfo().Mode().Type()&fs.ModeSymlink != 0 {
+
+	dataSize := i.h.Size
+	typ := i.h.FileInfo().Mode().Type()
+	var r *tar.Reader
+	switch {
+	case typ.IsRegular() && i.h.Typeflag != tar.TypeLink:
+		r = tar.NewReader(io.NewSectionReader(f.r, i.off, i.sz))
+	case typ.IsRegular() && i.h.Typeflag == tar.TypeLink:
```
RTann, line 467: curious would all this just be equivalent to `return f.ReadFile(i.h.Linkname)`? If so, should we just combine this case and the symlink check into a single case?

RTann:
> The more I think about this PR, the more I wonder if this is the right approach. `Open` and `ReadFile` share a lot of the same logic, and the `os` library's `ReadFile` calls its `Open` (https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/os/file.go;l=768).
> 
> It makes me wonder if we should move the `datasize` logic to `Open` then call `Open` in `ReadFile`. The main problem with `Open` (IIUC) is that hardlinks still aren't _technically_ implemented correctly, i.e., they return an inode with a `tar.Header` that has a `Size` of 0. That's not hugely problematic since the caller can follow the link themselves, but I think it goes against the idea of an inode

I like the idea of implementing this as closely to `os.ReadFile` as possible, especially since `Open` and `ReadFile`, as implemented here, share a lot of the same logic

pkg/tarfs/tarfs.go
RTann: No, symlinks are not "regular"

BradLugo:
> > The more I think about this PR, the more I wonder if this is the right approach. `Open` and `ReadFile` share a lot of the same logic, and the `os` library's `ReadFile` calls its `Open` (https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/os/file.go;l=768).
> > It makes me wonder if we should move the `datasize` logic to `Open` then call `Open` in `ReadFile`. The main problem with `Open` (IIUC) is that hardlinks still aren't _technically_ implemented correctly, i.e., they return an inode with a `tar.Header` that has a `Size` of 0. That's not hugely problematic since the caller can follow the link themselves, but I think it goes against the idea of an inode
> 
> I like the idea of implementing this as closely to `os.ReadFile` as possible, especially since `Open` and `ReadFile`, as implemented here, share a lot of the same logic
> 
> Also, `Open` uses the hardlink's "target" size, no?

Ya know... When I was digging into this before, I saw the note at the bottom of https://pkg.go.dev/archive/tar#Reader.Read and assumed the `tarfs.file.h` (i.e., the `*tar.Header` of the hardlink) we were setting was problematic, but after looking more thoroughly when writing my first response, I think you're right. Since we get `*tar.Reader` with the correct `offset` and `sz`, store it in `tarfs.file.h`, which (IIUC) is what `tar.Reader.Read` ultimately looks at, I think we avoid the case of getting `io.EOF` on hardlinks.

So never mind. Now I want to use `Open` even more

---

I guess it really comes down to the hint at the top of `ReadFile`:
https://github.com/quay/claircore/blob/813e7ccb43fb9607da6388dd71094c3fe72ef4e0/pkg/tarfs/tarfs.go#L452-L454
I'm open to having my mind changed, but maintaining these two functions that are very similar but diverge due to the allocation of a `tarfs.file` struct in such a critical code path isn't worth it. The PR proves it.

hdonnay (approved):
LGTM

On the rebase, can I trouble you to rewrite the commit summary line to not have `fix(tarfs)` but rather just `tarfs`? We try to follow something closer to the go commit style than "conventional commits".

pkg/tarfs/tarfs_test.go
```diff
@@ -460,6 +460,83 @@ func TestSymlinks(t *testing.T) {
...
+		Header tar.Header
+		Data   []byte
+	}
+
+	tmp := t.TempDir()
+	setupAndRun := func(openErr bool, tarfsFiles []tarfsFile, chk func(*testing.T, *FS)) {
+		t.Helper()
+		// This is a perfect candidate for using test.GenerateFixture, but
+		// creates an import cycle.
+		f, err := os.Create(filepath.Join(tmp, path.Base(t.Name())))
```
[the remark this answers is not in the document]
hdonnay: I think I used `path` throughout and swapped specific uses because `io/fs` mandates forward-slash paths and go just handles forward-slash paths, so it's moot unless the path could be absolute.

hdonnay:
I'll look though how this is handled in #669

BradLugo:
> LGTM
> 
> On the rebase, can I trouble you to rewrite the commit summary line to not have `fix(tarfs)` but rather just `tarfs`? We try to follow something closer to the go commit style than "conventional commits".

Yup, will do!
Is this documented anywhere? If not, I can open up another PR and add it somewhere.

---

Before we merge this, do you mind giving your thoughts on https://github.com/quay/claircore/pull/1305#issuecomment-2046286920? I wonder if we can/should remove `ReadFile` altogether

crozzy:
> > LGTM
> > On the rebase, can I trouble you to rewrite the commit summary line to not have `fix(tarfs)` but rather just `tarfs`? We try to follow something closer to the go commit style than "conventional commits".
> 
> Yup, will do! Is this documented anywhere? If not, I can open up another PR and add it somewhere.
> 
> Before we merge this, do you mind giving your thoughts on [#1305 (comment)](https://github.com/quay/claircore/pull/1305#issuecomment-2046286920)? I wonder if we can/should remove `ReadFile` altogether

https://github.com/quay/claircore/blob/main/docs/contributor/commit-style.md

PR #1310, opened 2024-04-08
chore: update go version to 1.21.9

crozzy:
This go version includes an fix for a vulnerability in the net/http package. https://pkg.go.dev/vuln/GO-2024-2687


hdonnay:
Trying to think through whether we should keep updating the `go` line, or add a `toolchain` line and update that, keeping the `go` line at the `.0` patch.

PR #1311, opened 2024-04-08
docs: add Go Version section

crozzy:
Document the recently agreed policy regarding go versions.

### DIFF
--- a/docs/contributor/go_version.md
+++ b/docs/contributor/go_version.md
@@ -1,0 +1,4 @@
+# Go Version
+
+Claircore endeavors to declare a `go` version in the go.mod directive that is the minimum required for all the language features to work as expected.
+The `go` version will also be updated if there is deemed to be a security fix that could directly affect claircore's codebase.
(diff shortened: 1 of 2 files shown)


docs/contributor/go_version.md
```diff
@@ -0,0 +1,3 @@
+# Go Version
+
+Claircore endeavors to declare a `go` version in the go.mod directive that is the minimum required for all the language features to work as expected. The `go` version will also be updated if there is deemed to be a security fix that could directly affect claircore's codebase.
```
hdonnay, line 3:
```suggestion
Claircore endeavors to declare a `go` version in the go.mod directive that is the minimum required for all the language features to work as expected.
The `go` version will also be updated if there is deemed to be a security fix that could directly affect claircore's codebase.
```

I've been trying to get the markdown into one sentence (or overly long clause of mine) per line, so the diffs are easier when we inevitably have to update them.

hdonnay (approved):
I'd like to see my suggested change, but it's not important enough to be pedantic about.

docs/contributor/go_version.md
crozzy: updated will just merge when CI goes green

hdonnay (approved):

PR #1315, opened 2024-04-11
datastore: Add vuln & enrich stream updates

jvdm:
Add iterator-based vulnerability/enrichment writes to the existing DB APIs.

### DIFF
--- a/datastore/enrichment.go
+++ b/datastore/enrichment.go
@@ -8,6 +8,9 @@ import (
 	"github.com/quay/claircore/libvuln/driver"
 )
 
+// EnrichmentIter is an [Iter] of enrichment records.
+type EnrichmentIter Iter[*driver.EnrichmentRecord]
+
 // EnrichmentUpdater is an interface exporting the necessary methods
 // for storing and querying Enrichments.
 type EnrichmentUpdater interface {
@@ -15,6 +18,9 @@ type EnrichmentUpdater interface {
 	// EnrichmentRecord(s), and ensures enrichments from previous updates are not
 	// queries by clients.
 	UpdateEnrichments(ctx context.Context, kind string, fingerprint driver.Fingerprint, enrichments []driver.EnrichmentRecord) (uuid.UUID, error)
+	// UpdateEnrichmentsIter performs the same operation as UpdateEnrichments, but
+	// accepting an iterator function.
+	UpdateEnrichmentsIter(ctx context.Context, kind string, fingerprint driver.Fingerprint, enIter EnrichmentIter) (uuid.UUID, error)
 }
 
 // Enrichment is an interface for querying enrichments from the store.
--- a/datastore/matcher_store.go
+++ b/datastore/matcher_store.go
@@ -1,5 +1,11 @@
 package datastore
 
+// Iter is an iterator function that accepts a callback 'yield' to handle each
+// iterator item. The consumer can signal the iterator to break or retry by
+// returning an error. The iterator itself returns an error if the iteration
+// cannot continue or was interrupted unexpectedly.
+type Iter[T any] func(yield func(T, error) bool)
+
 // MatcherStore aggregates all interface types
 type MatcherStore interface {
 	Updater
--- a/datastore/postgres/enrichment.go
+++ b/datastore/postgres/enrichment.go
@@ -17,6 +17,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/quay/zlog"
 
+	"github.com/quay/claircore/datastore"
 	"github.com/quay/claircore/libvuln/driver"
 	"github.com/quay/claircore/pkg/microbatch"
 )
@@ -60,10 +61,27 @@ var (
 	)
 )
 
+func (s *MatcherStore) UpdateEnrichmentsIter(ctx context.Context, updater string, fp driver.Fingerprint, it datastore.EnrichmentIter) (uuid.UUID, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "datastore/postgres/MatcherStore.UpdateEnrichmentsIter")
+	return s.updateEnrichments(ctx, updater, fp, it)
+}
+
 // UpdateEnrichments creates a new UpdateOperation, inserts the provided
 // EnrichmentRecord(s), and ensures enrichments from previous updates are not
 // queried by clients.
-func (s *MatcherStore) UpdateEnrichments(ctx context.Context, name string, fp driver.Fingerprint, es []driver.EnrichmentRecord) (uuid.UUID, error) {
+func (s *MatcherStore) UpdateEnrichments(ctx context.Context, updater string, fp driver.Fingerprint, es []driver.EnrichmentRecord) (uuid.UUID, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "datastore/postgres/MatcherStore.UpdateEnrichments")
+	enIter := func(yield func(record *driver.EnrichmentRecord, err error) bool) {
+		for i := range es {
+			if !yield(&es[i], nil) {
+				break
+			}
+		}
+	}
+	return s.updateEnrichments(ctx, updater, fp, enIter)
+}
+
+func (s *MatcherStore) updateEnrichments(ctx context.Context, name string, fp driver.Fingerprint, it datastore.EnrichmentIter) (uuid.UUID, error) {
 	const (
 		create = `
 INSERT
@@ -134,17 +152,29 @@ DO
 
 	batch := microbatch.NewInsert(tx, 2000, time.Minute)
 	start = time.Now()
-	for i := range es {
-		hashKind, hash := hashEnrichment(&es[i])
-		err := batch.Queue(ctx, insert,
-			hashKind, hash, name, es[i].Tags, es[i].Enrichment,
+	enCt := 0
+	it(func(en *driver.EnrichmentRecord, iterErr error) bool {
+		if iterErr != nil {
+			err = iterErr
+			return false
+		}
+		enCt++
+		hashKind, hash := hashEnrichment(en)
+		err = batch.Queue(ctx, insert,
+			hashKind, hash, name, en.Tags, en.Enrichment,
 		)
 		if err != nil {
-			return uuid.Nil, fmt.Errorf("failed to queue enrichment: %w", err)
+			err = fmt.Errorf("failed to queue enrichment: %w", err)
+			return false
 		}
 		if err := batch.Queue(ctx, assoc, hashKind, hash, name, id); err != nil {
-			return uuid.Nil, fmt.Errorf("failed to queue association: %w", err)
+			err = fmt.Errorf("failed to queue association: %w", err)
+			return false
 		}
+		return true
+	})
+	if err != nil {
+		return uuid.Nil, fmt.Errorf("iterating on enrichments: %w", err)
 	}
 	if err := batch.Done(ctx); err != nil {
 		return uuid.Nil, fmt.Errorf("failed to finish batch enrichment insert: %w", err)
@@ -160,7 +190,7 @@ DO
 	}
 	zlog.Debug(ctx).
 		Stringer("ref", ref).
-		Int("inserted", len(es)).
+		Int("inserted", enCt).
 		Msg("update_operation committed")
 	return ref, nil
 }
--- a/datastore/postgres/updatevulnerabilities.go
+++ b/datastore/postgres/updatevulnerabilities.go
@@ -15,6 +15,7 @@ import (
 	"github.com/quay/zlog"
 
 	"github.com/quay/claircore"
+	"github.com/quay/claircore/datastore"
 	"github.com/quay/claircore/libvuln/driver"
 	"github.com/quay/claircore/pkg/microbatch"
 )
@@ -45,14 +46,27 @@ var (
 	)
 )
 
+// UpdateVulnerabilitiesIter implements vulnstore.Updater.
+func (s *MatcherStore) UpdateVulnerabilitiesIter(ctx context.Context, updater string, fp driver.Fingerprint, it datastore.VulnerabilityIter) (uuid.UUID, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "datastore/postgres/MatcherStore.UpdateVulnerabilitiesIter")
+	return s.updateVulnerabilities(ctx, updater, fp, it, nil)
+}
+
 // UpdateVulnerabilities implements vulnstore.Updater.
 //
 // It creates a new UpdateOperation for this update call, inserts the
 // provided vulnerabilities and computes a diff comprising the removed
 // and added vulnerabilities for this UpdateOperation.
-func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string, fingerprint driver.Fingerprint, vulns []*claircore.Vulnerability) (uuid.UUID, error) {
+func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string, fp driver.Fingerprint, vulns []*claircore.Vulnerability) (uuid.UUID, error) {
 	ctx = zlog.ContextWithValues(ctx, "component", "datastore/postgres/MatcherStore.UpdateVulnerabilities")
-	return s.updateVulnerabilities(ctx, updater, fingerprint, vulns, nil, false)
+	iterVulns := func(yield func(*claircore.Vulnerability, error) bool) {
+		for i := range vulns {
+			if !yield(vulns[i], nil) {
+				break
+			}
+		}
+	}
+	return s.updateVulnerabilities(ctx, updater, fp, iterVulns, nil)
 }
 
 // DeltaUpdateVulnerabilities implements vulnstore.Updater.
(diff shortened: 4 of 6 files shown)


datastore/postgres/enrichment.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/quay/claircore/datastore"
 	"github.com/quay/zlog"
 
 	"github.com/quay/claircore/libvuln/driver"
```
hdonnay, line 21: Fix grouping.

datastore/postgres/updatevulnerabilities.go
```diff
@@ -12,11 +12,11 @@ import (
...
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"github.com/quay/zlog"
-
 	"github.com/quay/claircore"
+	"github.com/quay/claircore/datastore"
 	"github.com/quay/claircore/libvuln/driver"
 	"github.com/quay/claircore/pkg/microbatch"
+	"github.com/quay/zlog"
 )
```
hdonnay, line 20: Fix grouping.

datastore/enrichment.go
```diff
@@ -8,13 +8,22 @@ import (
 	"github.com/quay/claircore/libvuln/driver"
 )
 
+// EnrichmentIter is a function for iterating on enrichment records.
+//
+// It accepts a callback function 'yield' to handle each enrichment record. If the callback
+// returns an error, the iteration is halted, and the error is propagated to indicate cancellation.
+type EnrichmentIter func(yield func(enricher *driver.EnrichmentRecord) error) error
+
```
hdonnay, line 16:
```suggestion
// Iter ...
type Iter[T any] func(yield func(T) bool) error

// EnrichmentIter is a function for iterating on enrichment records.
//
// It accepts a callback function 'yield' to handle each enrichment record. If the callback
// returns an error, the iteration is halted, and the error is propagated to indicate cancellation.
type EnrichmentIter Iter[*driver.EnrichmentRecord]

```

I think specifying a generic type and then additional type definitions would help with having long-ass type definitions.

I'm waffling on changing up the "standard" signature, where the iterator function has no return (meaning the iterator function needs to close over an `error`).

datastore/postgres/updatevulnerabilities.go
```diff
@@ -45,14 +45,33 @@ var (
 	)
 )
 
+type deltaOpts struct {
+	deletedVulns []string
+	vulns        []*claircore.Vulnerability
+}
```
hdonnay, line 51: I dislike "options structs" as a style issue. I think a `deleted Iter[string]` would be fine in the "internal" signature. I can harass @crozzy into helping bash that into shape -- just pull the "delta" changes into a commit that we can mess with.

libvuln/jsonblob/jsonblob.go
hdonnay, line 1: I'd like to see the `jsonblob` changes split out into a separate commit.

libvuln/jsonblob/jsonblob.go
```diff
@@ -455,6 +530,16 @@ func (s *Store) DeltaUpdateVulnerabilities(ctx context.Context, updater string,
 	return uuid.Nil, nil
 }
 
+// UpdateEnrichmentsIter is unimplemented.
+func (s *Store) UpdateEnrichmentsIter(_ context.Context, _ string, _ driver.Fingerprint, _ datastore.EnrichmentIter) (uuid.UUID, error) {
+	return uuid.Nil, nil
```
hdonnay, line 535: Have this return something wrapping `errors.ErrUnsupported`.

libvuln/jsonblob/jsonblob.go
```diff
@@ -455,6 +530,16 @@ func (s *Store) DeltaUpdateVulnerabilities(ctx context.Context, updater string,
...
+}
+
+// UpdateVulnerabilitiesIter is unimplemented.
+func (s *Store) UpdateVulnerabilitiesIter(_ context.Context, _ string, _ driver.Fingerprint, _ datastore.VulnIter) (uuid.UUID, error) {
+	return uuid.Nil, nil
```
hdonnay, line 540: Have this return something wrapping `errors.ErrUnsupported`.

hdonnay (changes requested):
See comments on specific bits. I'd also prefer commits for `Enrichment` and `Vulnerability` changes individually.

datastore/enrichment.go
jvdm: @hdonnay:

> I'm waffling on changing up the "standard" signature, where the iterator function has no return (meaning the iterator function needs to close over an error).

Returning an error from the `yield` closure simplifies error handling for the iterator consumer. The closure doesn't have to set outer scope variables to stop the iteration. Secondly, but irrelevant to this change, it could allow the iterator to handle the error within a single iteration (like ignoring or retrying, for example).

Returning an error in the iterator is desirable because it would be complicated to communicate to the iterator consumer that the iteration stopped successfully or due to an error (and actions are necessary, such as rolling back).

```
var err error
vulnIter(func(vuln *claircore.Vulnerability) bool {
	err = writeVuln(vuln)
	return err != nil
}
if err != nil {
	writeVulnRollback()
	return fmt.Errorf("failed to write vuln: %w", err)
}

// Cannot tell if we need to rollback if the iteration stopped due
// an error on the iteration steps, although we are perfectly healthy
// up until now.
```

jvdm: With errors:

```
err := vulnIter(func(vuln *claircore.Vulnerability) bool {
	return writeVuln(vuln)
}
if err != nil {
	// This can be due to both our closure's failure or the iterator's failure to iterate fully.
	writeVulnRollback()
	return fmt.Errorf("failed to write vuln: %w", err)
}

File: datastore/postgres/updatevulnerabilities.go (hunk 4 above)
Comment (jvdm) [reply datastore/postgres/updatevulnerabilities.go:51]: @hdonnay:

> I dislike "options structs" as a style issue. 
What are the pain points, in your opinion?

> I think a `deleted Iter[string]` would be fine in the "internal" signature.
The internal method needs all vulns upfront when `delete=true`. It makes things a little less straightforward to change if I have to iterate over the vulns when `delete=true` to handle the deletion steps first (they come first in the update logic) to decide then not to iterate if `delete=false` and iterate if `delete=false`.

The proposed patch seems the most straightforward way to achieve what we want; Considering we are sunsetting this interface sometime in the future.

File: datastore/enrichment.go (hunk 3 above)
Comment (jvdm) [reply datastore/enrichment.go:16]: > I think specifying a generic type and then additional type definitions would help with having long-ass type definitions.

+1

File: datastore/postgres/updatevulnerabilities.go (hunk 4 above)
Comment (jvdm) [reply datastore/postgres/updatevulnerabilities.go:51]: Ah, missed the fact that the iterator for this particular call can be called twice. For the internal API I think this assumption is fair. I am going to update this.

File: datastore/enrichment.go [hunk 7]
```diff
@@ -4,17 +4,22 @@ import (
 	"context"
 
 	"github.com/google/uuid"
-
```
hdonnay, line 7: Whitespace pedantry: this line should be here.

datastore/matcher_store.go
```diff
@@ -1,5 +1,11 @@
 package datastore
 
+// Iter is an iterator function that accepts a callback 'yield' to handle each
+// iterator item. The consumer can signal the iterator to break or retry by
+// returning an error. The iterator itself returns an error if the iteration
+// cannot continue or was interrupted unexpectedly.
+type Iter[T any] func(yield func(T) error) error
+
```
hdonnay, line 8: I'd like to keep this aligned to the `rangefunc` experiment syntax (which looks likely to land in Go 1.23), so I'd rather we remove the error returns:
```suggestion
// Iter is an iterator function that accepts a callback 'yield' to handle each
// iterator item. The consumer can signal the iterator to break or retry by
// returning an error. The iterator itself returns an error if the iteration
// cannot continue or was interrupted unexpectedly.
type Iter[T any] func(yield func(T, error) bool)

```

Alternatively, if you feel strongly about always communicating the error, we could expand this a little:
```suggestion
type Iter[T any] func(yield func(T, error) bool)

type IntoIter[T any] func(/*ctx?*/) (Iter[T], func() error)
```
(This idea stolen from [here](https://github.com/golang/go/issues/65236#issuecomment-1953399166).)
And then the using code needs to call the second function (and the compiler should complain if that doesn't happen) and the implementer just needs to keep an `error` around to return for the second function to return.

jvdm: Cool, thanks for the context. Yielding errors suffice. I've updater the `Iter` interface. Here's an example of what users could do using the `jsonblob.Iter`:

```go
ref, err = u.store.UpdateVulnerabilitiesIter(ctx, it.Updater, it.Fingerprint, func(yield func(*claircore.Vulnerability, error) bool) {
	for it.Next() {
		if !yield(it.Vulnerability, nil) {
			return
		}
		count++
	}
	if loader.Err() != nil {
		_ = yield(nil, loader.Err())
	}
})
```

datastore/postgres/enrichment.go
hdonnay, line 1: :+1:

hdonnay (changes requested):
Datastore changes look good, especially once the restriction on the iterator needing to restart is documented.

The jsonblob changes might make more sense if implemented on a new type, instead of having already tricksy code now have two access patterns intertwined.

datastore/updater.go
```diff
@@ -19,6 +22,9 @@ type Updater interface {
 	// vulnerabilities, and ensures vulnerabilities from previous updates are
 	// not queried by clients.
 	UpdateVulnerabilities(ctx context.Context, updater string, fingerprint driver.Fingerprint, vulns []*claircore.Vulnerability) (uuid.UUID, error)
+	// UpdateVulnerabilitiesIter performs the same operation as
+	// UpdateVulnerabilities, but accepting an iterator function.
+	UpdateVulnerabilitiesIter(ctx context.Context, updater string, fingerprint driver.Fingerprint, vulnIter VulnerabilityIter) (uuid.UUID, error)
```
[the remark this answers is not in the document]
jvdm: > 	// The passed iterator must be able to iterate over the sequence multiple times.

That's not true for `UpdateVulnerabilitiesIter`. It would be for a potential `DeltaUpdateVulnerabilitiesIter`, but this patch does not aim to support that.

libvuln/jsonblob/jsonblob.go
```diff
@@ -117,6 +144,54 @@ func (l *Loader) Entry() *Entry {
 	return l.e
 }
 
+// Iter returns an iterator for read all objects for the current update operation.
+func (l *Loader) Iter() *iter {
```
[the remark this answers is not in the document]
jvdm: > This shouldn't return an unexported type.

My goal is to prevent code from outside the package to create `iter` objects. I need clarification of why this should not be an unexported type.

datastore/postgres/updatevulnerabilities.go
```diff
@@ -68,10 +82,24 @@ func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string
...
-	return s.updateVulnerabilities(ctx, updater, fingerprint, vulns, deletedVulns, true)
+	iterVulns := func(yield func(*claircore.Vulnerability, error) bool) {
+		for i := range vulns {
+			if !yield(vulns[i], nil) {
+				break
+			}
+		}
+	}
+	delVulns := func(yield func(string, error) bool) {
+		for _, s := range deletedVulns {
```
RTann, line 93: other implementations above used the index `i` instead. Curious why the change here?

datastore/postgres/updatevulnerabilities.go
```diff
@@ -68,10 +82,24 @@ func (s *MatcherStore) UpdateVulnerabilities(ctx context.Context, updater string
...
+			if !yield(s, nil) {
+				break
+			}
+		}
+	}
+	return s.updateVulnerabilities(ctx, updater, fingerprint, iterVulns, delVulns)
 }
 
-func (s *MatcherStore) updateVulnerabilities(ctx context.Context, updater string, fingerprint driver.Fingerprint, vulns []*claircore.Vulnerability, deletedVulns []string, delta bool) (uuid.UUID, error) {
+func (s *MatcherStore) updateVulnerabilities(ctx context.Context, updater string, fingerprint driver.Fingerprint, vulnIter datastore.VulnerabilityIter, delIter datastore.Iter[string]) (uuid.UUID, error) {
```
RTann, line 102: `delIter` is a bit confusing to me because it makes me think it's short for `delta` instead of `delete`. Can we make it more explicit and call it `deleteIter`?

libvuln/jsonblob/jsonblob.go
```diff
@@ -45,70 +45,97 @@ type Store struct {
...
 }
 
-// Next reports whether there's an [Entry] to be processed.
-func (l *Loader) Next() bool {
+// iter wraps a disk entry with methods to read the next object for the same
+// update operation.
+//
+// Users are expected to call [*iter.Next()] to read and parse the next object,
+// which can be retrieved using [iter.Vulnerability] or [iter.Enrichment] based
+// on the update kind.  Errors are reported in the [Loader.Err].
```
RTann, line 81: nit: I think other comments use a single space after a period

libvuln/jsonblob/jsonblob.go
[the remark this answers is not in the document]
jvdm: > Alternatively, maybe just adding a StreamingLoader or IterLoader with a better/different interface instead of intertwining this with Loader would prevent the possibility of misuse.

That's a valid point.

Considering the `Loader` already uses the `Next()/Entry()/Err()` pattern, and considering we don't want to break backward compatibility, it seems less effort to continue using the `Loader` internals vs. a new type. We would want the existing `Loader` to use the `StreamingLoader` anyway (if not, it's one additional code path for this, and I would rather avoid).

So we have the current proposal, where the `Loader` supports two iteration patterns:

1. Calling `Next()` + `Entry()` to retrieve batches per updater.
2. Calling `NextIter()` + `Iter()` to retrieve individual items per updater.

Or your proposal:

> It seems like it should be returning the equivalent of iter.Seq2[Entry, err], documenting that the returned Entry objects contain a single Enrichment or Vulnerability, and calling (*Loader).Iter() is mutually exclusive with calling (*Loader).Entry().

Let's ignore `rangefunc` for now -- which I don't think it worth doing anyway.

That option implies we would reuse `Next()` for both (1.) and (2.) in the previous proposal. It lets the user call `Next()` once and decide if either return the whole batch with `Entry()` or an iterator using `Iter()`. But, that's not possible unless I change the contract of `Entry()` dictating the data is already available when `Next() == true`. I would have to either return `nil` (which I don't think the current contract allows) or partial data (which could result in misuse before the caller checks things failed calling `Loader.Err()`).

A third option is to pass an option to the `Loader`, which would change the behavior of `Next()` and `Entry()` to be of `NextIter() ` and `Iter()` (we would reuse `Entry` to iterate over individual items instead of populating the slices). That's convoluted.

So, the two iterator patterns (the current approach) balance out the need for functionality, minimal changes, and test coverage. I want to maintain backward compatibility, avoid regressions, and add a new access pattern, at the expense of adding two access patterns to the `Loader` API. But I think that's OK because that API will die soon.

datastore/postgres/updatevulnerabilities.go
```diff
@@ -139,6 +167,7 @@ func (s *MatcherStore) updateVulnerabilities(ctx context.Context, updater string
 		return uuid.Nil, fmt.Errorf("failed to create update_operation: %w", err)
 	}
 
+	delta := delIter != nil
```
RTann, line 170: is this used?

datastore/postgres/updatevulnerabilities.go
```diff
@@ -181,18 +210,20 @@ func (s *MatcherStore) updateVulnerabilities(ctx context.Context, updater string
 		}
 
 		if len(oldVulns) > 0 {
-			for _, v := range vulns {
+			vulnIter(func(v *claircore.Vulnerability, _ error) bool {
```
RTann, line 213: should this and `delIter` make assumptions about the passed in `err` that it's safe to ignore?

libvuln/jsonblob/jsonblob.go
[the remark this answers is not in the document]
jvdm: Quoting myself:

> Let's ignore rangefunc for now -- which I don't think is worth doing anyway.

I lied. I played a bit with the idea today. The implementation turned out simpler, albeit a bit unusual, because I went ahead and used nested iterators. @hdonnay, let me know if you would prefer to pursue that instead https://github.com/quay/claircore/commit/d2531cb8f1fe63af5b2479563e9899f303c61aba

By the way, we would have to keep the previous `Next()/Entry()/Err()` implementation as is for now, so there would be two paths for the same thing.

hdonnay (approved):
We talked a little on Slack and decided split out the "provider" part of `jsonblob`.

All this LGTM.

PR #1320, opened 2024-04-12
rpm: implement an RPM filescanner to discern RPM filepaths

crozzy:
Using the filepaths discovered by the RPM filescanner we can judge whether or not a language package has been installed via RPM or not.


file.go
```diff
@@ -5,6 +5,7 @@ type FileKind string
 
 const (
 	FileKindWhiteout = FileKind("whiteout")
+	FileKindRPM      = FileKind("rpm")
```
hdonnay, line 8: So every file's "kind" is effectively the component that noticed it?

crozzy: I suppose in theory any filescanner can create any `FileKind` and any `Resolver` can be interested in any `FileType` or perhaps multiple in the future, so the name is kind of arbitrary. Do you think  the kind could be better named?

crozzy:
I think I'll probably add some more test cases to the resolver but it'd be good to get some eyes on again.

rhel/resolver.go
```diff
@@ -0,0 +1,60 @@
...
+		isRemovable := false
+		for _, fs := range ir.Files {
+			for _, f := range fs {
+				if f.Kind == claircore.FileKindRPM && !isRPMPackage && pkg.Filepath == f.Path {
+					isRemovable = true
+					zlog.Debug(ctx).
+						Str("package name", pkg.Name).
+						Str("package file", pkg.Filepath).
+						Str("rpm file", f.Path).
+						Msg("package determined to have come from RPM, deleting")
```
crozzy, line 48: Should break here to the next package

rhel/resolver.go
```diff
@@ -0,0 +1,60 @@
...
+				}
+			}
+		}
+		if !isRemovable {
```
crozzy, line 52: Could just delete?

file.go
hdonnay: I guess not? It's fine.

rhel/resolver.go
hdonnay, line 1: Why is this a `rhel` component? Shouldn't it work for any `rpm` system?

crozzy: As of now the way it identifies it as an RPM is via the package's `repositoryKey` which is RHEL specific. If it was generic for all RPMs we'd need some generic way to identify RPM packages I suppose. I believe this was the initial intent of the location

hdonnay: I suppose that makes sense.

hdonnay (dismissed):
SGTM

rpm/native_db.go
```diff
@@ -23,6 +28,128 @@ type nativeDB interface {
...
+type ObjectResponse interface {
+	[]*claircore.Package | []claircore.File
+}
+
+// getDBObjects does all the dirty work of extracting generic claircore objects
+// from an RPM database. Provide it with a foundDB, the sys and a fn extract function
+// it will create a implementation agnostic nativeDB and extract specific claircore
+// objects from it.
+func getDBObjects[T ObjectResponse](ctx context.Context, sys fs.FS, db foundDB, fn func(context.Context, string, nativeDB) (T, error)) (T, error) {
+	var nat nativeDB // see native_db.go:/nativeDB
```
RTann, line 42: This *is* native_db.go :)

rpm/packagescanner.go
RTann, line 1: Does the version need to be bumped since the IndexReport will change?

rpm/native_db.go
```diff
@@ -299,3 +426,106 @@ func constructHint(b *strings.Builder, info *Info) string {
...
+	return func(p string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			return nil
+		}
+
+		dir, n := path.Split(p)
+		dir = path.Clean(dir)
```
RTann, line 440: I know it's copied over, but why is it `path` instead of `filepath`? Guessing it's inside the TAR?

rpm/native_db.go
```diff
@@ -299,3 +426,106 @@ func constructHint(b *strings.Builder, info *Info) string {
...
+			ok := bdb.CheckMagic(ctx, f)
+			f.Close()
+			if !ok {
+				return nil
+			}
+			*out = append(*out, foundDB{
+				Path: dir,
+				Kind: kindBDB,
+			})
+		case `rpmdb.sqlite`:
```
RTann, line 456: I know it's copied, but the others check using a magic number. We don't bother with that here? I guess trusting the file extension `sqlite` is good enough

rpm/filescanner.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		done[db.Path] = struct{}{}
+		fs, err := getDBObjects(ctx, sys, db, filesFromDB)
+		if err != nil {
+			return nil, fmt.Errorf("rpm: error getting native DBs: %w", err)
+		}
+		files = append(files, fs...)
+	}
+
+	zlog.Debug(ctx).Int("count", len(found)).Msg("found possible databases")
```
RTann, line 87: this may make more sense above the loop

linux/coalescer.go
```diff
@@ -58,6 +58,12 @@ func (c *Coalescer) Coalesce(ctx context.Context, layerArtifacts []*indexer.Laye
 		for db, pkgs := range tmp {
 			dbs[db] = pkgs
 		}
+		for _, f := range artifacts.Files {
+			if c.ir.Files == nil {
+				c.ir.Files = make(map[string][]claircore.File)
+			}
```
RTann, line 64: another option is to check if `len(artifacts.Files) > 0` before the loop and initialize there so we don't have to check `c.ir.Files == nil` each iteration

rhel/resolver.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Resolve takes a claircore.IndexReport and uses the rpm Files
+// to determine if the package originate from an RPM. If the package
+// was deemed to have been installed via RPM it isn't included in the
+// final report.
+func (r *Resolver) Resolve(ctx context.Context, ir *claircore.IndexReport, layers []*claircore.Layer) *claircore.IndexReport {
+	for pkgID, pkg := range ir.Packages {
+		isRPMPackage := false
+	envLoop:
+		for i := 0; i < len(ir.Environments[pkgID]); i++ {
```
RTann, line 26: I think this can just be a range loop

rhel/resolver.go
```diff
@@ -0,0 +1,56 @@
...
+			if ir.Environments[pkgID][i].RepositoryIDs != nil {
+				for _, rID := range ir.Environments[pkgID][i].RepositoryIDs {
+					r := ir.Repositories[rID]
+					if r.Key == repositoryKey {
```
RTann, line 30: Does this happen? I was under the impression if we currently see a JAR file, wouldn't the only repo be the maven one?

crozzy: This block is checking if the package is an RHEL RPM package, and negating any exclusion if it is. In reality this is a "belt and braces" check to make sure that no rpm packages are discounted from the final index report but I don't think we set a Filepath when we index these packages at the moment (but could in the future)

rpm/native_db.go
crozzy: I don't know about this, I feel like I've heard a reason when we were debugging prod errors but can't recall. @hdonnay any ideas?

rpm/native_db.go
crozzy: Yeah I suppose so too, maybe related to https://github.com/quay/claircore/issues/1327

rpm/native_db.go
hdonnay: Yeah, this just hasn't come up as an issue, whereas relying on file names for the others has.

rpm/native_db.go
hdonnay: It's path because layer paths should always use normal slashes, at least to my reading of the [spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13_06). In practice, Windows layers also seem to use normal slashes.

linux/coalescer.go
```diff
@@ -58,6 +58,13 @@ func (c *Coalescer) Coalesce(ctx context.Context, layerArtifacts []*indexer.Laye
 		for db, pkgs := range tmp {
 			dbs[db] = pkgs
 		}
+
+		if len(artifacts.Files) > 0 {
+			c.ir.Files = make(map[string][]claircore.File)
+		}
```
RTann, line 64: I just realized, what happens if two layers have `len(artifacts.Files) > 0`? Seems like we may overwrite `c.ir.Files`. I lead you astray my b

RTann: Also, RHEL has its own coalescer. Did you mean to update that one?

linux/coalescer.go
```diff
@@ -58,6 +58,13 @@ func (c *Coalescer) Coalesce(ctx context.Context, layerArtifacts []*indexer.Laye
...
+		for _, f := range artifacts.Files {
+			c.ir.Files[artifacts.Hash.String()] = append(c.ir.Files[artifacts.Hash.String()], f)
```
RTann, line 66: nit: may be nice to set `artifacts.Hash.String()` to a variable, so we don't call it so many times

rpm/filescanner.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+)
+
+const (
+	scannerName    = "rpm"
+	scannerVersion = "1"
```
RTann, line 17: would updating this (or even just introducing it) trigger reindexing?

rhel/resolver.go
RTann: Oh ok got it. I misremembered how the `filesFromDB` function worked

linux/coalescer.go
crozzy: It needed to be updated as the `IndexReport` was updated.

rpm/filescanner.go
crozzy: It should cause the index_state to change, which an external application can use to trigger re-indexing

rpm/filescanner.go
```diff
@@ -0,0 +1,89 @@
...
+// Version implements [indexer.VersionedScanner].
+func (*FileScanner) Version() string { return scannerVersion }
+
+// Kind implements [indexer.VersionedScanner]
+func (*FileScanner) Kind() string { return scannerKind }
+
+// Scan reports any found Files that were installed via RPMs in the
+// layer.
+//
+// It's an expected outcome to return (nil, nil) no RPM packages are found in the Layer.
```
RTann, line 42:
```suggestion
// It's an expected outcome to return (nil, nil) when no RPM packages are found in the Layer.
```

rpm/filescanner.go
```diff
@@ -0,0 +1,89 @@
...
+func (s *FileScanner) Scan(ctx context.Context, layer *claircore.Layer) ([]claircore.File, error) {
+	if err := ctx.Err(); err != nil {
```
RTann, line 44: just curious, how come this was added?

rpm/filescanner.go
```diff
@@ -0,0 +1,89 @@
...
+		"layer", layer.Hash.String())
+	zlog.Debug(ctx).Msg("start")
+	defer zlog.Debug(ctx).Msg("done")
+
+	sys, err := layer.FS()
+	if err != nil {
+		return nil, fmt.Errorf("rpm: unable to open layer: %w", err)
+	}
+
+	found := make([]foundDB, 0)
```
RTann, line 61: I feel like other parts of the library just did `found := []foundDB{}` in this situation. I won't block the PR over this, though

linux/coalescer.go
RTann: Don't we still need to update the RHEL coalescer since this'll affect RHEL images?

rhel/resolver.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+var (
+	_ indexer.Resolver = (*Resolver)(nil)
+)
+
+type Resolver struct{}
+
+// Resolve takes a claircore.IndexReport and uses the rpm Files
+// to determine if the package originate from an RPM. If the package
```
RTann, line 19:
```suggestion
// to determine if the package originated from an RPM. If the package
```

rhel/resolver.go
```diff
@@ -0,0 +1,56 @@
...
+// was deemed to have been installed via RPM it isn't included in the
```
RTann, line 20:
```suggestion
// was deemed to have been installed via RPM, it isn't included in the
```

rhel/resolver.go
```diff
@@ -0,0 +1,56 @@
...
+						break envLoop
+					}
+				}
+			}
+		}
+		if !isRPMPackage {
+		filesLoop:
+			for _, fs := range ir.Files {
+				for _, f := range fs {
+					if f.Kind == claircore.FileKindRPM && pkg.Filepath == f.Path {
```
RTann, line 41: nit:

```suggestion
					if f.Kind == claircore.FileKindRPM && f.Path == pkg.Filepath {
```

RTann (commented):
few tiny things and one potentially important thing. should be good to go after this

rpm/filescanner.go
crozzy: :+1: Going to keep it as-is as it matches the PackageScanner

rpm/filescanner.go
crozzy: To early out of an expensive function if the context is cancelled.

linux/coalescer.go
crozzy: The RHEL coalescer doesn't come into contact with any RPM files, the linux coalescer is responsible for that as it is run as part of the RPM ecosystem (that includes the rpm.FileScanner). i.e. There files we're finding aren't RHEL specific but RPM specific.

rhel/resolver.go
```diff
@@ -0,0 +1,56 @@
...
+// Resolve takes a claircore.IndexReport and uses the rpm Files
+// to determine if the package originated from an RPM. If the package
+// was deemed to have been installed via RPM, it isn't included in the
+// final report.
+func (r *Resolver) Resolve(ctx context.Context, ir *claircore.IndexReport, layers []*claircore.Layer) *claircore.IndexReport {
+	for pkgID, pkg := range ir.Packages {
+		isRPMPackage := false
+	envLoop:
+		for _, env := range ir.Environments[pkgID] {
+			if env.RepositoryIDs != nil {
```
RTann, line 27: nit: not needed. The loop would just skipped. Is it possible `env` is `nil` and that should be checked?

RTann (approved):
Just one nit I noticed, otherwise LGTM

rhel/resolver.go
crozzy: Not sure I follow, the `envLoop` would be broken but we'd still get to the files loop with the contextual information that there's nothing to do so go around for another loop

RTann: I don't think there is a difference between 

```
if env.RepositoryIDs != nil {
	for _, rID := range env.RepositoryIDs {
		...
	}
}
```

and just

```
for _, rID := range env.RepositoryIDs {
	...
}
```

since the loop would not happen if `env.RepositoryIDs == nil`

crozzy: Got it, thanks, wasn't reading them as separate comments

RTann: sorry I wasn't clear. I was referring to the `env.RepositoryIDs != nil` check prior to the loop but I was also wondering if the intention of that check was indeed for `env.RepositoryIDs` instead of `env` since the loop would handle the case where `env.RepositoryIDs == nil`

RTann (approved):

PR #1323, opened 2024-04-16
urn: add URN parser and helpers

hdonnay:
This sets up some support infrastructure to unify naming across claircore. The goal is to be able to pass around names as strings instead of interface objects.

I'm unsure about the `Name` type -- thoughts on if that should live here and if it should get a method on `URN`?


hdonnay:
Re-drafting this until I figure out what to do about percent-encoding.

PR #1333, opened 2024-05-03
cpe: fix cross-version matching

hdonnay:
@crozzy discovered that the matching implementation behaved differently depending on whether the incoming CPE was unbound from a 2.2 URI or a 2.3 formatted string.


crozzy (approved):
LGTM

PR #1345, opened 2024-05-21
libvuln: implement DeltaUpdateVulnerabilities for jsonblob store

crozzy:
This change ensures that any updaters that implement the delta update flow also have their vulnerabilities persisted in the jsonblob store.


hdonnay (approved):
I think this makes sense to me -- the alternative would be returning `errors.ErrUnsupported` and making lots of changes to the update manager.

PR #1346, opened 2024-05-22
rhel: Add RHCOS content set location to repositoryscanner

Maddosaurus:
RHCOS stores the content-set files in /ur/share/buildinfo instead of /root/buildinfo.
As RHCOS and RHEL share CVE feeds, we can extend this scanner instead of building a new one for RHCOS.


crozzy (approved):
LGTM

crozzy (changes requested):
Oh, forgot. We should iterate the `RepositoryScanner.Version` to signal to clients (who care) that something has changed.

Issue #1327, opened 2024-04-24
tarfs breaks on Windows due to using `filepath` instead of `path`


rohansingh:
The [documentation for `io/fs`][1] states:

> Note that paths are slash-separated on all systems, even Windows. Paths containing other characters such as backslash and colon are accepted as valid, but those characters must never be interpreted by an [FS](https://pkg.go.dev/io/fs#FS) implementation as path element separators.

Because `tarfs` uses `filepath`, it ends up creating normalized paths that contain backslashes when on Windows:

https://github.com/quay/claircore/blob/762e6a4409d3f098cc8646cbbbe4815e6903a2d6/pkg/tarfs/tarfs.go#L35-L37

Trying to open a file in the `FS` using a path with forward slashes then fails, with `ErrNotExist`. This is contrary to other `fs.FS` implementations.

---

This is admittedly a bit of a drive-by issue. I'm not a ClairCore user, and I only noticed this when trying to use `pkg/tarfs` in a completely unrelated project. I'm not even sure if Windows support is a goal for this project. However, this is a correctness issue that's probably straightforward to fix by using `path` in all `fs.FS` operations.

Note that the implementation at `github.com/nlepage/go-tarfs` also suffered from the same issue, but it was fixed in nlepage/go-tarfs#7. That may be a useful reference

[1]: https://pkg.go.dev/io/fs#ValidPath

PR #1353, opened 2024-06-07
tarfs: remove `path/filepath` usage

hdonnay:
This was done out of a desire to use `filepath.Rel`, and force of habit, but had the unfortunate side-effect of mangling paths on Windows.

Closes: #1327


crozzy (approved):
LGTM

Issue #1359, opened 2024-07-04
matcher "error: upstream_version is empty" error received when package has no version and passed to debian matcher


bainsy88:
We have seen a debian image where trying to get a report causes it to fail with `error: upstream_version is empty`

Error comes from this library - https://github.com/knqyf263/go-deb-version/blob/master/version.go#L94

Which is called from here https://github.com/quay/claircore/blob/v1.5.26/debian/matcher.go#L48

Having pulled the index report for the image that sees this error we can see the follow invalid package: 

```
  "35191": {
      "id": "35191",
      "name": "curl",
      "version": "",
      "kind": "binary",
      "source": {
        "id": "8288260873",
        "name": "curl",
        "version": "",
        "kind": "source"
      },
      "arch": "x64-windows"
    },
```
The first thing that jumps out is the invalid arch for debian and the lack of version and if you look up the details for the ID you can see

```
"35191": [
     {
       "package_db": "root/.cpanm/work/1717691746.7/Alien-Build-2.83/corpus/vcpkg/r2/installed/vcpkg/status",
       "introduced_in": "sha256:77420ef9819acb73e9da379e3e5586396daaff5b185076ec2a1df1556774a0db",
       "distribution_id": "334152",
       "repository_ids": null
     },
     {
       "package_db": "root/.cpanm/work/1717694129.7/Alien-Build-2.83/corpus/vcpkg/r2/installed/vcpkg/status",
       "introduced_in": "sha256:01d0786122a87f4245f1ddcd66f25fa82708dc5deb59728ed72b6fc53f665e10",
       "distribution_id": "334152",
       "repository_ids": null
     }
```

Researching [Alien-Build](https://metacpan.org/pod/Alien) it looks like a tool for generating Perl CPAN for things that are not perl. It appears that Perl CPANfiles are the same format as a Ruby Gem which is supported by Clair so suspect that is how they are parsed but are not 100% sure. 

Would be interested on thoughts around if there is a more graceful failure mode, could the image be returned as unsupported or maybe there is way to handle this scenario and still give a result?

hdonnay:
It looks like `Alien::Build` has a `vcpkg` database they use in tests. It seems like `vcpkg` used to use a Debian-like "control" format.

- https://github.com/PerlAlien/Alien-Build/blob/96cdc77724c6c0b0be31e36e2ccc5803a1b5c510/corpus/vcpkg/r2/installed/vcpkg/status
[...]

PR #1363, opened 2024-07-08
dpkg: handle vcpkg "control" files

hdonnay:
This should prevent spurious packages appearing because of these files.

Closes: #1359

### DIFF
--- a/dpkg/scanner.go
+++ b/dpkg/scanner.go
@@ -25,7 +25,7 @@ import (
 const (
 	name    = "dpkg"
 	kind    = "package"
-	version = "6"
+	version = "7"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


dpkg/scanner.go
```diff
@@ -239,6 +248,21 @@ Restart:
 		}
 		name := hdr.Get("Package")
 		v := hdr.Get("Version")
+
+		// Is this a valid package?
+		if name == "" || v == "" {
```
crozzy, line 253: So check if the bare-minimum is in the header, if it's not, either return a fatal or non-fatal error depending on if some Header keys are present? Is it possible to mention the significance of these Header keys in a comment?

crozzy (changes requested):
SGTM, just one comment

crozzy (approved):

Issue #1366, opened 2024-07-19
Move limited error response reading to common package for reuse


crozzy:
A couple of places in the codebase use this logic to read a limited number of bytes from a server's error response:

```go
				bodyStart, err := io.ReadAll(io.LimitReader(res.Body, 256))
				if err != nil {
					zlog.Warn(ctx).Err(err).Msg("additional error while reading error response")
				} else {
					zlog.Warn(ctx).Bytes("response", bodyStart).Msg("received error response")
				}
				return fmt.Errorf("unexpected response from advisory URL %q: %s", req.URL, res.Status)
```
We should make this common to allow other parts to reuse w/o copying.

PR #1369, opened 2024-07-23
httputil: add a ResponseChecker to be reused when making http calls

crozzy:
We were repeating this pattern in a few places in the code and it would be a good idea to include it in other http calls to get some context about an error without having to read the whole body.

### DIFF
--- a/internal/httputil/responsechecker.go
+++ b/internal/httputil/responsechecker.go
@@ -1,0 +1,28 @@
+package httputil
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// CheckResponse takes a http.Response and a variadic of ints representing
+// acceptable http status codes. The error returned will attempt to include
+// some content from the server's response.
+func CheckResponse(resp *http.Response, acceptableCodes ...int) error {
+	acceptable := false
+	for _, code := range acceptableCodes {
+		if resp.StatusCode == code {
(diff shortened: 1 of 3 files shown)


crozzy:
closes https://github.com/quay/claircore/issues/1366

internal/httputil/responsechecker.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+	"io"
+	"net/http"
+)
+
+func ResponseChecker(resp *http.Response, acceptableCodes ...int) error {
+	acceptable := false
+	for _, code := range acceptableCodes {
+		if resp.StatusCode == code {
+			acceptable = true
```
hdonnay, line 13:
```suggestion
			acceptable = true
			break
```

internal/httputil/responsechecker.go
```diff
@@ -0,0 +1,24 @@
+package httputil
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+)
+
+func ResponseChecker(resp *http.Response, acceptableCodes ...int) error {
```
hdonnay, line 9: Should have a verb name, like `CheckResponse`. Also needs a doc comment.

hdonnay (changes requested):

hdonnay (approved):

PR #1370, opened 2024-07-24
csaf: add FindThreat() method

crozzy:
We will need to start looking up products from the threat object so add a helper method to make it easier.


hdonnay (approved):
LGTM

PR #1372, opened 2024-07-26
cpe: add additional handling for invalid URI inputs

hdonnay:
Found a case where invalid input resulted in the component being interpreted incorrect during a match. This should fix that bug.

### DIFF
--- a/toolkit/types/cpe/unbind.go
+++ b/toolkit/types/cpe/unbind.go
@@ -64,16 +64,25 @@ func UnbindURI(s string) (WFN, error) {
 		if i == 5 && strings.HasPrefix(c, "~") {
 			attrs := [...]Attribute{Edition, SwEdition, TargetSW, TargetHW, Other}
 			for i, c := range strings.SplitN(c, `~`, 6)[1:] {
-				r.Attr[attrs[i]].unbindURI(&b, c)
+				if err := r.Attr[attrs[i]].unbindURI(&b, c); err != nil {
+					return WFN{}, fmt.Errorf("cpe: %v: %w", attrs[i], err)
+				}
 			}
 			continue
 		}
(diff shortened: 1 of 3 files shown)


hdonnay:
The alternative to this sort of escaping is to report an error. But since we're accepting these as input currently, this might be the least bad solution.

Another fix would be to replace the `strings.Replacer` with some code that explicitly does a normalization pass, percent-encoding things that should be escaped and decoding things that shouldn't.

hdonnay:
After some offline discussion, decided to have these inputs error, instead.

toolkit/types/cpe/unbind.go
```diff
@@ -64,16 +64,24 @@ func UnbindURI(s string) (WFN, error) {
...
+		if err := r.Attr[attrs[i]].unbindURI(&b, c); err != nil {
+			return WFN{}, fmt.Errorf("cpe: %v: %w", attrs[i], err)
+		}
 	}
 	return r, r.Valid()
 }
 
-func (v *Value) unbindURI(b *strings.Builder, s string) {
+func (v *Value) unbindURI(b *strings.Builder, s string) error {
+	// From the characters that should be escaped, Table 6-1.
```
crozzy, line 81: I can't find the Table 6-1 (or [here](https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf) it seems to show something different), would it be possible to add the link as well?

crozzy (changes requested):
Sounds good, just one comment

toolkit/types/cpe/unbind.go
hdonnay: That's the matching specification, not the naming spec. I'll add a link.

crozzy (approved):

PR #1376, opened 2024-08-02
replace deprecated mockgen

janisz:
`github.com/golang/mock` was deprecated and now the mainained fork is `go.uber.org/mock`.

- https://github.com/golang/go/issues/61017
- https://github.com/golang/mock/commit/674a1c4a938e7b453416052327872571656801b7


Makefile
```diff
@@ -9,8 +9,8 @@ clear-cache:
 
 # generates mocks of interfaces for testing
 .PHONY: genmocks
-genmocks:
-	@command -v mockgen >/dev/null || go install github.com/golang/mock/mockgen
+genmocks: go.mod go.sum
+	go install go.uber.org/mock/mockgen
 	go generate ./...
```
hdonnay, line 14:
```suggestion
genmocks:
	go generate -run mockgen ./...
```

I think this is a better way to express this:

1. dependencies on a phony target don't make a lot of sense
1. all the `generate` directives should be using `go run`
1. the `go generate` command should only actually run `mockgen`

hdonnay (changes requested):

janisz:
@hdonnay PTAL

hdonnay (approved):

PR #1384, opened 2024-08-19
vex: check if debug logging enabled before appending

crozzy:
There are a couple of places that append to a slice with debug information, this patch checks if debug logging is enabled before appending to save unnecessary allocation.


hdonnay (approved):
LGTM

PR #1385, opened 2024-08-19
rhel: deprecate RHCC updater in favor of VEX updater

crozzy:
We can extract vulnerability information about containers from the VEX data. This negates the need to look for it in the cvemap.xml file. This change modifies the VEX updater to allow for ingesting vulnerabilities in a way that can be matched my the RHCC matcher.


rhel/vex/parser.go
```diff
@@ -290,8 +271,46 @@ func (c *creator) knownAffectedVulnerabilities(ctx context.Context, v csaf.Vulne
...
 		}
 		vuln.Repo = c.rc.Get(wfn)
+		// It is possible that we will not find a pURL, in that case
+		// the package.Name will be reported as-is.
+		purlHelper, ok := compProd.IdentificationHelper["purl"]
+		if ok {
+			purl, err := packageurl.FromString(purlHelper)
+			if err != nil {
+				zlog.Warn(ctx).
+					Str("purl", purl.String()).
```
RTann, line 281: There are times when `FromString` returns an empty `purl` when there is an error. I think the original message of `purlHelper` was better

rhel/vex/parser.go
```diff
@@ -508,12 +601,91 @@ func createPackageKey(repo, name, fixedIn string) string {
...
+		}
+		return r[1], nil
+	case packageurl.TypeRPM:
+		return purl.Name, nil
+	default:
+		return "", fmt.Errorf("unexpected purl type %s", purl.Type)
+	}
+}
+
+var acceptedTypesMap = map[string]bool{
```
RTann, line 656: nit: maybe just `acceptableTypes`?

rhel/vex/parser.go
```diff
@@ -508,12 +601,91 @@ func createPackageKey(repo, name, fixedIn string) string {
...
+			epoch = e
+		}
+		return epoch + ":" + purl.Version, nil
+	default:
+		return "", fmt.Errorf("unexpected purl type %s", purl.Type)
+	}
+}
+
+// ExtractPackageName deals with 2 pURL types, TypeRPM and TypeOCI
+//   - TypeODI: check if there is Namespace and Name i.e. rhel7/rhel-atomic
```
RTann, line 628:
```suggestion
//   - TypeOCI: check if there is Namespace and Name i.e. rhel7/rhel-atomic
```

rhel/vex/parser.go
```diff
@@ -508,12 +601,91 @@ func createPackageKey(repo, name, fixedIn string) string {
...
+	case packageurl.TypeOCI:
+		if purl.Namespace != "" {
+			return purl.Namespace + "/" + purl.Name, nil
+		}
+		// Try finding an image name from the tag qualifier
+		ru, ok := purl.Qualifiers.Map()["repository_url"]
+		if !ok {
+			return purl.Name, nil
+		}
+		r := strings.SplitN(ru, "/", 2)
```
RTann, line 644: nit: this is fine, but I tend to prefer `strings.Cut` instead of `strings.SplitN(<whatever>, <whatever>, 2)`

rhel/vex/updater.go
```diff
@@ -11,15 +11,21 @@ import (
...
-	_ driver.Configurable      = (*Updater)(nil)
-	_ driver.DeltaUpdater      = (*Updater)(nil)
-	_ driver.UpdaterSetFactory = (*Factory)(nil)
-	_ driver.Configurable      = (*Factory)(nil)
+	_        driver.Updater           = (*Updater)(nil)
+	_        driver.Configurable      = (*Updater)(nil)
+	_        driver.DeltaUpdater      = (*Updater)(nil)
+	_        driver.UpdaterSetFactory = (*Factory)(nil)
+	_        driver.Configurable      = (*Factory)(nil)
+	rhccRepo                          = &claircore.Repository{
```
RTann, line 25: nit: put a newline above this, so the lines above this won't need to be modified

RTann: Also doesn't the `goldRepo` already exist?

rhel/vex/parser.go
crozzy: yeah, better

rhel/vex/parser.go
```diff
@@ -290,8 +272,46 @@ func (c *creator) knownAffectedVulnerabilities(ctx context.Context, v csaf.Vulne
...
+			if err != nil {
+				zlog.Warn(ctx).
+					Str("purlHelper", purlHelper).
+					Err(err).
+					Msg("could not parse PURL")
+			}
+			if !checkPURL(purl) {
+				continue
+			}
+			pkgName, err = extractPackageName(purl)
```
RTann, line 289: `extractPackageName` returns `""` for `pkgName` when `err != nil`, and I think the hope is to leave it as-is

rhel/vex/parser.go
```diff
@@ -331,9 +352,59 @@ func (c *creator) lookupVulnerability(vulnKey string, protoVulnFunc func() *clai
...
+func (r *ranger) add(packageName, fixedInVersion string) (*claircore.Range, error) {
+	var rng = &claircore.Range{}
+
+	if fixedInVersion == "" {
+		zeroVer := &rhctag.Version{}
+		rng.Lower = zeroVer.Version(true)
+		highestVer := rhctag.Version{
+			Major: math.MaxInt32, // Everything is vulnerable
+		}
+		rng.Upper = highestVer.Version(true)
```
RTann, line 374: should this be `false` or does it really not matter much?

rhel/vex/parser.go
```diff
@@ -331,9 +352,59 @@ func (c *creator) lookupVulnerability(vulnKey string, protoVulnFunc func() *clai
...
+	} else {
+		firstPatch, err := rhctag.Parse(fixedInVersion)
+		if err != nil {
+			return nil, err
+		}
+		rng.Upper = firstPatch.Version(false)
+		start, err := rhctag.Parse(fixedInVersion)
+		if err != nil {
+			return nil, err
```
RTann, line 383: it'd be very weird if this managed to fail after it just succeeded a few lines above. Also a bit unfortunate you have to call the same thing twice

rhel/vex/parser.go
```diff
@@ -331,9 +352,59 @@ func (c *creator) lookupVulnerability(vulnKey string, protoVulnFunc func() *clai
...
+	lowest map[string]*claircore.Range
+}
+
+func newRanger() *ranger {
+	return &ranger{
+		lowest: map[string]*claircore.Range{},
+	}
+}
+
+func (r *ranger) add(packageName, fixedInVersion string) (*claircore.Range, error) {
```
RTann, line 365: what exactly is this doing? I'm a bit confused about how the fixed ones are handled

rhel/vex/parser.go
crozzy: Good catch, I think we can get away with just doing it once

rhel/vex/parser.go
crozzy: I think `true` is correct as the major is already maxed out, I think it's better as `v2147483647.0.0` rather than `v2147483647.0.2147483647`

rhel/vex/parser.go
crozzy: I added a the comment 
```
// Add takes a packageName and a fixedInVersion and returns a *claircore.Range relating
// to the fixedInVersion. Add also saves the lowest range per packageName that it's seen.
// This allows zeroLowest() to zero out the lowest values we saw per packageName.
```

rhel/vex/parser.go
```diff
@@ -331,9 +353,58 @@ func (c *creator) lookupVulnerability(vulnKey string, protoVulnFunc func() *clai
 	return &c.fixedVulns[idx], ok
 }
 
+type ranger struct {
```
RTann, line 356: I think I'm still struggling with figuring out the purpose of this and why it's shared per VEX doc

rhel/vex/parser.go
```diff
@@ -290,8 +272,47 @@ func (c *creator) knownAffectedVulnerabilities(ctx context.Context, v csaf.Vulne
...
+		if ok {
+			purl, err := packageurl.FromString(purlHelper)
+			if err != nil {
+				zlog.Warn(ctx).
+					Str("purlHelper", purlHelper).
+					Err(err).
+					Msg("could not parse PURL")
+			}
+			if !checkPURL(purl) {
+				continue
```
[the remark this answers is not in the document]
crozzy: I'm not sure I fully understand this, we need to extract the pURL before we can check it. Maybe we can move it before the creation of the vulnerability, but I think it's a little more readable setting the `vuln.Repo` and `vuln.Range` directly rather than dealing with more variables

rhel/vex/parser.go
```diff
@@ -508,12 +602,91 @@ func createPackageKey(repo, name, fixedIn string) string {
...
+func checkPURL(purl packageurl.PackageURL) bool {
+	if ok := acceptedTypes[purl.Type]; !ok {
+		return false
+	}
+	if strings.HasPrefix(purl.Name, "kernel") {
+		// We don't want to ingest kernel advisories as
+		// containers have no say in the kernel.
+		return false
+	}
+	if purl.Type == packageurl.TypeRPM && purl.Namespace != "redhat" {
```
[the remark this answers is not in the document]
crozzy: The package's pURL should never be an `rpmmod` so their handling will always be different and as such the constraints are different, also we do `checkPURL` after we deal with the module information. Not sure I 100% follow

rhel/vex/parser.go
```diff
@@ -508,12 +602,91 @@ func createPackageKey(repo, name, fixedIn string) string {
...
+		return false
+	}
+	return true
+}
+
+func extractArch(purl packageurl.PackageURL) string {
+	arch := purl.Qualifiers.Map()["arch"]
+	switch arch {
+	case "amd64", "x86_64":
+		return "amd64|x86_64"
```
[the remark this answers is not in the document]
crozzy: yeah, legacy reasons I suppose

rhel/vex/parser.go
```diff
@@ -384,41 +455,63 @@ func (c *creator) fixedVulnerabilities(ctx context.Context, v csaf.Vulnerability
...
-			// probably change down the line when we consolidate updaters.
+		fixedIn, err := extractFixedInVersion(purl)
+		if err != nil {
+			zlog.Warn(ctx).
+				Str("purl", purl.String()).
+				Err(err).
+				Msg("error extracting fixed_in version from pURL")
+			continue
+		}
+		packageName, err := extractPackageName(purl)
```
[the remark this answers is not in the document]
crozzy: With `KnownAffected` vulns it is possible (and probable) that we won't have a pURL so the `pkgName` that we extract from `walkRelationships` could just end up being the ultimate `pkgName`. With `Fixed` vulnerabilities we *should* always get a purl so the `pkgName` that we extract from `walkRelationships` will never be the eventual `packageName` that is saved in the `Vulnerability.Package.Name` field.

rhel/vex/parser.go
crozzy: The reason it is scoped to the VEX document is that we need to know the lowest affect range(derived from the fixed in version) for that oci image out of all the entries for that image in the document. I.e. for that vulnerability, what is the lowest fixed in version for image `x`  so we are about to make an infinite lower-bound for the lowest. For example, we extract information like this:
```
OCI     Fixed in version    Range
myimg   4.5.2               4.5.0->4.5.2
myimg   4.4.7               4.4.0->4.4.7
myimg   4.3.12              4.3.0->4.3.12
```
But we don't actually want that we want the last range to be 
```
0.0.0->4.3.12
```
..to address anything before (at least that was the guidance). Hence the function to zero out the lowest- lowest bound.

RTann: ooooo I see now, thanks

rhel/vex/parser.go
```diff
@@ -362,9 +385,58 @@ func (c *creator) lookupVulnerability(vulnKey string, protoVulnFunc func() *clai
...
+		highestVer := rhctag.Version{
+			Major: math.MaxInt32, // Everything is vulnerable
+		}
+		rng.Upper = highestVer.Version(true)
+	} else {
+		firstPatch, err := rhctag.Parse(fixedInVersion)
+		if err != nil {
+			return nil, err
+		}
+		rng.Upper = firstPatch.Version(false)
```
RTann, line 416: this set the range to something like 4.5.0 - 4.5.MAX_INT. Shouldn't we only be affected up to the fixedInVersion, though?

rhel/vex/parser_test.go
```diff
@@ -292,6 +294,53 @@ func TestEscapeCPE(t *testing.T) {
 	}
 }
 
+func TestRHTAGParse(t *testing.T) {
```
RTann, line 297: Should this be in `pkg/rhctag/version_test.go` instead?

rhel/vex/parser_test.go
```diff
@@ -373,11 +422,198 @@ func TestParse(t *testing.T) {
 				t.Fatalf("failed to parse CSAF JSON: %v", err)
 			}
 			if len(vulns) != tc.expectedVulns {
-				t.Fatalf("expected %d vulns but got %d", tc.expectedVulns, len(vulns))
+				t.Errorf("expected %d vulns but got %d", tc.expectedVulns, len(vulns))
```
RTann, line 425: curious why change this

test/rhel/testdata/server.txt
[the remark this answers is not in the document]
crozzy: I guess that's the standard now in claircore

rhel/vex/parser_test.go
crozzy: Moved and reformatted to fit in with the current tests there

rhel/vex/test_common.go
```diff
@@ -0,0 +1,74 @@
...
+	// Do headers
+	tp := textproto.NewReader(bufio.NewReader(bytes.NewReader(h)))
+	hdr, err := tp.ReadMIMEHeader()
+	if err != nil && err != io.EOF {
+		return "", nil, err
+	}
+	return string(compressedFilepath), http.Header(hdr), nil
+}
+
+func ServeSecDB(t *testing.T, txtarFile string) (string, *http.Client) {
```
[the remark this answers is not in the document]
crozzy: I believe this is not possible https://github.com/golang/go/issues/31135

rhel/vex/test_common.go
```diff
@@ -0,0 +1,74 @@
...
+	relFilepath, headers, err := parseFilenameHeaders(archive.Comment)
+	if err != nil {
+		t.Fatal(err)
+	}
+	filename := filepath.Base(relFilepath)
+	mux.HandleFunc("/"+filename, func(w http.ResponseWriter, _ *http.Request) {
+		for k, v := range headers {
+			w.Header().Set(k, v[0])
+		}
+
```
RTann, line 45: guessing `MethodHead` is never used since it was removed from here?

rhel/vex/parser_test.go
```diff
@@ -3,14 +3,15 @@ package vex
...
 	"io"
 	"net/http"
 	"net/url"
 	"os"
 	"testing"
 
 	"github.com/klauspost/compress/snappy"
-
+	"github.com/package-url/packageurl-go"
 	"github.com/quay/zlog"
```
RTann, line 15: I don't recall other files, but just checking the spacing here

rhel/vex/parser_test.go
```diff
@@ -373,11 +374,198 @@ func TestParse(t *testing.T) {
...
+				Version:   "1.24.6-5.module+el8.8.0+18083+cd85596b",
+				Qualifiers: packageurl.QualifiersFromMap(map[string]string{
+					"arch":  "ppc64le",
+					"epoch": "1",
+				}),
+			},
+			want: "buildah-debugsource",
+		},
+		{
+			name: "oci_with_repository_url",
```
RTann, line 502: want to add one more for when `repository_url` exists, but `Name` has a namespace?

RTann (dismissed):

rhel/vex/test_common.go
crozzy: The `ServeMux` routes the HEAD method by default (as long as the pattern doesn't exclude it) so having the switch isn't really necessary.

rhel/vex/parser_test.go
crozzy: I believe this is correct now but was incorrect previously

rhel/vex/parser_test.go
crozzy: Done

rhel/vex/test_common.go
crozzy: Turns out it was necessary, I was trying to track down why tests were flaking with:
```
HTTP/1.x transport connection broken: malformed HTTP response "..."
```
Then when debugging that I found a warning from a previous call:
```
Unsolicited response received on idle HTTP channel starting with "..."
```
So yeah, we shouldn't be sending anything back for HEAD requests, the `ServeMux` won't automatically do that.

RTann (dismissed):

crozzy:
@RTann realized I should iterate the updater version

RTann (dismissed):

RTann (approved):
From what I can tell, the only change is the fixed spelling

Issue #1382, opened 2024-08-18
Invalid CVSS v2 and v4 implementations


pandatix:
Hey, follow up on #1230, #1232 and its comments.

The current implementation is still invalid for both CVSS v2 and **v4.0**.

## CVSS v2.0

Vectors like `AV:A/AC:L/Au:N/A:C/I:C/C:C` are invalid due to invalid metric order (`C`, `I` then `A` is valid, anything else is not).
The current implementation validates such inputs, which makes it inconsistent and non-compliant according to the FIRST.ORG spec.

## CVSS v4.0

Vectors like `CVSS:4.0/AV:A/AC:L/AT:N/PR:L/UI:A/VC:L/VI:L/VA:N/SC:N/SI:N/SA:H/CR:X` should produce a **score of 4.1** according to the [official FIRST.ORG calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:A/AC:L/AT:N/PR:L/UI:A/VC:L/VI:L/VA:N/SC:N/SI:N/SA:H/CR:X), but the current implementation returns a **4.7**.
Same impact as for v2, inconsistency and non-compliance.

Moreover, it is a big issue for Vulnerability Management policies, as many of them considers CVSS scores thresholds for prioritization. Those large deltas could pollute the security efforts of the consumers.

PR #1386, opened 2024-08-20
cvss: bugs, tests

hdonnay:
This fixes some bugs and moves the test fixtures to external files (something that was bugging me).

Closes: #1382


toolkit/types/cvss/cvss_v2_parse.rl
```diff
@@ -53,9 +53,11 @@ cr  = ( "CR"  >metric_start %mark_metric) ":" ( ("L"|"M"|"H"|"ND")           >va
 ir  = ( "IR"  >metric_start %mark_metric) ":" ( ("L"|"M"|"H"|"ND")           >value_start %mark_value);
 ar  = ( "AR"  >metric_start %mark_metric) ":" ( ("L"|"M"|"H"|"ND")           >value_start %mark_value);
 
-metric = ( av | ac | au | c | i | a | e | rl | rc | cdp | td | cr | ir | ar );
+base = av "/" ac "/" au "/" c "/" i "/" a;
+temporal = e "/" rl "/" rc;
+environmental = cdp "/" td "/" cr "/" ir "/" ar;
```
RTann, line 58: I'm not as familiar with these. Let's say `E` is set. Does that mean `RL` and `RC` both must also be set?

toolkit/types/cvss/cvss_v4.go
```diff
@@ -62,7 +62,7 @@ func (v *V4) getString(m V4Metric) (string, error) {
 // GetScore implements [Vector].
 func (v *V4) getScore(m V4Metric) byte {
 	b := v.mv[int(m)]
-	if m >= V4ExploitMaturity && b == 0 {
+	if m >= V4ExploitMaturity && (b == 0 /* not present */ || b == 'X' /* not defined */) {
```
RTann, line 65: Are you able to show me where in the spec it says this? The 4.0 spec does not seem to show me a nice formula that v3.1 had

toolkit/types/cvss/cvss_v2_parse.rl
hdonnay: I think so, yes. The v2 spec is (IMO) unclear on this, but seems to say that if you use a metric group, all the metrics must be filled out.

toolkit/types/cvss/cvss_v4.go
hdonnay: Where it says what? That these have specific defaults? That's documented under the relevant tables in the [v4 scoring section](https://www.first.org/cvss/v4.0/specification-document#CVSS-v4-0-Scoring-using-MacroVectors-and-Interpolation).

crozzy (approved):
my ragel understanding is a WIP but I think the changes make sense

PR #1388, opened 2024-08-21
java: slow tests

hdonnay:
I got really annoyed with the speed of the `java` tests, so I worked out how to not use the entire `osv` package and instead use fixtures.


hdonnay:
ok  	github.com/quay/claircore/java	30.435s	coverage: 28.2% of statements

This normally takes ~500s

java/matcher_test.go
```diff
@@ -5,16 +5,19 @@ import (
...
 )
 
 func TestVulnerable(t *testing.T) {
 	matcher := &Matcher{}
 
 	testcases := []struct {
-		name   string
 		record *claircore.IndexRecord
 		vuln   *claircore.Vulnerability
+		name   string
```
RTann, line 20: just curious why move this down? Seems like it tends to be the first thing set

java/matcher_test.go
```diff
@@ -199,3 +202,7 @@ func TestVulnerable(t *testing.T) {
 		})
 	}
 }
+
+func TestMatcher(t *testing.T) {
+	test.RunMatcherTests(zlog.Test(context.Background(), t), t, "testdata/matcher", new(Matcher))
```
RTann, line 207: so when this runs, it keeps the current directory context? (as in, it finds `testdata/matcher` relative to the `java` directory and not the `test` directory. I didn't realize that, cool

java/testdata/matcher/Jackson-Databind.txtar
RTann, line 1: just curious do you use a tool to create this or is this a decent amount of manual copy/paste + adding headers yourself?

RTann (approved):
curious are there plans to use this for other languages, too?

hdonnay:
> curious are there plans to use this for other languages, too?

Yeah, I'm about to slap this on the other tests that currently use the `osv` updater.

java/matcher_test.go
hdonnay: My editor complained about the layout causing extra padding, so I just let it rearrange the struct. The structs are all keyed, so it doesn't matter to the declarations.

java/testdata/matcher/Jackson-Databind.txtar
hdonnay: This was taking the output VulnerabilityReport and chopping it up, and taking the old fixture. We could probably write a tool to take a vulnerability report and spit out something close to this final, pretty output.

PR #1389, opened 2024-08-22
vex: fetcher add changes before archived data

crozzy:
By switching how the VEX data is appended to the output spool we allow the parser to ignore duplicates rather than have to process things twice. This helps in situations where RH prodsec update a large number of files and haven't updated the archive data.

### DIFF
--- a/rhel/vex/fetcher.go
+++ b/rhel/vex/fetcher.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"path"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


crozzy:
I went through the `zip` (and `tar`) rabbitholes and still ended up here.

rhel/vex/fetcher.go
```diff
@@ -62,48 +74,46 @@ func (u *Updater) Fetch(ctx context.Context, hint driver.Fingerprint) (io.ReadCl
...
-		uri, err := u.url.Parse(compressedFilename)
-		if err != nil {
-			return nil, hint, err
-		}
+	err = u.processDeletions(ctx, cw, fp, changed)
+	if err != nil {
+		return nil, hint, err
+	}
 
+	if processArchive {
```
crozzy, line 109: I'm kind of tempted to just make this an
```go
if compressedURL != nil {
```
But it feels like maybe it'd lose something in the readability.

hdonnay (approved):

PR #1393, opened 2024-08-26
cicd: automate list of "golang" image versions

hdonnay:
This change means we do not have to change this list. It also only updates upstream-supported versions; this can change back to some number of versions if desired.


crozzy (approved):
Awesome

PR #1399, opened 2024-08-29
matcher: cache query results for the same query

crozzy:
There is a possiblity that if the index record is referring to the same package but different repositories and the Query() function isn't adding a matcher constraint for the repository, queries to the DB can end up being identical. This patch makes sure that the same query is only executed once by caching the results and using them for subsequent calls.

### DIFF
--- a/datastore/postgres/get.go
+++ b/datastore/postgres/get.go
@@ -36,6 +36,11 @@ var (
 	)
 )
 
+type recordQuery struct {
+	record *claircore.IndexRecord
+	query  string
+}
+
 // Get implements vulnstore.Vulnerability.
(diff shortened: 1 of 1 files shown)


crozzy:
Pre change:
```
queryCount 2
queryCount 85
queryCount 101
```

Post change:
```
queryCount 2
queryCount 85
queryCount 404
```
For image quay.io/projectquay/clair:4.7.3, so it looks like in some cases it reduces the # of queries by 75% for RHEL (depending on how many repos we find in the image).

hdonnay (approved):
This is a good candidate to get updated to use the `unique` package when the go1.23 change happens.

PR #1401, opened 2024-09-04
vex: update some minor parsing logic

crozzy:
In non-debug environments (and always for known-affected vulnerabilities) we carry on parsing when we don't find a relationship. This means we error at the next block and continue then but it makes the logs more confusing.


hdonnay (approved):
LGTM

PR #1410, opened 2024-09-11
vex: buffer changes.csv response

crozzy:
Moving to the new security domain has surfaced an issue with keeping the changes.csv connection open long enough for it to be reset-by-peer. This change buffers the response to file before reading it.

### DIFF
--- a/rhel/vex/fetcher.go
+++ b/rhel/vex/fetcher.go
@@ -196,6 +196,12 @@ func (u *Updater) Fetch(ctx context.Context, hint driver.Fingerprint) (io.ReadCl
 // to w means they are deemed to have changed since the compressed
 // file was last processed. w and fp can be modified.
 func (u *Updater) processChanges(ctx context.Context, w io.Writer, fp *fingerprint) error {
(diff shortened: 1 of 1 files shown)


rhel/vex/fetcher.go
```diff
@@ -196,6 +196,11 @@ func (u *Updater) Fetch(ctx context.Context, hint driver.Fingerprint) (io.ReadCl
 // to w means they are deemed to have changed since the compressed
 // file was last processed. w and fp can be modified.
 func (u *Updater) processChanges(ctx context.Context, w io.Writer, fp *fingerprint) error {
+	tf, err := tmp.NewFile("", "rhel-vex-changes.")
+	if err != nil {
+		return err
+	}
```
hdonnay, line 202: Need to defer the close, otherwise the file leaks at any early return.

hdonnay (changes requested):

rhel/vex/fetcher.go
crozzy: oh dur

hdonnay (approved):

PR #1416, opened 2024-09-18
vex: account for package module when parsing VEX

crozzy:
Previously, we'd made the decision to ignore the package module and rely on the repo and package for matching but now that the VEX data has the package module data and it helps filter the results on the DB side, it seems best to add the package module information and keep the matcher constraints as it is. Without this we'd need change the RHEL matcher to remove the `PackageModule` constraint.


rhel/vex/parser.go
```diff
@@ -183,31 +183,46 @@ type creator struct {
...
 }
 
-// ExtractProductNames recursively looks up the package product_id and the repo product_id.
-// The assumtion is that the repo is always the last found relates_to_product_reference and the
-// package is the last found product_reference.
-func extractProductNames(prodRelID string, repoRelID string, c *csaf.CSAF) (string, string) {
-	prodRel := c.FindRelationship(prodRelID, "default_component_of")
+// ExtractProductNames recursively looks up product_id relationships and adds them to a
+// component slice in order.
+func extractProductNames(right, left string, comps []string, c *csaf.CSAF) []string {
```
RTann, line 212: It's a little confusing that `right` is the leftmost argument and `left` is to the right of `right`. Can we change this?

rhel/vex/parser.go
```diff
@@ -183,31 +183,46 @@ type creator struct {
...
-func walkRelationships(productID string, doc *csaf.CSAF) (string, string, error) {
-	pkgID, repoID := extractProductNames(productID, productID, doc)
-	if pkgID == repoID {
-		return "", "", fmt.Errorf("could not extract a distict pkgID and repoID from %q", productID)
+// WalkRelationships attempts to resolve a relationship until we have a package product_id,
+// a repo product_id and possibly a package module product_id. Relationships can be nested.
+// If we don't get an initial relationship or we don't get two component parts we cannot
+// create a vulnerability. We never see more than 3 components in the wild but if we did
+// we'd assume the component next to the repo product_id is the package module.
+func walkRelationships(productID string, doc *csaf.CSAF) (string, string, string, error) {
```
RTann, line 191: hoping you can help me walk through this example:

I'm looking at https://security.access.redhat.com/data/csaf/v2/vex/2023/cve-2023-47038.json and I'm interested in `perl-Carp`.

I can find "AppStream-8.10.0.GA:perl:5.32:8100020240314121426:9fe1d287:perl-Carp-0:1.50-439.module+el8.10.0+21354+3ad137bb.noarch" in the "fixed" section, so I use that as `productID` for this function. Then, I look for the "default_component_of" with this product ID. This gives me a `prodRel` looking like

```
ProductRef:                 "perl-Carp-0:1.50-439.module+el8.10.0+21354+3ad137bb.noarch",
RelatesToProductRef: "AppStream-8.10.0.GA:perl:5.32:8100020240314121426:9fe1d287",
```

Then inside `extractProductNames` I look for "default_component_of" with `product_id` of `perl-Carp-0:1.50-439.module+el8.10.0+21354+3ad137bb.noarch`.

This is where I get confused. I don't see this existing anywhere. Can you tell me if I'm reading this wrong?

rhel/vex/parser.go
crozzy: It's referring to the sides of productID string, i.e 
```
repo:mod:pkg
```
 which will then become:
```
repo:mod    pkg
```
and then:
```
repo    mod     pkg
```
The order of the comps is reversed when they're returned so I can see that might be confusing, what would you suggest to make it less so? `prodRef` and `relatesToProdRef`?

rhel/vex/parser.go
crozzy: `perl-Carp-0:1.50-439.module+el8.10.0+21354+3ad137bb.noarch` exists as a product in the product branches, that's the end of the road for that product, no more relationships to resolve. `AppStream-8.10.0.GA:perl:5.32:8100020240314121426:9fe1d287` on the other hand refers to another relationship: 
```json
        "product_reference": "perl:5.32:8100020240314121426:9fe1d287",
        "relates_to_product_reference": "AppStream-8.10.0.GA"
```
Both of these refer to a product in the product branches. Does this make sense?

RTann: so `comps = extractProductNames(prodRel.ProductRef, prodRel.RelatesToProductRef, comps, doc)` will return ```
comps = []string{"perl-Carp-0:1.50-439.module+el8.10.0+21354+3ad137bb.noarch", "perl:5.32:8100020240314121426:9fe1d287", "AppStream-8.10.0.GA"}
```
?

Comment (RTann) [reply rhel/vex/parser.go:191]: package: `perl-Carp-0:1.50-439.module+el8.10.0+21354+3ad137bb.noarch`
module: `perl:5.32:8100020240314121426:9fe1d287`
repo: `AppStream-8.10.0.GA`

Comment (crozzy) [reply rhel/vex/parser.go:191]: Yeah, added a unit-test to confirm

File: rhel/vex/parser.go (hunk 1 above)
Comment (RTann) [reply rhel/vex/parser.go:212]: yeah maybe something descriptive like `prodRef ` and `relatesToProdRef ` like you suggested would be good.

File: rhel/vex/parser.go [hunk 3]
```diff
@@ -183,31 +183,46 @@ type creator struct {
...
+	comps := []string{}
```
RTann, line 192: nit: maybe initialize this closer to where it's used. No need to define this if we may just exit early

rhel/vex/parser.go
```diff
@@ -183,31 +183,46 @@ type creator struct {
...
+	prodRel := doc.FindRelationship(productID, "default_component_of")
+	if prodRel == nil {
+		return "", "", "", fmt.Errorf("cannot determine initial relationship for %q", productID)
+	}
+	comps = extractProductNames(prodRel.ProductRef, prodRel.RelatesToProductRef, comps, doc)
```
RTann, line 197: I'm wondering if 

```
comps := extractProductNames(prodRel.ProductRef, prodRel.RelatesToProductRef, []string{}, doc)
```

or 

```
comps := []string{}
extractProductNames(prodRel.ProductRef, prodRel.RelatesToProductRef, &comps, doc)
```

would be better

rhel/vex/parser.go
```diff
@@ -183,31 +183,46 @@ type creator struct {
...
-	return pkgID, repoID, nil
 }
 
-// ExtractProductNames recursively looks up the package product_id and the repo product_id.
-// The assumtion is that the repo is always the last found relates_to_product_reference and the
-// package is the last found product_reference.
-func extractProductNames(prodRelID string, repoRelID string, c *csaf.CSAF) (string, string) {
-	prodRel := c.FindRelationship(prodRelID, "default_component_of")
+// ExtractProductNames recursively looks up product_id relationships and adds them to a
+// component slice in order.
```
RTann, line 211: can you maybe expand on what the order is?

rhel/vex/parser.go
```diff
@@ -501,11 +543,30 @@ func cvssVectorFromScore(sc *csaf.Score) (vec string, err error) {
...
+func createPackageKey(repo, mod, name, fixedIn string) string {
 	// The other option here is just to use repo + PURL string
 	// w/o the qualifiers I suppose instead of repo + NEVR.
-	return repo + ":" + name + "-" + fixedIn
+	return repo + ":" + mod + ":" + name + "-" + fixedIn
+}
+
+func createPackageModule(p *csaf.Product) (string, error) {
+	var modName string
+	modPurlHelper, ok := p.IdentificationHelper["purl"]
```
RTann, line 555: nit: `modPURLHelper`

rhel/vex/parser.go
```diff
@@ -501,11 +543,30 @@ func cvssVectorFromScore(sc *csaf.Score) (vec string, err error) {
...
+	purl, err := packageurl.FromString(modPurlHelper)
+	switch {
+	case err != nil:
+		return "", err
+	default:
+		parts := strings.SplitN(purl.Version, ":", 2)
+		if len(parts) > 0 {
+			modName = purl.Name + ":" + parts[0]
+		}
+	}
```
RTann, line 568: Up to you:

```suggestion
	purl, err := packageurl.FromString(modPurlHelper)
	if err != nil {
		return "", err
	}
	if version, _, found := strings.Cut(purl.Version, ":"); found {
		modName = purl.Name + ":" + version
	}
```

rhel/vex/parser.go
```diff
@@ -501,11 +543,30 @@ func cvssVectorFromScore(sc *csaf.Score) (vec string, err error) {
...
+	return repo + ":" + mod + ":" + name + "-" + fixedIn
+}
+
+func createPackageModule(p *csaf.Product) (string, error) {
+	var modName string
+	modPurlHelper, ok := p.IdentificationHelper["purl"]
+	if !ok {
+		return modName, nil
+	}
+	purl, err := packageurl.FromString(modPurlHelper)
```
RTann, line 559: should we check the type here? `purl.Type == "rpmmod"`?

rhel/vex/parser_test.go
```diff
@@ -121,19 +181,54 @@ func TestWalkRelationships(t *testing.T) {
 			},
 			in:               "CROS:J-MOD:J-COMP:JMC",
```
RTann, line 182: It's odd `J-COMP` isn't used, but I know you said you've only seen something like `CROS:J-MOD:JMC` in the wild, and having that fourth thing is rather unexpected, so the behavior this is testing is fine with me

rhel/vex/parser.go
```diff
@@ -183,31 +183,46 @@ type creator struct {
...
-// product_id, either way we don't have enough data to create a vulnerability.
-func walkRelationships(productID string, doc *csaf.CSAF) (string, string, error) {
-	pkgID, repoID := extractProductNames(productID, productID, doc)
-	if pkgID == repoID {
-		return "", "", fmt.Errorf("could not extract a distict pkgID and repoID from %q", productID)
+// WalkRelationships attempts to resolve a relationship until we have a package product_id,
+// a repo product_id and possibly a package module product_id. Relationships can be nested.
+// If we don't get an initial relationship or we don't get two component parts we cannot
+// create a vulnerability. We never see more than 3 components in the wild but if we did
+// we'd assume the component next to the repo product_id is the package module.
```
RTann, line 190: do we have any logs output when unexpected formatting happens (like a 4th components)?

crozzy: >2 components (i.e. more complex relationships) are valid as per the spec so we account for it in the return values, so there is no warning currently.

rhel/vex/parser_test.go
crozzy: The test is to verify that no matter how many relationships we need to resolve we still pick/identify the correct components. Thinking about how many clair instances will be in the wild we wouldn't be able to react if the data changes (and it would be within the CSAF spec if prodsec created more complex relationships).

rhel/vex/parser.go
```diff
@@ -501,11 +546,32 @@ func cvssVectorFromScore(sc *csaf.Score) (vec string, err error) {
...
+	purl, err := packageurl.FromString(modPURLHelper)
+	if err != nil {
+		return "", err
+	}
+	if purl.Type != "rpmmod" {
+		return "", fmt.Errorf("invalid RPM module PURL: %q", purl.String())
+	}
+	if version, _, found := strings.Cut(purl.Version, ":"); found {
+		modName = purl.Name + ":" + version
+	} else {
```
crozzy, line 571: @RTann adding this clause as some of the PURLs don't conform to a standard format and will be potentially updated:
Conventional: `pkg:rpmmod/redhat/postgresql@13:8080020231114105206:63b34585`
Unconventional: `pkg:rpmmod/redhat/postgresql:15/postgresql`
Possible update for unconventional format: `pkg:rpmmod/redhat/postgresql@15`

This is being looked at by RH prodsec.

RTann: Based on https://go.dev/play/p/PamAwGzH8r0 I think you can just do

```
version, _, _ := strings.Cut(purl.Version, ":")
return purl.Name + ":" + version, nil
```

rhel/vex/parser.go
```diff
@@ -501,11 +546,32 @@ func cvssVectorFromScore(sc *csaf.Score) (vec string, err error) {
...
+// i.e. AppStream-8.2.0.Z.TUS:a_module:python3-idle-0:3.6.8-24.el8_2.2
+func createPackageKey(repo, mod, name, fixedIn string) string {
 	// The other option here is just to use repo + PURL string
 	// w/o the qualifiers I suppose instead of repo + NEVR.
-	return repo + ":" + name + "-" + fixedIn
+	return repo + ":" + mod + ":" + name + "-" + fixedIn
+}
+
+func createPackageModule(p *csaf.Product) (string, error) {
+	var modName string
```
RTann, line 557: This may not be needed anymore

rhel/vex/parser.go
RTann: > This is being looked at by RH prodsec.

Is this blocked, then?

crozzy: I'm not sure there is a point in editing it until we get a reply about whether this is:

1. A bug and will be fixed
2. A bug and won't be immediately fixed
3. Not a but and won't be changed

If we don't get that answer soon I think we just account for the conventional way and possibly modify in the future.

crozzy: It's been a couple of days with no reply, going to change to just account for conventional `rpmmod` PURLs

rhel/vex/parser.go
```diff
@@ -501,11 +546,41 @@ func cvssVectorFromScore(sc *csaf.Score) (vec string, err error) {
...
+			modName = purl.Name + ":" + purl.Version
+		}
+	case strings.HasPrefix(purl.Namespace, "redhat") && strings.Contains(purl.Namespace, "/"):
+		// Account for the case where the PURL is unconventional
+		// e.g. pkg:rpmmod/redhat/postgresql:15/postgresql
+		if _, module, found := strings.Cut(purl.Namespace, "/"); found {
+			modName = module
+		}
+	}
+	return modName, nil
```
crozzy, line 583: This results in distinct modules like this:
```
                     
389-ds:1.4           
container-tools:1.0  
container-tools:2.0  
container-tools:3.0  
container-tools:4.0  
container-tools:rhel8
freeradius:3.0       
gimp:2.8             
go-toolset:rhel8     
httpd:2.4            
idm:client           
idm:DL1              
llvm-toolset:rhel8   
mailman:2.1          
mariadb:10.3         
mariadb:10.5         
mariadb-devel:10.3   
maven:3.5            
maven:3.6            
mod_auth_openidc:2.3 
mysql:8.0            
nginx:1.14           
nginx:1.16           
nginx:1.18           
nginx:1.20           
nginx:1.22           
nodejs:10            
nodejs:12            
nodejs:14            
nodejs:16            
nodejs:18            
nodejs:20            
parfait:0.5          
perl:5.32            
php:7.2              
php:7.3              
php:7.4              
php:8.0              
php:8.1              
pki-core:10.6        
pki-deps:10.6        
postgresql:10        
postgresql:12        
postgresql:13        
postgresql:15        
postgresql:16        
postgresql:9.6       
python27:2.7         
python36:3.6         
python38:3.8         
python38-devel:3.8   
python39:3.9         
python39-devel:3.9   
redhat-ds:11         
redhat-ds:12         
redhat-pki:10        
redis:5              
redis:6              
redis:7              
ruby:2.5             
ruby:2.6             
ruby:2.7             
ruby:3.0             
ruby:3.1             
ruby:3.3             
rust-toolset:rhel8   
squid:4              
subversion:1.10      
subversion:1.14      
varnish:6            
virt:8.0.0           
virt:8.1             
virt:8.2             
virt:8.3             
virt:av              
virt-devel:8.1       
virt-devel:8.2       
virt-devel:8.3       
virt-devel:av        
virt-devel:rhel      
virt:rhel
```

rhel/vex/parser.go
```diff
@@ -501,11 +546,41 @@ func cvssVectorFromScore(sc *csaf.Score) (vec string, err error) {
...
+	case purl.Namespace == "redhat":
+		if version, _, found := strings.Cut(purl.Version, ":"); found {
+			modName = purl.Name + ":" + version
+		} else {
+			modName = purl.Name + ":" + purl.Version
+		}
+	case strings.HasPrefix(purl.Namespace, "redhat") && strings.Contains(purl.Namespace, "/"):
+		// Account for the case where the PURL is unconventional
+		// e.g. pkg:rpmmod/redhat/postgresql:15/postgresql
+		if _, module, found := strings.Cut(purl.Namespace, "/"); found {
```
RTann, line 579: nit: up to you, but we know at this point `found == true` because of the condition above, so we can just use do something like

```
_, modName, _ = strings.Cut(purl.Namespace, "/")
```

rhel/vex/parser.go
RTann, line 583: when the namespace does not start with `redhat` we return `"", nil`. Is that expected?

rhel/vex/parser.go
crozzy: Yeah, I think this is fair

rhel/vex/parser.go
crozzy: Going to return an error in the default case. It's a little difficult without official documentation on the format of pURLs to make decisions like this, but I think if we don't get something _like_ a `redhat` pURL then this function will error and the update will log and continue.

RTann (approved):
one minor nit otherwise LGTM

rhel/vex/parser.go
```diff
@@ -501,11 +546,42 @@ func cvssVectorFromScore(sc *csaf.Score) (vec string, err error) {
...
+	if purl.Type != "rpmmod" {
+		return "", fmt.Errorf("invalid RPM module PURL: %q", purl.String())
+	}
+	switch {
+	case purl.Namespace == "redhat":
+		if version, _, found := strings.Cut(purl.Version, ":"); found {
+			modName = purl.Name + ":" + version
+		} else {
+			modName = purl.Name + ":" + purl.Version
+		}
```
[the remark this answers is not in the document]
crozzy: Good one

rhel/vex/parser.go
RTann: won't block the PR, but this could be defined directly above the switch and the return directly below this can just be `""` to match the other scenarios

RTann (approved):

RTann (approved):

PR #1422, opened 2024-10-02
alpine: use alpine security vuln link

RTann:
It was brought to my attention https://security.alpinelinux.org exists, and we may find vulns in `https://security.alpinelinux.org/vuln/<CVE ID>`. For example: https://security.alpinelinux.org/vuln/CVE-2018-20187.

This is now my second time updating the link we use for Alpine vulns. I believe last time there was concern about migrations and whatnot, but I believe since then we added a materialized view when looking at the latest vulnerability data in the DB, so this may be a non-issue now. This is all to say: I do not believe I need to add a migration.

Went back to the other PR https://github.com/quay/claircore/pull/810, and it seems it was allowed after https://github.com/quay/claircore/pull/947 was merged, though I can see querybuilder.go has been updated since then.

### DIFF
--- a/alpine/parser.go
+++ b/alpine/parser.go
@@ -3,7 +3,6 @@ package alpine
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"io"
 
 	"github.com/quay/zlog"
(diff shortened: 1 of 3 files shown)


crozzy (approved):
SGTM

PR #1428, opened 2024-10-18
osv: account for event objects that have multiple streams

crozzy:
It was discovered that some OSV documents can order minor releases in the same affected.ranges object. This meant that only ever counted the last range in a vulnerability. This change gathers range information for the affected product and creates a vulnerability per range.

### DIFF
--- a/updater/osv/osv.go
+++ b/updater/osv/osv.go
@@ -492,6 +492,12 @@ func newECS(u string) ecs {
 	}
 }
 
+type rangeVer struct {
+	fixedInVersion string
+	semverRange    *claircore.Range
+	ecosystemRange url.Values
+}
+
 func (e *ecs) Insert(ctx context.Context, skipped *stats, name string, a *advisory) (err error) {
 	if a.GitOnly() {
 		return nil
@@ -544,12 +550,10 @@ func (e *ecs) Insert(ctx context.Context, skipped *stats, name string, a *adviso
 	proto.Links = b.String()
 	for i := range a.Affected {
 		af := &a.Affected[i]
-		v := e.NewVulnerability()
-		(*v) = proto
 		for _, r := range af.Ranges {
+			vers := []*rangeVer{}
 			switch r.Type {
 			case `SEMVER`:
-				v.Range = &claircore.Range{}
 			case `ECOSYSTEM`:
 				b.Reset()
 			case `GIT`:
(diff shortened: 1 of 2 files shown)


crozzy:
Example ranges from an OSV document.
```json
      "ranges": [
        {
          "type": "SEMVER",
          "events": [
            {
              "introduced": "0"
            },
            {
              "fixed": "1.21.12"
            },
            {
              "introduced": "1.22.0-0"
            },
            {
              "fixed": "1.22.5"
            }
          ]
        }
      ],
```

updater/osv/osv.go
```diff
@@ -554,24 +558,40 @@ func (e *ecs) Insert(ctx context.Context, skipped *stats, name string, a *adviso
...
+				ev := r.Events[ie]
 				var err error
 				switch r.Type {
 				case `SEMVER`:
 					var ver *semver.Version
 					switch {
-					case ev.Introduced == "0": // -Inf
-						v.Range.Lower.Kind = `semver`
+					case ev.Introduced != "" && seenIntroduced:
+						vs = &rangeVer{rng: &claircore.Range{}}
```
RTann, line 571: if we are assuming it's well-formed (introduced, fixed, introduced, fixed, etc), then I think we can just do `vs = &rangeVer{rng: &claircore.Range{}}` unconditionally within `case ev.Introduced` and don't need the `if ie == len(r.Events)-1` below. What do you think?

updater/osv/osv.go
```diff
@@ -554,24 +558,40 @@ func (e *ecs) Insert(ctx context.Context, skipped *stats, name string, a *adviso
 			}
 			// This does some heavy assumptions about valid inputs.
 			ranges := make(url.Values)
```
RTann, line 560: does this need the same kind of treatment?

updater/osv/osv.go
crozzy: I think we need to deal with just (introduced) and (introduced, fixed, introduced) i.e. the cases when the last event is introduced (make sure to append the `vs`), and where it is well terminated (introduced, fixed, introduced, fixed) (leave the `vs` open ended to be potentially filled by another event). The solo introduced case definitely needs to be accounted for, the introduced, fixed, introduced, seems less likely but would also be accounted for.

updater/osv/osv.go
crozzy: https://api.osv.dev/v1/vulns/PYSEC-2008-3 Good catch, it appears like yes

crozzy: Updated

updater/osv/osv.go
```diff
@@ -600,14 +619,26 @@ func (e *ecs) Insert(ctx context.Context, skipped *stats, name string, a *adviso
 				case `ECOSYSTEM`:
 					switch af.Package.Ecosystem {
 					case ecosystemMaven, ecosystemPyPI, ecosystemRubyGems:
+						vs.semverRange = nil
```
RTann, line 622: should this be mirrored above to make `vs.ecosystemRanges = nil`?

updater/osv/osv.go
```diff
@@ -484,6 +484,12 @@ func newECS(u string) ecs {
 	}
 }
 
+type rangeVer struct {
+	fixedInVersion  string
+	semverRange     *claircore.Range
+	ecosystemRanges url.Values
```
RTann, line 490: nit: should they both be singular/plural, or does this make most sense as-is?

updater/osv/osv_test.go
```diff
@@ -138,18 +141,383 @@ func TestParse(t *testing.T) {
...
+					Ranges: []_range{
+						{
+							Type: "SEMVER",
+							Events: []rangeEvent{
+								{
+									Introduced: "0",
+								},
+								{
+									Fixed: "2.1.16",
+								},
```
RTann, line 323: is it also valid to not have `Introduced` here (I think no, but I think we cover that case? If we do, can you add another test for that case)?

updater/osv/osv.go
crozzy: We can deduce that the `vs.ecosystemRanges` haven't been touched later by checking the `len()` we can't really do that with `vs.semverRange` so that's why its explicitly nil(l)ed

updater/osv/osv.go
crozzy: Both singular I think :+1:

updater/osv/osv_test.go
crozzy: Added

updater/osv/osv.go
```diff
@@ -608,14 +627,26 @@ func (e *ecs) Insert(ctx context.Context, skipped *stats, name string, a *adviso
...
-						case ev.Introduced == "0":
+						case ev.Introduced != "" && seenIntroduced:
+							vs = &rangeVer{ecosystemRange: url.Values{}}
+							fallthrough
 						case ev.Introduced != "":
-							ranges.Add("introduced", ev.Introduced)
+							seenIntroduced = true
+							switch {
+							case ev.Introduced == "0":
+							case ev.Introduced != "":
```
RTann, line 639: want to use `default` to mirror above?

updater/osv/osv_test.go
```diff
@@ -138,18 +141,451 @@ func TestParse(t *testing.T) {
...
+						Kind: "semver",
+						V:    [10]int32{65535, 0, 0, 0, 0, 0, 0, 0, 0, 0},
+					},
+				},
+				FixedInVersion: "",
+			},
+		},
+	},
+	{
+		name: "ecosystem_multi",
```
RTann, line 483: can you also add one more with separate events to ensure we still handle that?

updater/osv/osv_test.go
```diff
@@ -138,18 +141,509 @@ func TestParse(t *testing.T) {
...
+			{
+				Name:           "test1",
+				Updater:        "test",
+				Range:          nil,
+				FixedInVersion: "introduced=3.0",
+			},
+		},
+	},
+	{
+		name: "same package different ranges",
```
RTann, line 562: sorry I meant for `ecosystem` since we already have one for `semver` a few tests above ^

crozzy: ahh

RTann (approved):

PR #1431, opened 2024-10-28
osv: disable more ecosystems

crozzy:
Disable more OSV ecosystems for situations where we already have a dedicated updater.

### DIFF
--- a/updater/osv/osv.go
+++ b/updater/osv/osv.go
@@ -197,12 +197,20 @@ func (f *Factory) UpdaterSet(ctx context.Context) (s driver.UpdaterSet, err erro
 
 // Ignore is a set of incoming ecosystems that we can throw out immediately.
 var ignore = map[string]struct{}{
+	"almalinux":      {}, // Currently not supported.
 	"alpine":         {}, // Have a dedicated alpine updater.
 	"android":        {}, // AFAIK, there's no Android container runtime.
+	"chainguard":     {}, // Currently not supported.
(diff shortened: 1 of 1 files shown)


updater/osv/osv.go
```diff
@@ -203,6 +203,9 @@ var ignore = map[string]struct{}{
 	"github actions": {}, // Shouldn't be in containers?
 	"linux":          {}, // Containers have no say in the kernel.
 	"oss-fuzz":       {}, // Seems to only record git revisions.
+	"ubuntu":         {}, // Have a dedicated ubuntu updater.
+	"red hat":        {}, // Have a dedicated Red Hat VEX updater.
+	"suse":           {}, // Have a dedicated SUSE updater.
```
RTann, line 208: nit: alphabetical order

RTann, line 208: Should we consider ones which we *may* have a dedicated updater for in the future? Like `almalinux`, `chainguard`, `rocky linux`, and `wolfi` (ie those which are related to Linux distros which do have their own vuln feeds?

Also, let's include `openSUSE` because I think the current suse support covers both?

crozzy: Probably, even if it was needed from OSV I think we'd be able to remove them from the list without too much hassle. I initially created it as I was testing something else and saw over a million OSV ubuntu rows.

crozzy:
This changes the number of OSV vuln rows from 1709163 to ~67000

RTann (approved):

PR #1432, opened 2024-10-31
suse: dynamic distribution discovery

crozzy:
Currently Suse distributions are predefined in the code, this change adds dynamic support for two Suse distro flavors:
suse.linux.enterprise.server and opensuse.leap.

This change is purposefully picky, it only considers versions:
```
suse.linux.enterprise.server.11
suse.linux.enterprise.server.12
suse.linux.enterprise.server.15

opensuse.leap.15.5
opensuse.leap.15.6
```

But should expand over time


suse/factory.go
```diff
@@ -0,0 +1,233 @@
...
+	dir, err := html.Parse(res.Body)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// It's kind of gross having to parse the HTML but the other path,
+	// parsing the SHA256SUMs, is more fragile. At least this way we
+	// know the file exists with the expected compression.
+	var fc func(*html.Node) error
+	fc = func(n *html.Node) error {
```
RTann, line 83: nit: `fc := func(n *html.Node) error { ... }`

suse/factory.go
```diff
@@ -0,0 +1,233 @@
...
+	default:
+		return us, fmt.Errorf("suse: unexpected status requesting OVAL dir: %v", res.Status)
+	}
+
+	dir, err := html.Parse(res.Body)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// It's kind of gross having to parse the HTML but the other path,
```
RTann, line 79: I am pretty thankful for https://security.access.redhat.com/data/oval/v2/PULP_MANIFEST when I look at the alternative below 😂

suse/factory.go
```diff
@@ -0,0 +1,233 @@
...
+		if n.Type == html.ElementNode && n.Data == "a" {
+			for _, a := range n.Attr {
+				if a.Key == "href" {
+					var d *claircore.Distribution
+					if parts := reELFile.FindAllStringSubmatch(a.Val, -1); len(parts) == 1 {
+						d = mkELDist(a.Val, parts[0][1])
+					}
+					if parts := reLeapFile.FindAllStringSubmatch(a.Val, -1); len(parts) == 1 {
+						d = mkLeapDist(a.Val, parts[0][1])
+					}
```
RTann, line 93: from StackRox perspective, this is fine, but from Clair's perspective, is it worth making this some kind of if/else or doing some kind of prefix match first? I know regexp evaluation is not exactly performant. Not sure how much that matters, though, nor how much of a perf hit it'd be (probably not a lot?)

suse/factory.go
```diff
@@ -0,0 +1,233 @@
...
+					if d == nil {
+						continue
+					}
+					uri, err := f.base.Parse(a.Val)
+					if err != nil {
+						return fmt.Errorf("unable to construct request for %q: %w", a.Val, err)
+					}
+					u, err := NewUpdater(d, WithURL(uri.String(), "bz2"))
+					if err != nil {
+						return fmt.Errorf("failed to parse : %v", err)
```
RTann, line 103:
```suggestion
						return fmt.Errorf("failed to parse uri %v: %w", uri, err)
```

something like that?

suse/factory.go
```diff
@@ -0,0 +1,233 @@
...
+					err = us.Add(u)
+					if err != nil {
+						return err
+					}
+
+				}
+			}
+		}
+		for c := n.FirstChild; c != nil; c = c.NextSibling {
+			err = fc(c)
```
RTann, line 114: ah, is this why it's declared above? never really made a recursive closure like this in Go before

RTann, line 114: this error may be overwritten and ignored if this child fails but the next succeeds. Is that desired or should we catch the error?

go.mod
```diff
@@ -59,6 +59,7 @@ require (
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
 	go.opentelemetry.io/otel/metric v1.31.0 // indirect
 	golang.org/x/mod v0.21.0 // indirect
+	golang.org/x/net v0.30.0 // indirect
```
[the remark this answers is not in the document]
crozzy: I mean it comes from this import `golang.org/x/net/html` do you mean a mistake that it's indirect?

suse/factory.go
crozzy: Yeah, it needs to be predefined I believe, I rewrote it as a separate function to be more clear

suse/factory.go
crozzy: This is handled differently with the separate function

suse/factory.go
crozzy: I think this is fine given that:
* This is a "background" process
* The regex is defined previously
* The regex is providing 2 functions, matching and returning relevant captured groups

go.mod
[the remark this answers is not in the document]
crozzy: I think it's how you'd expect now

pkg/ovalutil/pool.go
```diff
@@ -9,15 +9,20 @@ import (
 )
 
 var (
-	gzipPool sync.Pool
-	zstdPool sync.Pool
+	gzipPool = sync.Pool{
```
RTann, line 12: just curious: what is the purpose of these pools? If multiple updaters update OVAL data, wouldn't there be a chance  something is accidentally overwritten?

suse/factory.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/quay/zlog"
+	"golang.org/x/net/html"
+)
+
+//doc:url updater
+const base = `https://ftp.suse.com/pub/projects/security/oval/`
+
+var (
+	reELFile   = regexp.MustCompile(`suse.linux.enterprise.server.([1-9][1-9]).xml.gz`)
+	reLeapFile = regexp.MustCompile(`opensuse.leap.(1[0-9].[0-9]+).xml.gz`)
```
RTann, line 25: does this add a lot more rows? Do you think we should go back to the bzip2 files and just accept SUSE 11 is no longer supported (or just make a workaround for it)?

suse/factory.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, f.base.String(), nil)
+	if err != nil {
+		return us, fmt.Errorf("suse: unable to construct request: %w", err)
+	}
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	if f.etag != "" {
+		req.Header.Set(`if-none-match`, f.etag)
+	}
```
RTann, line 54: I don't think this is ever set? Also, I didn't see an etag returned when I curl'd an example endpoint

suse/distributionscanner.go
```diff
@@ -111,11 +88,51 @@ func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]
...
+	d, err := cpeToDist(c, ver)
+	if err != nil {
+		zlog.Warn(ctx).Err(err).Msg("error converting cpe to distribution")
+		return nil
+	}
+
+	return d
+}
+
+func cpeToDist(r cpe.WFN, ver string) (*claircore.Distribution, error) {
```
RTann, line 120: do we want to bother verifying the CPE is valid, too, like checking for `part == o` or nah?

suse/distributionscanner.go
```diff
@@ -111,11 +88,51 @@ func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]
...
+	if vendor, err := cpe.NewValue("opensuse"); err == nil && r.Attr[cpe.Vendor] == vendor {
+		if prod, err := cpe.NewValue("leap"); err == nil && r.Attr[cpe.Product] == prod {
+			return mkLeapDist(r.String(), ver), nil
+		}
+	}
+	if vendor, err := cpe.NewValue("suse"); err == nil && r.Attr[cpe.Vendor] == vendor {
+		if prod, err := cpe.NewValue("sles"); err == nil && r.Attr[cpe.Product] == prod {
+			// Canonicalize the version to the major.
+			v, err := semver.NewVersion(ver)
```
RTann, line 129: should we just use the version from the CPE?

RTann (commented):
a few more things, then should be good to go soon

pkg/ovalutil/pool.go
crozzy: From the docs:
```
A Pool is safe for use by multiple goroutines simultaneously.

Pool's purpose is to cache allocated but unused items for later reuse, relieving pressure on the garbage collector. That is, it makes it easy to build efficient, thread-safe free lists. However, it is not suitable for all free lists. 
```
In this case an updater can grab a reader, do what it wants and return it to the pool, as long as the returning happens after the data is read there isn't a way to overwrite the data. Once the reader is returned it's reset and ready for the next client/updater/goroutine to use.

suse/factory.go
crozzy: You can see from the PR desc that it does add some earlier opensuse LEAP versions. I added a minimum LEAP version, now the updates are only concerned with:
```
suse.linux.enterprise.server.11
suse.linux.enterprise.server.12
suse.linux.enterprise.server.15
opensuse.leap.15.5
opensuse.leap.15.6
```
The 3 supported ELS versions and the 2 [LEAP supported versions](https://en.opensuse.org/Lifetime).

suse/factory.go
crozzy: Fair enough, looks like the initial updaterset creation cannot rely on not-modifed directives, luckily the OVAL files themselves include etags.

suse/distributionscanner.go
crozzy: We are checking if it's a valid CPE, I think that is enough

suse/factory.go
RTann: oh I didn't notice you edited it. I thought it was the same description as before 😂

suse/distributionscanner.go
crozzy: Yeah, that'd probably be cleaner

suse/factory.go
crozzy: updated

suse/distributionscanner.go
```diff
@@ -29,38 +32,12 @@ const (
 	suseReleasePath = `etc/SuSE-release`
```
RTann, line 32: are these used anywhere? sorry just noticed this

suse/distributionscanner.go
```diff
@@ -3,13 +3,16 @@ package suse
...
+	"github.com/Masterminds/semver"
 	"github.com/quay/zlog"
 
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/indexer"
+	"github.com/quay/claircore/osrelease"
+	"github.com/quay/claircore/pkg/cpe"
 )
 
 // Suse Enterprise Server has service pack releases however their security database files are bundled together
```
RTann, line 18: also just realized this: just checking: no need to update the version, right? I don't think anything noticeable to end users changed here

suse/factory.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/quay/claircore/pkg/ovalutil"
+	"github.com/quay/zlog"
+	"golang.org/x/net/html"
+)
+
+//doc:url updater
+const base = `https://ftp.suse.com/pub/projects/security/oval/`
+
+var (
+	reELFile   = regexp.MustCompile(`suse.linux.enterprise.server.([1-9][1-9]).xml.gz`)
```
RTann, line 25: good for now. Maybe add a note to revisit this when/if 20 comes out

suse/factory.go
```diff
@@ -0,0 +1,236 @@
...
+		zlog.Info(ctx).
+			Msg("unconfigured")
+		return us, nil
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, f.base.String(), nil)
+	if err != nil {
+		return us, fmt.Errorf("suse: unable to construct request: %w", err)
+	}
+	f.mu.Lock()
+	defer f.mu.Unlock()
```
RTann, line 52: also just noticed this, sorry. What is this protecting? I'm wondering if we have to lock around the HTTP call or if this can be done later

suse/distributionscanner.go
```diff
@@ -111,11 +88,47 @@ func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]
...
+		return nil
+	}
+
+	return d
+}
+
+func cpeToDist(r cpe.WFN) (*claircore.Distribution, error) {
+	if vendor, err := cpe.NewValue("opensuse"); err == nil && r.Attr[cpe.Vendor] == vendor {
+		if prod, err := cpe.NewValue("leap"); err == nil && r.Attr[cpe.Product] == prod {
+			fmt.Println(r.Attr[cpe.Version].String())
```
RTann, line 119: remove before merge?

suse/distributionscanner.go
crozzy: yeah, the indexing should be the same I believe

suse/distributionscanner.go
crozzy: yeah, they are the files we check for during the indexing

suse/factory.go
crozzy: I believe this was ensuring that multiple calls to UpdaterSet were protecting setting the fields belonging to the Factory, but we're no longer saving etag stuff there I think it's ok to remove

suse/distributionscanner.go
crozzy: :facepalm:

suse/distributionscanner.go
```diff
@@ -111,11 +88,46 @@ func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]
...
 //
 // separated into its own method to aid testing.
-func (ds *DistributionScanner) parse(buff *bytes.Buffer) *claircore.Distribution {
-	for _, ur := range suseRegexes {
-		if ur.regexp.Match(buff.Bytes()) {
-			return releaseToDist(ur.release)
+func (ds *DistributionScanner) parse(ctx context.Context, buff *bytes.Buffer) *claircore.Distribution {
+	kv, err := osrelease.Parse(ctx, buff)
+	if err != nil {
+		zlog.Warn(ctx).Err(err).Msg("malformed os-release file")
```
RTann, line 94: `return nil` here, too? I guess the next line will cover it, too, though?

suse/distributionscanner.go
```diff
@@ -111,11 +88,46 @@ func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]
...
+	if err != nil {
+		zlog.Warn(ctx).Err(err).Msg("error converting cpe to distribution")
+		return nil
+	}
+
+	return d
+}
+
+func cpeToDist(r cpe.WFN) (*claircore.Distribution, error) {
+	if vendor, err := cpe.NewValue("opensuse"); err == nil && r.Attr[cpe.Vendor] == vendor {
```
RTann, line 117: nit: can make the CPEs globals so they're only evaluated once. Haven't looked into the function to see if it saves a ton of work or not, though

suse/factory.go
```diff
@@ -0,0 +1,234 @@
...
+	if res.StatusCode != http.StatusOK {
+		return us, fmt.Errorf("suse: unexpected status requesting OVAL dir: %v", res.Status)
+	}
+
+	// It's kind of gross having to parse the HTML but the other path,
+	// parsing the SHA256SUMs, is more fragile. At least this way we
+	// know the file exists with the expected compression.
+	dir, err := html.Parse(res.Body)
+	if err != nil {
+		log.Fatal(err)
```
RTann, line 68: just checking if this is preferred vs a normal error?

RTann (dismissed):
A few tiny things, but I think should be good to go after this. Let me know if my approval is dismissed

suse/distributionscanner.go
crozzy: My instinct is that any minimal efficiency gain would be lost in legibility:

```go
var (
	openSUSECPEVal, leapCPEVal, suseCPEVal, slesCPEVal cpe.Value
)

func init() {
	var err error
	openSUSECPEVal, err = cpe.NewValue("opensuse")
	if err != nil {
		panic(err)
	}
	leapCPEVal, err = cpe.NewValue("leap")
	if err != nil {
		panic(err)
	}
	suseCPEVal, err = cpe.NewValue("suse")
	if err != nil {
		panic(err)
	}
	slesCPEVal, err = cpe.NewValue("sles")
	if err != nil {
		panic(err)
	}
}

func cpeToDist(r cpe.WFN) (*claircore.Distribution, error) {
	if r.Attr[cpe.Vendor] == openSUSECPEVal {
		if r.Attr[cpe.Product] == leapCPEVal {
			return mkLeapDist(r.String(), r.Attr[cpe.Version].String()), nil
		}
	}
...
```
what do you think?

RTann: fine as-is 👍

RTann (approved):

PR #1433, opened 2024-11-05
vex: add non-vuln advisories to the deleted slice

crozzy:


rhel/vex/parser.go
```diff
@@ -100,7 +100,12 @@ func (u *Updater) DeltaParse(ctx context.Context, contents io.ReadCloser) ([]*cl
 		}
 	}
 	vulns := []*claircore.Vulnerability{}
-	for _, vs := range out {
+	for n, vs := range out {
+		if len(vs) == 0 {
+			// If there are no vulns for this CVE make sure we signal that
+			// it is deleted in case it once had vulns.
+			deleted = append(deleted, n)
```
RTann, line 107: nit: `continue`?

rhel/vex/parser_test.go
```diff
@@ -307,10 +307,10 @@ func TestParse(t *testing.T) {
 		expectedDeleted int
 	}{
 		{
-			name:            "six_advisories_two_deletions",
+			name:            "six_advisories_four_deletions",
 			filename:        "testdata/example_vex.jsonl",
 			expectedVulns:   546,
-			expectedDeleted: 2,
+			expectedDeleted: 4,
```
RTann, line 313: guessing there are two vulns in there which do not have `fixed` nor `known_affected`?

crozzy: Yeah, there are a couple of `samba` advisories in the `example_vex.jsonl`

rhel/vex/parser.go
crozzy: errrm, I guess it would guard against anything else being hypothetically added after appending to the vulns but then again it's hard to know if we'd want that hypothetical thing to run anyway. I guess I'll add it

RTann (approved):

PR #1439, opened 2024-11-12
EPSS enricher

daynewlee:
Epss enricher should be able to update, parse EPSS data and enrich vulnerability report with the data


daynewlee:
As of Nov 20th, we decide to include model version and score date in the enrichment data

enricher/epss/epss.go
```diff
@@ -0,0 +1,380 @@
...
+		return nil, "", fmt.Errorf("failed to fetch file: received status %d", resp.StatusCode)
+	}
+
+	gzipReader, err := gzip.NewReader(resp.Body)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to decompress file: %w", err)
+	}
+	defer gzipReader.Close()
+
+	scanner := bufio.NewScanner(gzipReader)
```
RTann, line 146: I think we should try to utilize the stdlib's `csv` package:

```
csvReader := csv.NewReader(gzipReader)
csvReader.FieldsPerRecord = 3 // Sanity check
csvReader.ReuseRecord = true // For efficiency purposes

record, err := csvReader.Read()
if err == nil {
	return nil, "", fmt.Errorf("unexpected start of csv file: %v", record)
}

// PARSE THE FIRST LINE HERE

record, err = csvReader.Read()
if err != nil {
	return nil, "", fmt.Errorf("unexpected start of csv file: %w", err)
}

if record[0] != "cve" || record[1] != "epss" || record[2] != "percentile" {
	return nil, "", fmt.Errorf("unexpected csv record format: %v", record)
}

// Any future comments will be ignored
// Another option is to keep this field unset, as any future comments may be considered unexpected, too
csvReader.Comment = '#'

var readErr error
for {
	record, readErr = csvReader.Read()
	if errors.Is(readErr, io.EOF) {
		break
	}
	if readErr != nil {
		return nil, "", fmt.Errorf("unable to read line in CSV: %w", readErr)
	}

	// DO STUFF WITH THE RECORD HERE
}
```

enricher/epss/epss.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	b, err := json.Marshal(m)
+	if err != nil {
+		return Type, nil, err
+	}
+	return Type, []json.RawMessage{b}, nil
+}
+
+func newItemFeed(record []string, headers []string, modelVersion string, scoreDate string) (driver.EnrichmentRecord, error) {
```
RTann, line 326: will review this once the other comments are resolved

enricher/epss/epss_test.go
```diff
@@ -0,0 +1,383 @@
+package epss
+
+import (
```
RTann, line 3: need to adjust import ordering here, too

enricher/epss/epss.go
```diff
@@ -0,0 +1,380 @@
...
+		e.sourceURL()
+	}
+
+	return nil
+}
+
+func (e *Enricher) FetchEnrichment(ctx context.Context, _ driver.Fingerprint) (io.ReadCloser, driver.Fingerprint, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "enricher/epss/Enricher/FetchEnrichment")
+	newUUID := uuid.New()
+	hint := driver.Fingerprint(newUUID.String())
```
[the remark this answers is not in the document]
daynewlee: I am not sure the best way to utilize fingerprint. I thought every fetch is unique so it needs a fingerprint. Do you have any insights?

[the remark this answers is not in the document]
crozzy: The updaters are going to run multiple times per day so wouldn't the data be the same in subsequent requests a lot of the time? If yes, we should use the etag/fingerprint to determine if anything has changed, and if not, don't bother processing again.

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+		}
+
+		ts := make([]string, 0, len(t))
+		for m := range t {
+			ts = append(ts, m)
+		}
+		sort.Strings(ts)
+
+		cveKey := strings.Join(ts, "_")
+		zlog.Debug(ctx).Str("cve_key", cveKey).Strs("cve", ts).Msg("generated CVE cache key")
```
crozzy, line 298: This is going to be extremely chatty, is it needed for normal operations?

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+
+	b, err := json.Marshal(m)
+	if err != nil {
+		return Type, nil, err
+	}
+	return Type, []json.RawMessage{b}, nil
+}
+
+func newItemFeed(record []string, headers []string, modelVersion string, scoreDate string) (driver.EnrichmentRecord, error) {
+	item := make(map[string]interface{}) // Use interface{} to allow mixed types
```
crozzy, line 338: Is there some reason not to use an object here:
```go
type EPSSItem struct {
    ModelVersion string
    Date string (time.Time?)
    CVE string
    EPSS float64
    Percentile float64
}
```
Or something like that?

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+	for i, value := range record {
+		// epss details are numeric values
+		if f, err := strconv.ParseFloat(value, 64); err == nil {
```
crozzy, line 341: Are there certain known values that will not parse into floats that we'd still want?

enricher/epss/epss.go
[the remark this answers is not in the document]
RTann: You can use the `etag`:

```
$ curl --head https://epss.cyentia.com/epss_scores-2024-11-26.csv.gz
HTTP/2 200 
content-type: application/gzip
content-length: 1608953
x-amz-id-2: QDADGoIt9QpBscFOp12tdT2pozPzWegwtR+04xinHHZmkpsF6GT6SgTSyEmzyBR5+27XvRJihT6xJccELkToPxUK/T0nutqU
x-amz-request-id: F64KYZMZ529HCPEK
date: Tue, 26 Nov 2024 09:09:35 GMT
last-modified: Tue, 26 Nov 2024 09:08:42 GMT
etag: "f76d77fefb8c5aff388eb47c5bba7339"
x-amz-server-side-encryption: AES256
x-amz-version-id: 3YtWn__SOrTVNKzXg2zozUdxODXLstQ4
accept-ranges: bytes
server: AmazonS3
x-cache: Hit from cloudfront
via: 1.1 40dcbda4bd91e45eab647af7e1631a5e.cloudfront.net (CloudFront)
x-amz-cf-pop: LAX53-P1
x-amz-cf-id: QR9wGoi76VQg9w2yc3YhOrOLJ2DCSdaby33NpyU6YMOt5SGwKKLp4g==
age: 55844
x-xss-protection: 1; mode=block
x-frame-options: SAMEORIGIN
referrer-policy: strict-origin-when-cross-origin
x-content-type-options: nosniff
strict-transport-security: max-age=31536000
vary: Origin
```

There's precedence for this [here](https://github.com/quay/claircore/blob/v1.5.33/alpine/fetcher.go#L15) for reference

enricher/epss/epss.go
daynewlee: I think originally `value` here was string type

enricher/epss/epss.go
```diff
@@ -0,0 +1,280 @@
...
+// This is a slightly more relaxed version of the validation pattern in the NVD
+// JSON schema: https://csrc.nist.gov/schema/nvd/feed/1.1/CVE_JSON_4.0_min_1.1.schema
+//
+// It allows for "CVE" to be case insensitive and for dashes and underscores
+// between the different segments.
+var cveRegexp = regexp.MustCompile(`(?i:cve)[-_][0-9]{4}[-_][0-9]{4,}`)
+
+const (
+	// Type is the type of data returned from the Enricher's Enrich method.
+	Type = `message/vnd.clair.map.vulnerability; enricher=clair.epss schema=https://csrc.nist.gov/schema/nvd/feed/1.1/cvss-v3.x.json`
```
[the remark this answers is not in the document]
daynewlee: Do you have any suggestions? Do we need to define a schema in the enrichment package?

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/libvuln/driver"
+	"github.com/quay/claircore/pkg/tmp"
+	"github.com/quay/zlog"
```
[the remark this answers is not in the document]
crozzy: `github.com/quay` dependencies are all internal dependencies, in this case `zlog` is external despite sharing a namespace.

enricher/epss/epss.go
```diff
@@ -0,0 +1,405 @@
...
+
+	// DefaultFeed is the default place to look for EPSS feeds.
+	// epss_scores-YYYY-MM-DD.csv.gz needs to be specified to get all data
+	DefaultFeed = `https://epss.cyentia.com/`
+
+	// epssName is the name of the enricher
+	epssName = `clair.epss`
+)
+
+func init() {
```
crozzy, line 51: I'm not sure the functionality of this

enricher/epss/epss.go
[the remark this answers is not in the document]
crozzy: It should be the JSON schema of the EPSSItem when it is marshaled but I'm not entirely sure where would be a good place for it, any ideas @hdonnay ?

enricher/epss/epss.go
```diff
@@ -0,0 +1,380 @@
...
+// Config is the configuration for Enricher.
+type Config struct {
+	FeedRoot *string `json:"feed_root" yaml:"feed_root"`
+}
+
+func (e *Enricher) Configure(ctx context.Context, f driver.ConfigUnmarshaler, c *http.Client) error {
+	ctx = zlog.ContextWithValues(ctx, "component", "enricher/epss/Enricher/Configure")
+	var cfg Config
+	e.c = c
+	if f == nil {
```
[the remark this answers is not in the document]
crozzy: Well, yeah it should never be `nil`, but I'm not sure I can argue for taking out a `nil` check :D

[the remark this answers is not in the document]
crozzy: I'd probably say return an error if `f` is nil

enricher/epss/epss.go
```diff
@@ -0,0 +1,405 @@
...
+		record, err = csvReader.Read()
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if err != nil {
+			return nil, "", fmt.Errorf("unable to read line in CSV: %w", err)
+		}
+
+		if len(record) != len(headers) {
+			zlog.Warn(ctx).Str("record", fmt.Sprintf("%v", record)).Msg("skipping record with mismatched fields")
```
crozzy, line 211: You can use `Strs()` here to avoid formatting yourself (and in the next one)

crozzy:
Most everything I brought up seems to be addressed (apart from the new comments)

enricher/epss/epss.go
```diff
@@ -0,0 +1,405 @@
...
+		if len(t) == 0 {
+			zlog.Debug(ctx).Msg("no CVEs found in vulnerability metadata")
+			continue
+		}
+
+		ts := make([]string, 0, len(t))
+		for m := range t {
+			ts = append(ts, m)
+		}
+		sort.Strings(ts)
```
RTann, line 325: nit: `slices.Sort` is preferred these days https://pkg.go.dev/sort#Strings

enricher/epss/epss.go
```diff
@@ -0,0 +1,405 @@
...
+		// Skip if no enrichment records are found
+		if len(rec) == 0 {
+			zlog.Debug(ctx).Strs("cve", ts).Msg("no enrichment records found for CVEs")
+			continue
+		}
+
+		for _, r := range rec {
+			if _, exists := m[id]; !exists {
+				m[id] = []json.RawMessage{}
+			}
```
RTann, line 350: I don't think this check is required. I think we can just do the `m[id] = append(m[id], r.Enrichment)` below (we can append something to a `nil` slice in Go)

enricher/epss/epss_test.go
```diff
@@ -0,0 +1,395 @@
...
+func mockServer(t *testing.T) *httptest.Server {
+	const root = `testdata/`
+
+	// Define a static ETag for testing purposes
+	const etagValue = `"test-etag-12345"`
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch path.Ext(r.URL.Path) {
+		case ".gz": // only gz feed is supported
+			w.Header().Set("ETag", etagValue)
```
RTann, line 173: this is typically lowercase `etag`

enricher/epss/epss.go
```diff
@@ -0,0 +1,405 @@
...
+
+	return ret, nil
+}
+
+func (*Enricher) Name() string {
+	return epssName
+}
+
+func currentFeedURL() string {
+	currentDate := time.Now()
```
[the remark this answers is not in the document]
daynewlee: I wonder if we should always pull "yesterday"'s data. It should be acceptable by the user and less possible to fail for us.

enricher/epss/epss.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	return ret, nil
+}
+
+func (*Enricher) Name() string {
+	return epssName
+}
+
+func currentFeedURL() string {
```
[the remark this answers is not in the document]
daynewlee: https://github.com/quay/claircore/pull/1439#discussion_r1872195339

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
+package epss
```
RTann, line 1: Can you add a package-level Go doc to quickly explain the purpose of this? For example, here's the CVSS enricher:

```
// Package cvss provides a cvss enricher.
package cvss
```

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+	ModelVersion string  `json:"modelVersion"`
+	Date         string  `json:"date"`
+	CVE          string  `json:"cve"`
+	EPSS         float64 `json:"epss"`
+	Percentile   float64 `json:"percentile"`
+}
+
+const (
+	// Type is the type of data returned from the Enricher's Enrich method.
+	Type = `message/vnd.clair.map.vulnerability; enricher=clair.epss schema=https://csrc.nist.gov/schema/nvd/baseURL/1.1/cvss-v3.x.json`
```
RTann, line 45: What was decided for the schema? Should we just go schema-less?

enricher/epss/epss.go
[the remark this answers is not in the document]
RTann: I'll defer this decision to @crozzy  and/or @hdonnay

enricher/epss/epss.go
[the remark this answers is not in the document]
RTann: not a bad idea. Yes, a day behind, but simpler for us. I'm ok with that

enricher/epss/epss.go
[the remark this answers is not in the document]
RTann: I'm ok with always pulling yesterday's data

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+		return err
+	}
+	if cfg.BaseURL != nil {
+		// validate the URL format
+		if _, err := url.Parse(*cfg.BaseURL); err != nil {
+			return fmt.Errorf("invalid URL format for BaseURL: %w", err)
+		}
+
+		// only .gz file is supported
+		if strings.HasSuffix(*cfg.BaseURL, ".gz") {
```
RTann, line 89: ah ok I just noticed this. So I'm a bit confused: if someone configures the URL themselves, then are we expecting them to provide the full URL or just the base? If the full path, then we should call it `URL` instead of `BaseURL`. Apologies, I missed this before

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+	defer resp.Body.Close()
+
+	if err = httputil.CheckResponse(resp, http.StatusOK); err != nil {
+		return nil, "", fmt.Errorf("unable to fetch file: %w", err)
+	}
+
+	etag := resp.Header.Get("etag")
+	if etag == "" {
+		return nil, "", fmt.Errorf("ETag not found in response headers")
```
RTann, line 134:
```suggestion
		return nil, "", errors.New("ETag not found in response headers")
```

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+
+	// assume metadata is always in the first line
+	record, err := csvReader.Read()
+	if err != nil {
+		return nil, "", fmt.Errorf("unable to read metadata line: %w", err)
+	}
+
+	var modelVersion, date string
+	for _, field := range record {
+		field = strings.TrimSpace(field)
```
RTann, line 161: This line is no longer needed

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+
+	if modelVersion == "" || date == "" {
+		return nil, "", fmt.Errorf("missing metadata fields in record: %v", record)
+	}
+	csvReader.Comment = '#'
+	csvReader.FieldsPerRecord = 3 // Expect exactly 3 fields per record
+
+	if modelVersion == "" || date == "" {
+		return nil, "", fmt.Errorf("missing metadata fields in record: %v", record)
+	}
```
RTann, line 183: already done above

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+			modelVersion = value
+		case "score_date":
+			date = value
+		}
+	}
+
+	if modelVersion == "" || date == "" {
+		return nil, "", fmt.Errorf("missing metadata fields in record: %v", record)
+	}
+	csvReader.Comment = '#'
```
RTann, line 178: nit: newline between these

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+
+	b, err := json.Marshal(m)
+	if err != nil {
+		return Type, nil, err
+	}
+	return Type, []json.RawMessage{b}, nil
+}
+
+func newItemFeed(record []string, modelVersion string, scoreDate string) (driver.EnrichmentRecord, error) {
+	// Assuming record has already been validated to have 3 fields
```
RTann, line 357:
```suggestion
	// Validate the record has the expected length
```

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+	return out, newFingerprint, nil
+}
+
+// ParseEnrichment implements driver.EnrichmentUpdater.
+func (e *Enricher) ParseEnrichment(ctx context.Context, rc io.ReadCloser) ([]driver.EnrichmentRecord, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "enricher/epss/Enricher/ParseEnrichment")
+
+	defer func() {
+		_ = rc.Close()
+	}()
```
RTann, line 235: I don't think this is needed. From what I can tell, the caller should already `Close` the `io.ReadCloser`. I do see this is called in the CVSS enricher, though... Perhaps it's ok, though maybe just do `defer rc.Close()`

Unrelated, but it's a bit odd this takes in an `io.ReadCloser`. I think it can just be an `io.Reader`, but that's outside the scope of this PR

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+			break
+		}
+		ret = append(ret, record)
+	}
+
+	zlog.Debug(ctx).
+		Int("count", len(ret)).
+		Msg("decoded enrichments")
+
+	if !errors.Is(err, io.EOF) {
```
RTann, line 253: just confirming, we expect an EOF once we exhaust all records?

enricher/epss/epss.go
daynewlee: Yeah, I agree with schema-less. Is that ok with you guys? @crozzy

enricher/epss/epss.go
daynewlee: Yeah, `URL` makes sense. Because I want to give us the flexibility to cache the EPSS data in GCS bucket and then fetched by EPSS enricher.

enricher/epss/epss.go
crozzy: sure

enricher/epss/epss.go
daynewlee: Yes, I think so.

crozzy (commented):
A couple of small things

enricher/epss/epss.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+	defer resp.Body.Close()
+
+	if err = httputil.CheckResponse(resp, http.StatusOK); err != nil {
+		return nil, "", fmt.Errorf("unable to fetch file: %w", err)
+	}
+	var str string
+	var newFingerprint driver.Fingerprint
+	str = resp.Header.Get("etag")
+	if str == "" {
```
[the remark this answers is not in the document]
daynewlee: So we would just return an empty fingerprint is etag is missing?
https://github.com/quay/claircore/pull/1439/commits/fd434d87291609043901662a2aec315591e2b02a
Let me know if this works

crozzy (approved):

PR #1440, opened 2024-11-14
gobin: ignore flags in stdlib version

RTann:
I pulled the binary from an OpenShift image (`/usr/bin/openshift-controller-manager`) and ran `TestBin` in `gobin/exe_test.go`:

```
stdlib@1.20.12 / 1.20.12
```

This is what I expected to see.

Previously, it was:

```
stdlib@go1.20.12 X:strictfipsruntime / 0
```

`gobin`'s matcher relies on PostgreSQL to handle the version comparison based on the `NormalizedVersion`. In this case, it'll be `1.20.12`, which is what we'd want/expect.

Note: This will change the package's version by removing the `X` flags, as they are not used at this time. The unique index for the `package` table is based on `(name, version, kind, module, arch)`. So, if we do not change the version, then instances of Clair which upgrade to a version with this change will not have the package's table modified to account for these changes. A solution would be to have a DB migration to potentially wipe any Go-related packages, but that seemed like overkill, as we do not utilize knowledge of the flags at this time. We will consider re-adding them at a later time should they be needed then.

### DIFF
--- a/gobin/exe.go
+++ b/gobin/exe.go
@@ -54,7 +54,11 @@ func toPackages(ctx context.Context, out *[]*claircore.Package, p string, r io.R
 	// TODO(hank) The "go version" is documented as the toolchain that produced
 	// the binary, which may be distinct from the version of the stdlib used?
 	// Need to investigate.
-	runtimeVer, err := ParseVersion(strings.TrimPrefix(bi.GoVersion, "go"))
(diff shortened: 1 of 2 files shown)


crozzy (dismissed):

RTann:
@crozzy I added some comments since this isn't exactly self-documenting. Can you take another look?

crozzy (approved):

PR #1445, opened 2024-11-22
periodic: account for differences with Konflux built images

crozzy:
For the RPM test we check the index report against the pyxis manifest for that image. The format of this manifest has recently changed when they started producing builds from Konflux, this change accounts for those changes.

### DIFF
--- a/test/rpmtest/manifest.go
+++ b/test/rpmtest/manifest.go
@@ -17,13 +17,14 @@ type Manifest struct {
 	RPM []ManifestRPM `json:"rpms"`
 }
 type ManifestRPM struct {
-	Name    string `json:"name"`
-	Version string `json:"version"`
-	Release string `json:"release"`
-	Arch    string `json:"architecture"`
-	Source  string `json:"srpm_nevra"`
-	GPG     string `json:"gpg"`
-	Module  string `json:"module"`
+	Name        string `json:"name"`
+	Version     string `json:"version"`
+	Release     string `json:"release"`
+	Arch        string `json:"architecture"`
+	SourceNEVRA string `json:"srpm_nevra"`
+	SourceName  string `json:"srpm_name"`
+	GPG         string `json:"gpg"`
+	Module      string `json:"module"`
 }
 
 func PackagesFromRPMManifest(t *testing.T, r io.Reader) []*claircore.Package {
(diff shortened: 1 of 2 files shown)


crozzy:
closes https://github.com/quay/claircore/issues/1444

test/rpmtest/manifest.go
```diff
@@ -44,15 +45,28 @@ func PackagesFromRPMManifest(t *testing.T, r io.Reader) []*claircore.Package {
...
-		if s, ok := src[rpm.Source]; ok {
+
+		// Newer images produced from Konflux shove all the source information
+		// into the SourceName and omit the SourceNEVRA. Try both.
+		var source string
+		switch {
+		case rpm.SourceNEVRA != "":
+			source = rpm.SourceNEVRA
+		case rpm.SourceName != "":
+			source = rpm.SourceName
```
RTann, line 56: is it possible for both to be populated?

test/rpmtest/manifest.go
```diff
@@ -44,15 +45,28 @@ func PackagesFromRPMManifest(t *testing.T, r io.Reader) []*claircore.Package {
...
+		default:
+			continue
+		}
+
+		if s, ok := src[source]; ok {
 			p.Source = s
 		} else {
-			s := strings.TrimSuffix(rpm.Source, ".src")
+			s := strings.TrimSuffix(strings.TrimSuffix(source, ".rpm"), ".src")
```
RTann, line 64: so it's either `.src` or `.src.rpm`?

test/rpmtest/manifest.go
```diff
@@ -44,15 +45,28 @@ func PackagesFromRPMManifest(t *testing.T, r io.Reader) []*claircore.Package {
...
 			pos := len(s)
 			for i := 0; i < 2; i++ {
 				pos = strings.LastIndexByte(s[:pos], '-')
 				if pos == -1 {
-					t.Fatalf("malformed NEVRA: %q", rpm.Source)
+					t.Fatalf("malformed NEVRA: %q for %q", source, rpm.Name)
```
RTann, line 69: nit: may not be NEVRA

test/rpmtest/manifest.go
crozzy: Yes, older manifests will have `SourceNEVRA` populated, newer ones won't but they have the same information in `SourceName`.

test/rpmtest/manifest.go
crozzy: yeah, from the data the `SourceNEVRA` will have the `.rpm` , `SourceName` will have `.src.rpm`

test/rpmtest/manifest.go
crozzy: We're expecting something like either 
```
acl-2.3.1-4.el9.src.rpm // SourceName
acl-0:2.3.1-3.el9.src      // SourceNEVRA
```
So I guess yeah, one is NEVRA and one is NVRA

RTann (approved):

PR #1449, opened 2024-12-09
datastore: fix get query caching

crozzy:
This change replaces caching the DB pgx.Rows, which is an iterator, with the actual potential vulnerabilities. Previously, subsequent records manifesting in the same query were given a pgx.Rows object that had been exhausted.

### DIFF
--- a/datastore/postgres/get.go
+++ b/datastore/postgres/get.go
@@ -51,7 +51,7 @@ func (s *MatcherStore) Get(ctx context.Context, records []*claircore.IndexRecord
 	defer tx.Rollback(ctx)
 	// start a batch
 	batch := &pgx.Batch{}
-	resCache := map[string]pgx.Rows{}
+	resCache := map[string][]*claircore.Vulnerability{}
 	rqs := []*recordQuery{}
 	for _, record := range records {
 		query, err := buildGetQuery(record, &opts)
@@ -72,7 +72,6 @@ func (s *MatcherStore) Get(ctx context.Context, records []*claircore.IndexRecord
 		resCache[query] = nil
 	}
 	// send the batch
-
 	start := time.Now()
 	res := tx.SendBatch(ctx, batch)
 	// Can't just defer the close, because the batch must be fully handled
@@ -83,17 +82,28 @@ func (s *MatcherStore) Get(ctx context.Context, records []*claircore.IndexRecord
 	results := make(map[string][]*claircore.Vulnerability)
 	vulnSet := make(map[string]map[string]struct{})
 	for _, rq := range rqs {
-		rows, ok := resCache[rq.query]
+		rid := rq.record.Package.ID
+		vulns, ok := resCache[rq.query]
 		if !ok {
 			return nil, fmt.Errorf("unexpected vulnerability query: %s", rq.query)
 		}
-		if rows == nil {
-			rows, err = res.Query()
-			if err != nil {
-				res.Close()
-				return nil, err
+		if vulns != nil { // We already have results we don't need to go back to the DB.
+			if _, ok := vulnSet[rid]; !ok {
+				vulnSet[rid] = make(map[string]struct{})
 			}
-			resCache[rq.query] = rows
+			for _, v := range vulns {
+				if _, ok := vulnSet[rid][v.ID]; !ok {
+					vulnSet[rid][v.ID] = struct{}{}
+					results[rid] = append(results[rid], v)
+				}
+			}
+			continue
+		}
+		resCache[rq.query] = []*claircore.Vulnerability{}
+		rows, err := res.Query()
+		if err != nil {
+			res.Close()
+			return nil, fmt.Errorf("error getting rows: %w", err)
 		}
 
 		// unpack all returned rows into claircore.Vulnerability structs
@@ -140,7 +150,6 @@ func (s *MatcherStore) Get(ctx context.Context, records []*claircore.IndexRecord
 				return nil, fmt.Errorf("failed to scan vulnerability: %v", err)
 			}
 
-			rid := rq.record.Package.ID
 			if _, ok := vulnSet[rid]; !ok {
 				vulnSet[rid] = make(map[string]struct{})
 			}
@@ -149,6 +158,7 @@ func (s *MatcherStore) Get(ctx context.Context, records []*claircore.IndexRecord
 				results[rid] = append(results[rid], v)
 			}
 		}
+		resCache[rq.query] = results[rid]
 	}
 	if err := res.Close(); err != nil {
 		return nil, fmt.Errorf("some weird batch error: %v", err)


crozzy:
Post-change:
```json
    "186": {
      "id": "186",
      "name": "xz-libs",
      "version": "5.2.2-1.el7",
      "kind": "binary",
      "source": {
        "id": "185",
        "name": "xz",
        "version": "5.2.2-1.el7",
        "kind": "source"
      },
      "arch": "x86_64"
    },
...
    "1924": {
      "id": "1924",
      "name": "xz-libs",
      "version": "5.2.2-1.el7",
      "kind": "binary",
      "source": {
        "id": "185",
        "name": "xz",
        "version": "5.2.2-1.el7",
        "kind": "source"
      },
      "arch": "i686"
    },
...
    "2675687": {
      "id": "2675687",
      "updater": "rhel-vex",
      "name": "CVE-2022-1271",
      "description": "An arbitrary file write vulnerability was found in GNU gzip's zgrep utility. When zgrep is applied on the attacker's chosen file name (for example, a crafted file name), this can overwrite an attacker's content to an arbitrary attacker-selected file. This flaw occurs due to insufficient validation when processing filenames with two or more newlines where selected content and the target file names are embedded in crafted multi-line file names. This flaw allows a remote, low privileged attacker to force zgrep to write arbitrary files on the system.",
      "issued": "2022-04-07T00:00:00Z",
      "links": "https://access.redhat.com/security/cve/CVE-2022-1271 https://bugzilla.redhat.com/show_bug.cgi?id=2073310 https://www.cve.org/CVERecord?id=CVE-2022-1271 https://nvd.nist.gov/vuln/detail/CVE-2022-1271 https://security.access.redhat.com/data/csaf/v2/vex/2022/cve-2022-1271.json https://access.redhat.com/errata/RHSA-2022:5052",
      "severity": "CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:H",
      "normalized_severity": "High",
      "package": {
        "id": "",
        "name": "xz-libs",
        "version": "",
        "kind": "binary",
        "arch": "i686|ppc|ppc64|ppc64le|s390|s390x|amd64|x86_64"
      },
      "distribution": {
        "id": "",
        "did": "",
        "name": "",
        "version": "",
        "version_code_name": "",
        "version_id": "",
        "arch": "",
        "cpe": "",
        "pretty_name": ""
      },
      "repository": {
        "name": "cpe:2.3:o:redhat:enterprise_linux:7:*:server:*:*:*:*:*",
        "key": "rhel-cpe-repository",
        "cpe": "cpe:2.3:o:redhat:enterprise_linux:7:*:server:*:*:*:*:*"
      },
      "fixed_in_version": "0:5.2.2-2.el7_9",
      "arch_op": "pattern match"
    },
...
"package_vulnerabilities": {
    "186": [
      "2675687"
    ],
...
    "1924": [
      "2675687"
    ],
}
```

hdonnay (approved):

crozzy:
@hdonnay sorry just had to update

datastore/postgres/get.go
```diff
@@ -83,17 +82,28 @@ func (s *MatcherStore) Get(ctx context.Context, records []*claircore.IndexRecord
...
+			for _, v := range vulns {
+				if _, ok := vulnSet[rid][v.ID]; !ok {
+					vulnSet[rid][v.ID] = struct{}{}
+					results[rid] = append(results[rid], v)
+				}
+			}
+			continue
+		}
+		results[rid] = []*claircore.Vulnerability{}
+		rows, err := res.Query()
```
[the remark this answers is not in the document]
crozzy: `pgx.Rows` should close at the end of `Next`ing:

```
// Rows is the result set returned from *Conn.Query. Rows must be closed before
// the *Conn can be used again. Rows are closed by explicitly calling Close(),
// calling Next() until it returns false, or when a fatal error occurs.
```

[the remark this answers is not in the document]
RTann: The docs I found says this:

```
// Next prepares the next row for reading. It returns true if there is another
// row and false if no more rows are available. It automatically closes rows
// when all rows are read.
Next() bool

// Scan reads the values from the current row into dest values positionally.
// dest can include pointers to core types, values implementing the Scanner
// interface, and nil. nil will skip the value entirely. It is an error to
// call Scan without first calling Next() and checking that it returned true.
Scan(dest ...interface{}) error
```

So I get the happy case will be good, but if `Scan` fails, I don't thinks `rows` will automatically close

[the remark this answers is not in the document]
crozzy: If `Scan` fails, Rows should be implicitly closed. You can see that [here](https://github.com/jackc/pgx/blob/v4.18.3/rows.go#L197) and [here](https://github.com/jackc/pgx/blob/v4.18.3/rows.go#L173) . It is possible to check by creating a Scan error and trying to call `Values()`:

```go
			if err != nil {
				res.Close()
				_, valErr := rows.Values()
				return nil, fmt.Errorf("failed to scan vulnerability: %v and got a Values error %v", err, valErr)
			}
// failed to start scan: matcher error: failed to scan vulnerability: can't scan into dest[25]: unable to assign to *int64 and got a Values error rows is closed
```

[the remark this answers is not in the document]
crozzy: I'm fine deferring a `rows.Close()`  in an anonymous function so we avert any change in behavior from the library, just pointing out that it hasn't existed in this function up until this point because it looks like the rows are always closed by the library.

RTann (dismissed):
Seems like `rows` does seem to close on its own. LGTM, though I would not be opposed to throwing in a `defer rows.Close`, if desired

crozzy:
> Seems like `rows` does seem to close on its own. LGTM, though I would not be opposed to throwing in a `defer rows.Close`, if desired

Updated, could probably do with another once over

datastore/postgres/get.go
```diff
@@ -83,72 +82,89 @@ func (s *MatcherStore) Get(ctx context.Context, records []*claircore.IndexRecord
...
+					&v.FixedInVersion,
+					&v.Updater,
+				)
+				v.ID = strconv.FormatInt(id, 10)
+				if err != nil {
+					res.Close()
+					return fmt.Errorf("failed to scan vulnerability: %w", err)
+				}
+
+				if _, ok := vulnSet[rid]; !ok {
```
RTann, line 154: I realize this was here before but I'm wondering: is it possible we see the same `rid` twice in the `rqs` loop and they both have different queries and different sets of vulnerabilities? If so, then it seems like, in the end, we give them both the same set of vulns instead of separate. Is this expected? If so, when?

crozzy: Yes, I think this is a dormant bug, it probably never comes up in practice because:

1. With VEX the repository doesn't matter until `Vulnerable()` is called (so after the querying) so the queries per package are the same.
2. Before that, in the OVAL files, each vulnerability referenced all the possible repos so I suppose you were bound to get a comparable set of Vulns for each package.ID anyway.

I think this probably deserves more discussion and a separate issue.

crozzy: 3. Every other time a matcher queries on `driver.RepositoryName`, it's a mono-repository representing that ecosytem

RTann: So I looked into it a bit, and I realize that's supposed to happen due to the nature of IndexRecords. As in, there is a record for every unique for every repository in every environment in every package. So, there should be an example where we see the same `rid` with different queries (for example two different potential repositories) and it's possible we get two different results. At least in this particular example that I'm thinking of (rpms), this was something which has popped up and was a bug, but I believe switching to VEX resolved it? Or, it's still a problem because we have to guess which repository is responsible for the found rpms (we try them all and see what happens)

RTann (approved):

datastore/postgres/get.go
crozzy: Yeah, it's not an issue for VEX because the specific repo isn't part of the query anymore, it's checked after the vulns are returned. But, it's still a little bit of a bear trap for future stuff.

PR #1456, opened 2025-01-02
osrelease: handle completely empty keys

hdonnay:
We hadn't run into this before. This seems like the sensible action to take when there's no value.


crozzy (approved):
LGTM

hdonnay:
Will merge with green CI

PR #1460, opened 2025-01-03
migrations: add missing index creating migration

crozzy:
While the sql command was added to claircore (and the admin command was added to clairctl) to create an index it was never actually reference in the migrations. This should be a no-op for all instances that ran the associated clairctl admin command. Also, added a test to try and avoid this in the future.

### DIFF
--- a/datastore/postgres/migrations/migrations_test.go
+++ b/datastore/postgres/migrations/migrations_test.go
@@ -1,0 +1,68 @@
+package migrations
+
+import (
+	"bufio"
+	iofs "io/fs"
+	"os"
+	"path"
+	"path/filepath"
+	"regexp"
+	"slices"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestMigrationsMismatch(t *testing.T) {
+	var migrations, files []string
+
+	// Get referenced migrations
+	migrationLine, err := regexp.Compile(`runFile\(\"(.*)\"\)`)
+	if err != nil {
+		t.Fatal(err)
+	}
+	f, err := os.Open("migrations.go")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+
+	s := bufio.NewScanner(f)
+	for s.Scan() {
+		ms := migrationLine.FindSubmatch(s.Bytes())
+		switch {
+		case ms == nil, len(ms) == 1:
+			continue
+		case len(ms) == 2:
+			migrations = append(migrations, path.Clean(string(ms[1])))
+		}
+	}
+	if err := s.Err(); err != nil {
+		t.Error(err)
+	}
+	slices.Sort(migrations)
+
+	// Get migration files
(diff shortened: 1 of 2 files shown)


datastore/postgres/migrations/migrations_test.go
```diff
@@ -0,0 +1,67 @@
...
+		files = append(files, p)
+		return nil
+	})
+	if err != nil {
+		t.Error(err)
+	}
+	sort.Strings(files)
+
+	// Check referenced files exist and existing files are referenced.
+	if !cmp.Equal(migrations, files) {
```
RTann, line 63: I believe this will fail on Windows machines due to the `\`? If so, is that ok?

datastore/postgres/migrations/migrations_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestMigrationsMismatch(t *testing.T) {
+	var migrations, files []string
+
+	// Get referenced migrations
+	migrationLine, err := regexp.Compile(`runFile\(\"(.*)\"\)`)
```
RTann, line 19: This feels brittle. An idea can be to create a base slice like

```
const indexerMigrations = []string{
	"filename1",
	"filename2",
	...
}
```

and then have

```
func init() {
	for i, migration := range indexerMigrations {
		IndexerMigrations = append(IndexerMigrations, {
			ID: i+1,
			Up: runFile(migration),
		})
	}
}
```

something like that so we can just ensure `indexerMigrations` lists all the files

datastore/postgres/migrations/migrations_test.go
```diff
@@ -0,0 +1,67 @@
...
+	defer f.Close()
+
+	s := bufio.NewScanner(f)
+	for s.Scan() {
+		ms := migrationLine.FindSubmatch(s.Bytes())
+		switch {
+		case ms == nil, len(ms) == 1:
+			continue
+		case len(ms) == 2:
+			migrations = append(migrations, path.Clean(string(ms[1])))
```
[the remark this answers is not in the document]
crozzy: I will change it but yeah

datastore/postgres/migrations/migrations_test.go
crozzy: We don't support running on Windows so I think it's ok

datastore/postgres/migrations/migrations_test.go
crozzy: I think one thing we'd want to keep is the ID being explicitly associated with the migration file so a user can check their DB migration version and know which was the last file being run (without having to work it out).

It is brittle in the sense that the runFile might change, but the proposed solution doesn't address fact that a sql file might exist on the FS but not be referenced in the go code, maybe this isn't actually an issue if you aren't forgetful (like myself) and remember to reference the migration. The reason this case was slightly different was that there was an admin command added to `clairctl` to pre-create the index so when testing the code in stage and production the index was already create and hence, it didn't look like anything was missing.

hdonnay (approved):

PR #1462, opened 2025-01-06
epss: add NewFactory()

daynewlee:


crozzy:
How are you calling this? Distinct from how [Clair instantiates](https://github.com/quay/clair/blob/main/initialize/services.go#L289) it?

daynewlee:
> How are you calling this? Distinct from how [Clair instantiates](https://github.com/quay/clair/blob/main/initialize/services.go#L289) it?

Something like this
```
updates.WithFactories(map[string]driver.UpdaterSetFactory{
			"epss": epss.NewFactory(),
		}),
```

crozzy (approved):

RTann:
From what I can tell, it's not required, as I believe you can just do `WithOutOfTree(epss.UpdaterSet())` but then the configuration is a bit awkward (you'd have to map the configuration with the name `outOfTree`)

PR #1465, opened 2025-01-08
rhel: don't consider hyphen as valid in dockerfile ENV var names

crozzy:
This change means label values like `"$PRODNAME-$COMPNAME-container"` will be expanded to something like `"rhdh-hub-container"` instead of `""` which I think is preferred as the security data refers to "rhdh-hub-container" as the package_name.


hdonnay (approved):
LGTM

Looking through the POSIX spec, there's no reason it's disallowed in an environment variable, but experimentally shell tokenization happens there.

PR #1466, opened 2025-01-10
csaf: read aggregate_severity

RTann:
I am interested in obtaining the aggregate severity of a CSAF advisory. For example: https://security.access.redhat.com/data/csaf/v2/advisories/2024/rhsa-2024_0024.json

It has an aggregate severity of `Important`. Without this change, I'd have to look at the severity rating of each vulnerability and find the highest (assuming that's how the aggregate is determined 100% of the time). With this change, it's much simpler to obtain.


hdonnay (changes requested):
Code changes are good, could you please reword the commit?

"Track" implies it's getting recorded somewhere, where this is just adding the required types.

hdonnay (approved):

PR #1468, opened 2025-01-14
spdx: initial sbom framwork + spdx encoder

BradLugo:
## Description

These changes introduce the initial SBOM framework into claircore, generating an SPDX 2.3 JSON document from a `claircore.IndexReport`.

### Design

Design doc: https://docs.google.com/document/d/1dRKMWjmkpYO5oN_Y-S_0R2vwyvcdXHCNejM3C-F2DIc/edit?tab=t.0


hdonnay:
Should be moved out of the `pkg` directory.

<details><summary>Rambling on why<summary>
The `pkg` in a package name is just line noise: we all know it's a go package. Marking something for "external" use is done in the opposite direction: `internal` is actually enforced by the compiler. If the API for use is squishier than that, it's on humans to observe it anyway.
</details>

BradLugo:
If anyone has a suggestion on testing something that bakes in `time.Now()` the output, please let me know. Example of the problem in my current unit test:
```
            - 		"created":  string("2025-01-14T21:48:35Z"),
            + 		"created":  string("2025-01-14T09:51:40Z"),
```

hdonnay:
> If anyone has a suggestion on testing something that bakes in `time.Now()` the output, please let me know.

You can put together a `cmp.Option` for [`cmp`](https://pkg.go.dev/github.com/google/go-cmp/cmp) that turns these back into `time.Time` and allows some delta, or just ignores them.

sbom/sbom.go
```diff
@@ -0,0 +1,20 @@
...
+
+type Encoder interface {
+	Encode(ctx context.Context, ir *claircore.IndexReport) (io.Reader, error)
+}
+
+type Decoder interface {
+	Decode(ctx context.Context, r io.Reader) (*claircore.IndexReport, error)
+}
+
+// TODO(DO NOT MERGE): Probably don't need this anymore?
```
crozzy, line 17: Yeah, we should remove this from the API

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,301 @@
...
+		// associated with two different Ecosystems and one of those Ecosystems
+		// doesn't have the RepositoryScanner. If something like that happens,
+		// we'll have the Repository information in another IndexRecord.
+		if r.Repository == nil || r.Repository.ID == "" {
+			continue
+		}
+
+		pkg, ok := pkgMap[r.Package.ID]
+
+		// TODO(DO NOT MERGE): Is there a way to distinguish "APPLICATION" and "LIBRARY" packages? Probably not but
```
crozzy, line 138: IMO if we are talking about distinguishing between language packages and package manager packages, then it's moot because we can know definitively from the index report IIRC

BradLugo: Sorry, forgot the remove this comment after we discussed it the other day.

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,299 @@
...
+		if err := convert.Document(spdx, targetDoc); err != nil {
+			return nil, err
+		}
+		tmpConverterDoc = targetDoc
+	case V2_2:
+		var targetDoc v2_2.Document
+		if err := convert.Document(spdx, targetDoc); err != nil {
+			return nil, err
+		}
+		tmpConverterDoc = targetDoc
```
dcaravel, line 73: nit: Don't know of any plans to support generation for 2.1 or 2.2 SPDX docs, if keeping these conversions around consider adding tests.

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,299 @@
...
+
+	switch e.Format {
+	case JSON:
+		// TODO(DO NOT MERGE): Should this be outside the switch? i.e., should
+		//  we use this for all formatting cases? We could return it outside the
+		//  switch as the "default" case
+		buf := &bytes.Buffer{}
+		if err := spdxjson.Write(tmpConverterDoc, buf); err != nil {
+			return nil, err
+		}
```
dcaravel, line 89: With the error handling IMHO it's OK either way, inside the switch or outside. It does seem the most common format is `json`, making that the 'default' could make consumption a tiny bit easier.

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,299 @@
...
+		spdxCreators[i].CreatorType = creator.CreatorType
+	}
+
+	// Initial metadata
+	out := &v2_3.Document{
+		SPDXVersion:    v2_3.Version,
+		DataLicense:    v2_3.DataLicense,
+		SPDXIdentifier: "DOCUMENT",
+		// TODO(DO NOT MERGE): Is this ok?
+		DocumentName:      ir.Hash.String(),
```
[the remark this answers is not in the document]
BradLugo: 👍 I like the idea of adding the field to the Encoder.

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,299 @@
...
+
+		pkg, ok := pkgMap[r.Package.ID]
+
+		// Record the package if we haven't seen it yet.
+		if !ok {
+			pkgDB := ""
+			for _, env := range ir.Environments[r.Package.ID] {
+				if env.PackageDB != "" {
+					pkgDB = env.PackageDB
+				}
```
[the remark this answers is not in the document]
BradLugo: I suppose it _technically_ possible for a package to have multiple environments. Maybe there's a larger discussion here about that case. What'd you think @hdonnay @crozzy? I'm going to add the `break` for now.

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,299 @@
...
+		if r.Repository != nil {
+			repo, ok := repoMap[r.Repository.ID]
+			if !ok {
+				extRefs := []*v2_3.PackageExternalReference{
+					{
+						Category: "SECURITY",
+						RefType:  "cpe23Type",
+						Locator:  r.Repository.CPE.String(),
+					},
+				}
```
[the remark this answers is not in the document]
BradLugo: Nope. Good catch.

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,299 @@
...
+							Category: "OTHER",
+							RefType:  "version_id",
+							Locator:  r.Distribution.VersionID,
+						},
+						{
+							Category: "OTHER",
+							RefType:  "pretty_name",
+							Locator:  r.Distribution.PrettyName,
+						},
+					},
```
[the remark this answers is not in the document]
BradLugo: According to the database, no.

BradLugo:
I'll start using fixup commits from this point on to help reviewers until I get approvals, then I'll rebase all the fixup commits.

dcaravel (approved):
LGTM

sbom/spdx/encoder_test.go
```diff
@@ -0,0 +1,89 @@
...
+			var got map[string]interface{}
+			if err := json.NewDecoder(r).Decode(&got); err != nil {
+				t.Error(err)
+			}
+
+			// TODO(DO NO MERGE): This feels terrible
+			ignoreCreatedTimestamp := cmp.FilterPath(func(p cmp.Path) bool {
+				sf, ok := p.Index(3).(cmp.MapIndex)
+				return ok && sf.String() == `["created"]`
+			}, cmp.Ignore())
```
BradLugo, line 82: @hdonnay is this what you were thinking? Is there a cleaner way to do it?

hdonnay (changes requested):

sbom/sbom.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+	"github.com/quay/claircore"
+	"io"
+)
+
+// Encoder is an interface to convert a claircore.IndexReport into an io.Reader
+// that contains a Software Bill of Materials representation.
+type Encoder interface {
+	Encode(ctx context.Context, ir *claircore.IndexReport) (io.Reader, error)
```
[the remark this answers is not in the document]
RTann: I'm wondering if we should consider having this match https://pkg.go.dev/encoding/json#Encoder (ie accept `io.Writer` instead of outputting `io.Reader` or `io.ReadCloser`

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,377 @@
...
+
+const (
+	V2_3 Version = "v2.3"
+)
+
+type Format string
+
+const JSON Format = "json"
+
+type Creator struct {
```
RTann, line 33: Is this required? There is already https://github.com/spdx/tools-golang/blob/v0.5.5/spdx/v2/common/creation_info.go

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,377 @@
...
+	return nil, fmt.Errorf("unknown requested format: %v", e.Format)
+}
+
+func (e *Encoder) parseIndexReport(ctx context.Context, ir *claircore.IndexReport) (*v2_3.Document, error) {
+	creatorInfo := e.Creators
+	spdxCreators := make([]v2common.Creator, len(creatorInfo))
+	for i, creator := range creatorInfo {
+		spdxCreators[i].Creator = creator.Creator
+		spdxCreators[i].CreatorType = creator.CreatorType
+	}
```
RTann, line 88:
```suggestion
	creators := make([]v2common.Creator, len(e.Creators))
	for i, creator := range e.Creators {
		creators[i] = Creator{
			Creator: creator.Creator,
			CreatorType: creator.CreatorType,
		}
	}
```

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,377 @@
...
+	// Initial metadata
+	out := &v2_3.Document{
+		SPDXVersion:       v2_3.Version,
+		DataLicense:       v2_3.DataLicense,
+		SPDXIdentifier:    "DOCUMENT",
+		DocumentName:      e.DocumentName,
+		DocumentNamespace: e.DocumentNamespace,
+		CreationInfo: &v2_3.CreationInfo{
+			Creators: spdxCreators,
+			Created:  time.Now().Format("2006-01-02T15:04:05Z"),
```
RTann, line 99: Is there a specific format that's required or was this chosen? Does this happen to match any of the builtin RFC times?

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,377 @@
...
+			return nil, err
+		}
+
+		pkg, ok := pkgMap[rPkgId]
+		// Record the package if we haven't seen it yet.
+		if !ok {
+			pkgDB := ""
+			for _, env := range ir.Environments[r.Package.ID] {
+				if env.PackageDB != "" {
+					pkgDB = env.PackageDB
```
RTann, line 127: mentioned offline, but this has a prefix, which we may want to consider removing and putting elsewhere

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,377 @@
...
+			}
+			pkgRels[rPkgId] = append(pkgRels[rPkgId], rel)
+		}
+	}
+
+	// Now that we have all the data necessary to create the SPDX document,
+	// we need to order it since the IndexRecords aren't in a deterministic order.
+	// This is particular helpful for testing, but it wouldn't be unreasonable
+	// for a user to want to diff different versions of an SPDX of the same IndexReport.
+	sort.Ints(pkgIds)
```
RTann, line 235:
```suggestion
	slices.Sort(pkgIds)
```

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,377 @@
...
+
+		pkg := pkgMap[id]
+		out.Packages = append(out.Packages, pkg)
+
+		rels := pkgRels[id]
+		slices.SortFunc(rels, cmpRelationship)
+		out.Relationships = append(out.Relationships, rels...)
+	}
+
+	sort.Ints(distIds)
```
RTann, line 249:
```suggestion
	slices.Sort(distIds)
```

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,377 @@
...
+	for _, id := range distIds {
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
+
+		dist := distMap[id]
+		out.Packages = append(out.Packages, dist)
+	}
+
+	sort.Ints(repoIds)
```
RTann, line 259:
```suggestion
	slices.Sort(repoIds)
```

sbom/spdx/encoder.go
BradLugo: I was thinking decoupling the spdx version with this information might be helpful. For example, the package you linked is specifically for spdx v2. I don't have a strong preference though.

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,377 @@
...
+var _ sbom.Encoder = (*Encoder)(nil)
+
+type Encoder struct {
+	Version           Version
+	Format            Format
+	Creators          []Creator
+	DocumentName      string
+	DocumentNamespace string
+	DocumentComment   string
+}
```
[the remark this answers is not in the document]
BradLugo: I've gone back and forth on this. Many of the zero values aren't to spec, but disallowing empty strings felt weird. One idea I toyed around with was having a `NewDefaultEncoder()` that created an `Encoder` with some claircore values, e.g., `Creator[0].CreatorType = "Tool", Creator[0].Creator = "claircore"`. What do you think? (cc @RTann)

sbom/spdx/encoder.go
BradLugo: Here's what the spec says:
> The date is to be specified according to combined date and time in UTC format as specified in ISO 8601 standard.

sbom/spdx/encoder.go
BradLugo: I'm open to suggestions. This seemed like a _fine_ alternative in lieu of pURLs.

sbom/spdx/encoder.go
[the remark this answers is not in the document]
BradLugo: @dcaravel, can I get your opinion on this as well?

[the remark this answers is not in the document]
dcaravel: &nbsp; | &nbsp;
--- | ----
`Version` | I'd be OK with a default, would need to be configurable by the client in the future as multiple formats/version are supported. 
`Format` | same ^^
`Creators` | be OK adding claircore as a default `Tool` but must have the option to add other tools, people, organizations, etc. that claircore will not know about. Adding a claircore version by default would likely require adding and maintaining a file in the source repo that is updated with each commit (if not already existing), which could be a burden. 
`DocumentName` | current state I see the client having to provide this - not sure how a default value could be derived - an index report does not contain a public friendly identifier that would be meaningful to a client/users (at least in the ACS case), but it's also contextual: for an image this could be the image digest, or the full image ref - for a node this could be the node name or IP, etc.
`DocumentComment` | same ^, this may be the only field available for 'extra context' from a client, i'd be OK with a generic default value, however the client should be able to override the value - as an example, for ACS we're planning on adding a note to indicate this is tech preview.
`DocumentNamespace` | same ^, leaning toward making this clients choice as well, the spec says this should be an _"unambiguous mechanism for other SPDX documents to reference SPDX elements within this SPDX document"_, which the index report does not have that context, and that unique addressability may not apply to type 'Analyzed' SBOMs either. The example from the [spec](https://spdx.github.io/spdx-spec/v2-draft/document-creation-information/#652-intent): `https://[CreatorWebsite]/[pathToSpdx]/[DocumentName]-[UUID]` hints this may be associated with some external storage/path/etc. which also is not something claircore knows about or is in index report. Am therefore leaning toward putting this on the client to specify.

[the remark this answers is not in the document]
BradLugo: Just pushed an update for this. @dcaravel, what'd you think?

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+			if m.Replace != nil && m.Replace.Version != m.Version {
+				core = m.Replace.Version
+			}
+		}
+	}
+	if core == "" {
+		core = "unknown revision"
+	}
+	return core
+}
```
dcaravel, line 417: Clever 👍 , can confirm it works with the `go.mod` `replace` directive as well, neat.

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+func NewDefaultEncoder(documentName, documentNamespace, documentComment string) *Encoder {
+	return &Encoder{
+		Version: V2_3,
+		Format:  JSON,
+		Creators: []Creator{
+			{
+				Creator:     "Tool",
+				CreatorType: fmt.Sprintf("Claircore-%s", getVersion()),
+			},
+		},
```
dcaravel, line 61:
```suggestion
		Creators: []Creator{
			{
				Creator:     fmt.Sprintf("Claircore-%s", getVersion()),
				CreatorType: "Tool",
			},
		},
```

sbom/spdx/encoder.go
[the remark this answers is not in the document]
dcaravel: Nice, left a review below.  

super nit and possibly a personal preference: Setting the doc name, namespace, comment on the Encoder object seemed easier to read vs. strings passed to `NewDefaultEncoder(string, string, string)` (I had put them in the wrong order initially).  Perhaps a good candidate for functional options? - but that may be over complicating it.

sbom/spdx/encoder.go
BradLugo: Credit goes to @crozzy 🙂 https://github.com/crozzy/claircore/blob/ad4e718aa3728b16f9977556a8f05d7cb6ec94d0/pkg/sbom/spdx/spdx.go#L305-L325

sbom/spdx/encoder.go
RTann: Sounds fair

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+				CreatorType: "Tool",
+			},
+		},
+		DocumentName:      documentName,
+		DocumentNamespace: documentNamespace,
+		DocumentComment:   documentComment,
+	}
+}
+
+// Encode encodes a [claircore.IndexReport] to an [io.Reader].
```
RTann, line 68: not anymore 😄

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	// TODO(blugo): support SPDX versions before 2.3
+	var tmpConverterDoc common.AnyDocument
+	switch e.Version {
+	case V2_3:
+		// parseIndexReport currently returns a v2_3.Document so do nothing
+		tmpConverterDoc = spdx
+	default:
+		return fmt.Errorf("unknown SPDX version: %v", e.Version)
```
RTann, line 85:
```suggestion
		return fmt.Errorf("unknown SPDX version: %w", e.Version)
```

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+
+	switch e.Format {
+	case JSON:
+		if err := spdxjson.Write(tmpConverterDoc, w); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	return fmt.Errorf("unknown requested format: %v", e.Format)
```
RTann, line 96:
```suggestion
	return fmt.Errorf("unknown requested format: %w", e.Format)
```

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+		DocumentName:      e.DocumentName,
+		DocumentNamespace: e.DocumentNamespace,
+		CreationInfo: &v2_3.CreationInfo{
+			Creators: creators,
+			Created:  time.Now().Format("2006-01-02T15:04:05Z"),
+		},
+		DocumentComment: e.DocumentComment,
+	}
+
+	pkgMap := map[int]*v2_3.Package{}
```
RTann, line 122: won't block the PR but I tend to prefer `pkgMap := make(map[int]*v2_3.Package)` though I do see this pattern throughout the Claircore codebase

Also maybe rename to `pkgs`?

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+	var pkgIds []int
+	distMap := map[int]*v2_3.Package{}
```
RTann, line 124: nit: `dists`?

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+	var distIds []int
+	repoMap := map[int]*v2_3.Package{}
```
RTann, line 126: nit: `repos`

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+		extRefs = append(extRefs, &v2_3.PackageExternalReference{
+			Category: "OTHER",
+			RefType:  "pretty_name",
+			Locator:  d.PrettyName,
+		})
+	}
+
+	dist := &v2_3.Package{
+		PackageName:               d.Name,
+		PackageSPDXIdentifier:     v2common.ElementID("Distribution-" + d.ID),
```
RTann, line 337: earlier there is a call like `fmt.Sprintf()` similar to this. Is there a reason for the change (I also may prefer the `+`)

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+		PackageSummary:            "distribution",
+		PrimaryPackagePurpose:     "OPERATING-SYSTEM",
+	}
+
+	return dist
+}
+
+func newSpdxPackageFromRepository(r *claircore.Repository) *v2_3.Package {
+	var extRefs []*v2_3.PackageExternalReference
+	if r.CPE.String() != "" {
```
RTann, line 351: the previous function has a newline. Can we either add one here or remove from the other for consistency?

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	return strings.Compare(a.Relationship, b.Relationship)
+}
+
+// getVersion will attempt to read out the current binary's debug info, find the
+// claircore version (this was copied from Clair).
+func getVersion() string {
+	info, infoOK := debug.ReadBuildInfo()
+	var core string
```
RTann, line 401: I think flip the order of these

sbom/spdx/encoder_test.go
```diff
@@ -0,0 +1,91 @@
+package spdx
+
+import (
```
RTann, line 3: similar import comment

sbom/spdx/encoder.go
BradLugo: Could you explain this suggestion? `e.Version` isn't an error, so it's unclear to me why `%w` is preferred.

dcaravel (dismissed):
LGTM, tested on a few images (namely `nginx:1.24.0` and `quay.io/rhacs-eng/main:4.6.2`) comparing the `v4.IndexReport` from stackrox to the produced SBOMs, and the output was as expected / explainable.

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,444 @@
...
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/sbom"
+
+	"github.com/spdx/tools-golang/spdx/common"
+	v2common "github.com/spdx/tools-golang/spdx/v2/common"
+	"github.com/spdx/tools-golang/spdx/v2/v2_3"
+)
+
+type Version string
```
RTann, line 23: won't block the PR, but I think each public field usually has a godoc in this repo

sbom/spdx/encoder.go
RTann: sorry, used to this being used for errors 😄

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,444 @@
...
+	var distIds []int
+	repos := make(map[int]*v2_3.Package)
+	var repoIds []int
+	pkgRels := map[int][]*v2_3.Relationship{}
+	for _, r := range ir.IndexRecords() {
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
+
+		rPkgId, err := strconv.Atoi(r.Package.ID)
```
RTann, line 158: won't block the PR over this, but nit: `pkgID` or `rPkgID` (what does the `r` stand for?)

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,444 @@
...
+		DocumentNamespace: e.DocumentNamespace,
+		CreationInfo: &v2_3.CreationInfo{
+			Creators: creators,
+			Created:  time.Now().Format("2006-01-02T15:04:05Z"),
+		},
+		DocumentComment: e.DocumentComment,
+	}
+
+	pkgs := make(map[int]*v2_3.Package)
+	var pkgIds []int
```
RTann, line 147: won't block the PR over this, but nit: `pkgIDs`

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,444 @@
...
+			pkg = newSpdxPackageFromPackage(r.Package)
+			pkg.PackageFileName = pkgDB
+			pkg.FilesAnalyzed = true
+			pkg.PrimaryPackagePurpose = pkgPurpose
+
+			pkgs[rPkgId] = pkg
+			pkgIds = append(pkgIds, rPkgId)
+
+			if r.Package.Source != nil && r.Package.Source.Name != "" {
+				rSrcPkgId, err := strconv.Atoi(r.Package.Source.ID)
```
RTann, line 188: same nit here, but optional

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,444 @@
...
+						break
+					}
+				}
+				pkg.PackageFileName = pkgDB
+			}
+		}
+
+		// Record Distributions for this package.
+		if r.Distribution != nil {
+			rDistId, err := strconv.Atoi(r.Distribution.ID)
```
RTann, line 227: just realized, do we use the IDs for anything other than converting them into `int`s? If not, then can we just keep them as strings and make the map key string instead of int? Same idea for the slices. Not sure if we have to keep doing this conversion

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,444 @@
...
+	info, infoOK := debug.ReadBuildInfo()
+	if infoOK {
+		for _, m := range info.Deps {
+			if m.Path != "github.com/quay/claircore" {
+				continue
+			}
+			core = m.Version
+			if m.Replace != nil && m.Replace.Version != m.Version {
+				core = m.Replace.Version
+			}
```
RTann, line 435: are we able to break here, or would we have anything to potentially overwrite this with after the first time? Similarly, can we just return here instead? and then just default return "unknown revision" outside the loop

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,444 @@
...
+		}
+	}
+
+	// Now that we have all the data necessary to create the SPDX document,
+	// we need to order it since the IndexRecords aren't in a deterministic order.
+	// This is particular helpful for testing, but it wouldn't be unreasonable
+	// for a user to want to diff different versions of an SPDX of the same IndexReport.
+	slices.Sort(pkgIds)
+	for _, id := range pkgIds {
+		if ctx.Err() != nil {
```
RTann, line 279: I actually have no idea if this is worth doing or not. Fine to keep as-is, but I guess I still don't really know every situation when we should check the context's status.

Google's AI overview says to do it upon each iteration of a loop like this, though take that with a grain of salt... Guess it's not hurting anyone here

sbom/spdx/encoder.go
```diff
@@ -0,0 +1,444 @@
...
+			pkgs[rPkgId] = pkg
+			pkgIds = append(pkgIds, rPkgId)
+
+			if r.Package.Source != nil && r.Package.Source.Name != "" {
+				rSrcPkgId, err := strconv.Atoi(r.Package.Source.ID)
+				if err != nil {
+					return nil, err
+				}
+
+				srcPkg, ok := pkgs[rSrcPkgId]
```
RTann, line 193: won't block the PR over this, but wondering why we add the source here instead of just waiting until we see it? Do we not always see it? In the case we add it here and then see it again later, do we risk potentially duplicating data?

RTann (dismissed):
Approving to ensure this can move along. No true blockers from me, though perhaps the import ordering may be one? I also ask for consideration of the `strconv.Atoi` comments (do we need these calls?) and the comment about `break`ing out of the loop

RTann:
don't forget to sign your commits :) Also looks like there are still a bunch. Seems like you are on top of that, though with the `fixup!` prefixes

sbom/spdx/encoder.go
BradLugo: > what does the r stand for?
"record", as in "the current claircore record we're working with." I think I was creating the spdx package ID and saving that to `pkgId` in some other commit.

sbom/spdx/encoder.go
BradLugo: We convert them to `int`s to sort them. The return from `IndexReport.IndexRecords()` is non-deterministic because it iterates over `IndexReport.Packages`, which is a map. In order to have a deterministic output, both to test against and for users to be able to run a diff between reports, we need to have some sort of ordering structuring. This is what I came up with, but I'm happy to revisit how we do it (perhaps in a follow up PR 🙂)

Some of this context is capture in a comment here: Some context here: https://github.com/quay/claircore/pull/1468/files#diff-9730b2b5baa2c38c549f0b3c1077fddf654896f316b79c2c8f3f77d5fb7ccd1cR273-R276

sbom/spdx/encoder.go
BradLugo: Yeah, I wasn't sure either. I figured if we have the context, we might as well? I'm fine to throw this out, though.

sbom/spdx/encoder.go
BradLugo: @crozzy will likely be able to explain this better, but my understanding is that source packages aren't always recorded as root-level packages. In fact, I _think_ rhcc source packages are the only ones that are also recorded as root-level packages.
For clarity, when I say "root-level packages" here, I'm referring to `IndexReport.Packages` as opposed to `IndexReport.Packages.Source`.

BradLugo:
> don't forget to sign your commits :) Also looks like there are still a bunch. Seems like you are on top of that, though, with the fixup! prefixes

Yup, thank you! It's a new thing I'm trying. The fixup commits have a message body detailing more about the commit to help reviewers. Then, when I get approvals from all participating reviewers, I `rebase --autosquash` at the end to clean up the history and pass CI. Not sure if this is helpful or not. Would love yall's feedback on it!

sbom/spdx/encoder.go
RTann: so if we just cared about sorting for determinism, then I think we can keep it as strings. If we also want to have the IDs in increasing order, then converting to int makes sense. Feel free to change in a followup if you decide to change it

sbom/sbom.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"context"
+	"io"
+
+	"github.com/quay/claircore"
+)
+
+// Encoder is an interface to convert a claircore.IndexReport and writes it to
+// w.
+// that contains a Software Bill of Materials representation.
```
RTann, line 12: nit: I think this is a mix of the old version and the new version of this comment?

BradLugo: 🤦 I didn't mean to add this to that commit. Should be fixed in the latest one + all the other godoc changes.

RTann (approved):

crozzy (approved):
LGTM, nice work

BradLugo:
autosqashing commits; no changes.

PR #1473, opened 2025-01-24
rpm: add function to determine if paths are RPM

crozzy:
Allow language indexers to check if the filepaths they are interested in have been installed via RPM.

The first iteration of this came in the form of https://github.com/quay/claircore/pull/1320 which would save all the RPM files with an rpm.Filescanner and compare them to the language Package's file during the `Resolve` step however it required saving lots of filepath data to the database in order to make the determination. 

This changeset allows dynamic checking of RPM files during language indexing (akin to how we use the [INSTALLER](https://peps.python.org/pep-0376/#installer) file currently) without the extra storage requirements (that could be huge with a large, diverse corpus of images). Also, as the results are cached for the request duration we don't need to trade reduced storage for increased memory/io.


rpm/files.go
```diff
@@ -0,0 +1,102 @@
...
+	"sync"
+
+	"github.com/quay/claircore"
+	"github.com/quay/zlog"
+)
+
+// filesCache is used for concurrent access to the map containing layer.Hash -> map RPM files.
+// The value is a map to allow for quick member checking.
+type filesCache struct {
+	c  map[string]map[string]struct{}
```
crozzy, line 16: It's tempting to use a sync.Map here but I think we want the lock to extend over outer and inner map

crozzy:
supersedes https://github.com/quay/claircore/pull/1320

rpm/files.go
```diff
@@ -0,0 +1,103 @@
...
+func init() {
+	fc = &filesCache{
+		c:  map[string]map[string]struct{}{},
+		mu: &sync.Mutex{},
+	}
+}
+
+// gc deletes the layer's entry from the map if the ctx is done, this ties the lifecycle of
+// the cached information to the request lifecycle to avoid excessive memory consumption.
+func (fc *filesCache) gc(ctx context.Context, key string) {
```
RTann, line 31: Is it assumed each language scanner's context is the same? If not, is it possible we cache a layer, delete it, then have to re-read it because it was just deleted?

rpm/files.go
RTann: I am wondering if we should take some inspiration from [`singleflight`](https://pkg.go.dev/golang.org/x/sync/singleflight) here. Lock per key (layer) and then cleanup once nobody needs it anymore (perhaps similar to the gc we have now. track all the contexts and then cleanup once all known, related contexts are closed).

Not sure if that's work, but if so, I think it'll allow for more concurrency

ruby/packagescanner.go
```diff
@@ -103,6 +104,16 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
 
 	var ret []*claircore.Package
 	for _, g := range gs {
+		isRPM, err := rpm.FileInstalledByRPM(ctx, layer, g)
+		if err != nil {
+			return nil, fmt.Errorf("ruby: unable to check RPM db: %w", err)
```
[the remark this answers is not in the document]
crozzy: I think it's a case of double-checking as sometimes the INSTALLER file is missing (or contains the wrong information)

rpm/files.go
```diff
@@ -0,0 +1,103 @@
...
+	sys, err := layer.FS()
+	if err != nil {
+		return nil, fmt.Errorf("rpm: unable to open layer: %w", err)
+	}
+
+	files := map[string]struct{}{}
+	defer func() {
+		// Defer setting the cache so any early-outs don't have to worry.
+		fc.c[layer.Hash.String()] = files
+	}()
```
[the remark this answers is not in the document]
crozzy: Yes, I think if there was an annoying `fs.FS` where extracting the files failed, repeating the process might exacerbate resource consumption for no gain. If it was reaching out to an external dep I think I'd be more inclined to retry .

rpm/files.go
crozzy: https://github.com/quay/claircore/blob/main/indexer/layerscanner.go#L234 they all start off the same, I guess there is a chance they're changed before the function is called.

ruby/packagescanner.go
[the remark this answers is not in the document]
crozzy: I think given the fact we see in the wild that this file isn't always populated (or populated correctly) we should make extra sure if the cost is minimal (as it should be with this check)

python/packagescanner.go
```diff
@@ -143,14 +154,14 @@ func findDeliciousEgg(ctx context.Context, sys fs.FS) (out []string, err error)
 	// Is this layer an rpm layer?
 	//
 	// If so, files in the disto-managed directory can be skipped.
-	var rpm bool
+	var isRPM bool
```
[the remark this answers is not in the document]
crozzy: I think given the fact we see in the wild that this file isn't always populated (or populated correctly) we should make extra sure if the cost is minimal (as it should be with this check)

ruby/packagescanner.go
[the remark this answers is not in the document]
crozzy: I think keep the errors consistent inside the indexers and then tackle an "across-indexer error consistency" pass in a separate PR, otherwise things get a bit all over the place.

rpm/files.go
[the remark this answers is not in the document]
crozzy: I changed my mind, deferring this means the same call with the same problematic layer can produce two separate results (i.e. error the first time and empty files the next). Just for the sake of someone's sanity during debugging it's worth keeping consistent results for the same inputs.

gobin/gobin_test.go
```diff
@@ -61,6 +61,12 @@ func TestEmptyFile(t *testing.T) {
 	if err := l.Init(ctx, &test.AnyDescription, f); err != nil {
 		t.Error(err)
 	}
+	t.Cleanup(func() {
```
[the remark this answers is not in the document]
crozzy: Yeah.. I can't actually trigger that again, I'm wondering if the GC was being triggered at a different point for some reason that was allowing for the layer not being closed (or GC wasn't run at all before the test finished).

rpm/files.go
crozzy: I have simplified the locking a little (now concurrent RPM DBs reads can happen) and expanded the functionality to track all the calling contexts to inform when deletion happens.

rpm/files.go
```diff
@@ -0,0 +1,118 @@
...
+
+// getFiles looks up RPM files that exist in the RPM database using the
+// filesFromDB function and memorizes the result to avoid repeated work
+// for the same [claircore.Layer].
+func (fc *filesCache) getFiles(ctx context.Context, layer *claircore.Layer) (map[string]struct{}, error) {
+	if fc == nil {
+		panic("programmer error: filesCache nil")
+	}
+	key := layer.Hash.String()
+	if files, ok := fc.get(ctx, key); ok {
```
RTann, line 69: I'm concerned about the scenario where we call `fc.get(ctx, key)` a bunch of times before we ever get to call `fc.set(key, files)`, which would nullify the cache.

I think we can throw lines [73, 106] into a function and use it for a `(*singleflight.Group).DoChan`. Something like:

```
ch := fc.sf.DoChan(key, findFiles(ctx, key, layer))
select {
case <-ctx.Done():
	return nil, ctx.Err()
case res := <-ch:
	return res.Val.(map[string]struct{}), res.Err
}
```

This way, we only ever get the layer's files once until its entry is deleted from the cache

rpm/files.go
```diff
@@ -0,0 +1,118 @@
...
+// get increments the reference counter for the key and looks up the key
+// in the cache. It will subsequently start a goroutine to track when the
+// calling context is done, calling the gc method.
+func (fc *filesCache) get(ctx context.Context, key string) (map[string]struct{}, bool) {
+	fc.mu.Lock()
+	defer fc.mu.Unlock()
+	fc.refs[key]++
+	f, ok := fc.c[key]
+	go func() {
+		<-ctx.Done()
```
RTann, line 48: Can you confirm the contexts will actually close and they aren't just `context.Background()`?

gobin/gobin_test.go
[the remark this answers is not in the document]
RTann: I did once run into something where I forgot to close a layer for a test case and it only popped up when I ran the test as part of a suite of tests instead of individual. I guess in the latter case, the GC just never ran, but it did in the former

rpm/files.go
crozzy: https://github.com/quay/claircore/blob/main/libindex/libindex.go#L187 The context that gets passed to the indexing-related components should be `Done` after the indexing process finishes.

rpm/native_db.go
```diff
@@ -23,6 +28,128 @@ type nativeDB interface {
...
+	}
+	ps, err := fn(ctx, db.String(), nat)
+	if err != nil {
+		return nil, fmt.Errorf("rpm: error reading native db: %w", err)
+	}
+
+	return ps, nil
+}
+
+func filesFromDB(ctx context.Context, _ string, db nativeDB) ([]claircore.File, error) {
```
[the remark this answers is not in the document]
crozzy: I'm not sure I follow, `packagesFromDB` defined after this function

rpm/files.go
crozzy: Thanks for the suggestion, I re-worked this a little, I'm kind of between `Do` and `DoChan` I guess the context cancelling before the result is returned is useful

rpm/native_db.go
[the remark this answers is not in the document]
RTann: you're right I got mixed up

rpm/native_db.go
```diff
@@ -23,6 +28,128 @@ type nativeDB interface {
...
```
RTann, line 128: would we want to add a `defer trace.StartRegion(ctx, "filesFromDB").End()` here? What usually merits a tracing?

rpm/files_test.go
```diff
@@ -0,0 +1,76 @@
...
+	ctx := zlog.Test(context.Background(), t)
+	a := test.NewCachedArena(t)
+	layer := test.LayerRef{
+		Registry: "registry.access.redhat.com",
+		Name:     "ubi9/nodejs-18",
+		Digest:   `sha256:1ae06b64755052cef4c32979aded82a18f664c66fa7b50a6d2924afac2849c6e`,
+	}
+	a.LoadLayerFromRegistry(ctx, t, layer)
+	r := a.Realizer(ctx).(*test.CachedRealizer)
+	t.Cleanup(func() {
```
RTann, line 44: can we add another cleanup task (or something that will definitely run once all the subtests are done) which confirms the files cache is empty and the ref counts are all deleted (to test the gc)?

rpm/files_test.go
```diff
@@ -0,0 +1,76 @@
...
+			URI:       "http://example.com",
+			MediaType: test.MediaType,
+			Headers:   make(map[string][]string),
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	for _, tt := range rpmFilesTestcases {
+		ctx, cancel := context.WithCancel(zlog.Test(context.Background(), t))
```
RTann, line 62: probably doesn't matter much that this is outside the `Run`, right? As in, is this here for a particular reason or just because it doesn't matter?

rpm/files.go
```diff
@@ -0,0 +1,139 @@
...
+		switch {
+		case res.Err == nil:
+			files = res.Val.(map[string]struct{})
+		case errors.Is(res.Err, errNoDBFound):
+		default:
+			return nil, res.Err
+		}
+	}
+
+	fc.set(key, files)
```
RTann, line 126: I think we should call this from within the `DoChan`. Otherwise, we may see a bunch of calls to this for the same layer. We can avoid it by adding the files from within the `DoChan`, so it'll only be called once per layer

rpm/native_db.go
```diff
@@ -23,6 +28,128 @@ type nativeDB interface {
 	Validate(context.Context) error
 }
 
+// ObjectResponse is a generic object that we're expecting to extract from
+// RPM database, currently either a slice of Packages or Files.
+type ObjectResponse interface {
+	[]*claircore.Package | []claircore.File
+}
+
+// getDBObjects does all the dirty work of extracting generic claircore objects
```
RTann, line 37: nit: `GetDBObjects`

rpm/files_test.go
crozzy: It makes it a little more interesting that the context is cancelled in a different goroutine but I don't think it really adds to the validity of the test, I can move it inside for readability

rpm/files.go
```diff
@@ -0,0 +1,139 @@
...
+// GC decrements the reference counter and deletes the [claircore.Layer]'s
+// entry from the cache map and the reference counter once the references
+// are 0 (or less).
+func (fc *filesCache) gc(key string) {
+	fc.mu.Lock()
+	defer fc.mu.Unlock()
+	fc.refs[key]--
+	if fc.refs[key] <= 0 {
+		delete(fc.c, key)
+		delete(fc.refs, key)
```
BradLugo, line 38: Should we also garbage collect the `singleflight.Group` memoization? Maybe something like:
```suggestion
		delete(fc.c, key)
		delete(fc.refs, key)
		s.sf.Forget(key)
```

crozzy: This would be an odd situation where the DB extraction was still in progress but the refs were 0 but I guess if there is some unexpected context cancellation then there's a chance that key still exists.

rpm/files.go
```diff
@@ -0,0 +1,139 @@
...
+
+var errNoDBFound = errors.New("no RPM DB found")
+
+// GetFiles looks up RPM files that exist in the RPM database using the
+// filesFromDB function and memorizes the result to avoid repeated work
+// for the same [claircore.Layer].
+func (fc *filesCache) getFiles(ctx context.Context, layer *claircore.Layer) (map[string]struct{}, error) {
+	if fc == nil {
+		panic("programmer error: filesCache nil")
+	}
```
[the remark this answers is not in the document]
crozzy: I was trying to guard against a developer error in the future ensuring it fails fast, we don't usually check it but we are also not usually doing package scoped objects. Maybe we could remove, the results would be somewhat similar with a different panic message.

rpm/files.go
BradLugo: Do we ever `Forget()` (or otherwise clear) the results cached in `fc.sf`? My current understanding is that `fc.sf` and `fc.c` contain essentially the same data.

crozzy: `fc.sf` isn't caching results, it's suppressing concurrent executions of the same function by blocking the "non-first" calls until the first call finishes and then sharing the results. If another call is made after the initial calls have finished the function will be run again. i.e Singleflight is implicitly "forgetting".

RTann: @BradLugo the internal `doCall` has a line `delete(g.m, key)` which is essentially the same as `(*singleflight.Group).Forget(key)`, so it's not needed - https://cs.opensource.google/go/x/sync/+/refs/tags/v0.11.0:singleflight/singleflight.go;l=160

Then again, I guess putting it here won't hurt anyone...? I'm ok either way I think

crozzy: Yeah, I'll stick it in for old time's sake

RTann: Then again, I think it is possible a single instance of Clair/Claircore may analyze the same layer at the same time, so calling this `Forget` may somehow cause us to redo some of this work. Fun thinking about concurrency 😂

rpm/native_db.go
```diff
@@ -23,6 +28,128 @@ type nativeDB interface {
...
+		if err != nil {
+			return nil, fmt.Errorf("rpm: error reading sqlite db: %w", err)
+		}
+		defer func() {
+			if err := os.Remove(f.Name()); err != nil {
+				zlog.Error(ctx).Err(err).Msg("unable to unlink sqlite db")
+			}
+			if err := f.Close(); err != nil {
+				zlog.Warn(ctx).Err(err).Msg("unable to close sqlite db")
+			}
```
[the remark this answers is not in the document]
crozzy: I believe this a pattern when dealing with temporary files where you first unlink/remove them (so they aren't visible in the fs but once the file handles are closed the contents are deleted).

In this case where they are called in succession it doesn't make a difference and as this code is just moved from a different file I'm inclined to keep it as is unless there is a correctness reason to change it.

rpm/files_test.go
crozzy: So we can, but we can't guarantee that the `gc` go routine will complete before the test so we have two bad options:
* Add a waitgroup to the filesCache that only the test will call.
* [Sleep](https://cs.opensource.google/go/x/sync/+/refs/tags/v0.11.0:singleflight/singleflight_test.go;l=125)

rpm/files.go
BradLugo: dang, this `singleflight` package is cool. Thanks for the explanation yall! 😁

BradLugo (approved):
LGTM after @RTann's comments are resolved

rpm/files_test.go
crozzy: I went down the first route

rpm/native_db.go
crozzy: I think this is valid to bring it in line with `packagesFromDB`

rpm/files_test.go
```diff
@@ -0,0 +1,106 @@
...
+			isRPM, err := FileInstalledByRPM(ctx, &realizedLayers[0], tt.filePath)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if tt.isRPM != isRPM {
+				t.Errorf("expected isRPM: %t, got isRPM: %t", tt.isRPM, isRPM)
+			}
+			cancel()
+		})
+	}
```
RTann, line 93: instead of adding a `sync.WaitGroup` to `filesCache` can we just do this instead?

```suggestion
	var wg sync.WaitGroup
	for _, tt := range rpmFilesTestcases {
		wg.Add(1)
		t.Run(tt.name, func(t *testing.T) {
			defer wg.Done()
			ctx, cancel := context.WithCancel(zlog.Test(context.Background(), t))
			t.Parallel()
			isRPM, err := FileInstalledByRPM(ctx, &realizedLayers[0], tt.filePath)
			if err != nil {
				t.Fatal(err)
			}
			if tt.isRPM != isRPM {
				t.Errorf("expected isRPM: %t, got isRPM: %t", tt.isRPM, isRPM)
			}
			cancel()
		})
	}
```

If we do end up requiring it for testing purposes, can you add a comment in the struct to make it clear it's only there for testing purposes?

RTann: ah ok just saw your other comment. We need to know when `gc` is called. Makes sense. Can you add a comment to the filesCache struct to make it clear it's just there for testing purposes? I was confused what it was doing there until I saw this. Thanks

rpm/files.go
```diff
@@ -0,0 +1,140 @@
...
+			fs, err := getDBObjects(ctx, sys, db, filesFromDB)
+			if err != nil {
+				return nil, fmt.Errorf("rpm: error getting native DBs: %w", err)
+			}
+			for _, f := range fs {
+				files[f.Path] = struct{}{}
+			}
+			fc.set(key, files)
+		}
+		return files, nil
```
RTann, line 113:
```suggestion
		}
		fc.set(key, files)
		return files, nil
```

crozzy: Hmm, good catch, I guess this would override files if there were more than one rpm DB in a layer (cornercase but valid)

rpm/files.go
```diff
@@ -0,0 +1,142 @@
...
+
+// FilesCache is used for concurrent access to the map containing
+// [claircore.Layer] -> map[string]struct{}.
+type filesCache struct {
+	c    map[string]map[string]struct{}
+	refs map[string]int
+	mu   sync.Mutex
+	sf   singleflight.Group
+	// This wg is added so tests can wait for goroutines to finish to
+	// be able to confirm functionality.
```
BradLugo, line 23: It took me a bit to figure out what's going on here, and it was only after I came back to Ross's review comments that I realized the intention. Could we make it clearer what exactly this WaitGroup is used for? Here's kind of what I'm thinking:
```suggestion
	// wg tracks the gc calls made during get. This is particularly useful
	// during testing, so we can wait on all the gc calls before asserting
	// the cache is indeed empty and all the references were accounted for.
```

BradLugo (approved):
I have just one hang-up, but it's non-blocking. I think we should get more reviewers' approval before merging.

BradLugo (approved):
LGTM! I still think we should get more reviewers' approval before merging 🙂

RTann (approved):

hdonnay (approved):
SGTM

PR #1478, opened 2025-02-12
gobin: always report the main version in Package.Version

crozzy:
Go 1.24 changes the way in which the main module's version is reported when the module is build via `go build`. What used to be just reported as `(devel)` will now be reported as the version + revision info (+dirty if status is uncommitted) unless the -buildvcs=false is set. All this means that the unit tests failed as the expected `(devel)` version wasn't found and instead "" was reported because the gobin detector doesn't set any Version if the main module's version was parseable (this isn't a big deal because the matching always works of NormalizedVersion which is always set). This change always adds the Version and the test checks the Version with the same regex `ParseVersion` uses.

### DIFF
--- a/gobin/exe.go
+++ b/gobin/exe.go
@@ -84,15 +84,22 @@ func toPackages(ctx context.Context, out *[]*claircore.Package, p string, r io.R
 	vs := map[string]string{
 		"stdlib": bi.GoVersion,
 	}
-	var mmv string
-	mainVer, err := ParseVersion(bi.Main.Version)
+
+	// The go main module version is reported differently depending on the go
+	// toolchain, pre-go1.24 module versions built with `go build` will report
+	// `(devel)`, go1.24 and above will report this version like this:
+	// v1.5.36-0.20250212170732-e3af313feaab+dirty i.e. the version stamped in
+	// the compiled binary is based on the version control system tag and/or commit.
+	// A +dirty suffix will be appended if there are uncommitted changes. Previous
+	// behavior can be obtained by using the flag `-buildvcs=false`.
+	mmv := bi.Main.Version
+	mainVer, err := ParseVersion(mmv)
 	switch {
 	case errors.Is(err, nil):
 	case bi.Main.Version == `(devel)`, bi.Main.Version == ``:
-		// This is currently the state of any main module built from source; see
-		// the package documentation. Don't record it as a "bad" version and
-		// pull out any vcs metadata that's been stamped in.
-		mmv = bi.Main.Version
+		// This is currently the state of any main module built from source
+		// before go1.24; see the package documentation. Don't record it as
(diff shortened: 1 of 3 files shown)


crozzy:
Links:
* [go change](https://cs.opensource.google/go/go/+/8aa2eed8fb90303c0876e51e097105eca7299734:src/cmd/go/internal/load/pkg.go;dlc=b67443459a5bbe0e125e34538b4ce3c895de9efd)
* [test failure](https://github.com/quay/claircore/actions/runs/13290889064/job/37111163101)

crozzy:
To be explicit, the issue is not something that will affect clair once it starts to use go1.24, it's that behaviour will change when we're detecting things built with go1.24, which is imminently. The behaviour change isn't necessarily "breaking", it just means we'll potentially match more things.

gobin/exe.go
```diff
@@ -84,15 +84,14 @@ func toPackages(ctx context.Context, out *[]*claircore.Package, p string, r io.R
 	vs := map[string]string{
 		"stdlib": bi.GoVersion,
 	}
-	var mmv string
+	mmv := bi.Main.Version
 	mainVer, err := ParseVersion(bi.Main.Version)
 	switch {
 	case errors.Is(err, nil):
 	case bi.Main.Version == `(devel)`, bi.Main.Version == ``:
 		// This is currently the state of any main module built from source; see
```
RTann, line 92: Can you update this to say this is for anything built for go1.23 and below?

gobin/exe.go
```diff
@@ -84,15 +84,14 @@ func toPackages(ctx context.Context, out *[]*claircore.Package, p string, r io.R
 	vs := map[string]string{
 		"stdlib": bi.GoVersion,
 	}
-	var mmv string
+	mmv := bi.Main.Version
 	mainVer, err := ParseVersion(bi.Main.Version)
```
RTann, line 88: nit: might as well use `mmv` in here?

RTann (commented):
Ah that's what this is referring to:

> The go build command now sets the [main module’s version](https://tip.golang.org/pkg/runtime/debug#BuildInfo.Main) in the compiled binary based on the version control system tag and/or commit. A +dirty suffix will be appended if there are uncommitted changes. Use the -buildvcs=false flag to omit version control information from the binary.

https://tip.golang.org/doc/go1.24#go-command

I also see https://github.com/golang/go/issues/50603 referenced in gobin/gobin.go at the very top. Can we adjust the comment based on the new changed?

RTann:
Can you also give some examples of how this changes things? What did we see in the past and what will we now see with things built with go1.24?

crozzy:
> Can you also give some examples of how this changes things? What did we see in the past and what will we now see with things built with go1.24?

I added a new comment with an example, do you mean add a comment to the PR?

RTann (approved):

PR #1488, opened 2025-02-24
replace io.ReadCloser with io.Reader in Parse methods

RTann:
There is no reason that I can see why these methods must take in `io.ReadCloser`. The update manager already closes them, so might as well just pass in `io.Reader` instead.

### DIFF
--- a/alpine/parser.go
+++ b/alpine/parser.go
@@ -17,10 +17,9 @@ const (
 
 var _ driver.Parser = (*updater)(nil)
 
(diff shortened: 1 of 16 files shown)


crozzy:
This is a change to the public interface meaning there is potential to cause pain to anyone who has implemented it. In practice, I think the pain is likely to be minimal considering `io.Reader` will always satisfy `io.ReadCloser` so it would only affect people who were `Close()`ing the reader (which I don't really know why someone would.

PR #1489, opened 2025-02-24
chore: update to go1.23

crozzy:
According to the [Go Release policy](https://go.dev/doc/devel/release#policy) Go 1.22 is no longer supported, and things coming down the pipe will rely on 1.23's iterator support.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/quay/claircore
 
-go 1.22.0
+go 1.23.0
 
 require (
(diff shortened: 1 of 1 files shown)


RTann (approved):
I imagine we are ok with using an older version for the `toolkit`, right?

crozzy:
> I imagine we are ok with using an older version for the `toolkit`, right?

I don't think there is anything that requires go1.23 in `toolkit` and `updater/driver`

PR #1502, opened 2025-03-17
chore: remove toolchain from go.mod

RTann:
I don't have go1.24.1 on my machine, and I don't think we need to force people to use it (at this time)


crozzy (approved):
LGTM

PR #1504, opened 2025-03-26
test: layer caching race

hdonnay:
The CI for quay/claircore#869 seems to get caught on this race, which it introduces by trying to cache the same layer in multiple processes.


hdonnay:
Please merge if this doesn't need a rebase

crozzy (approved):
Just confirming the cache package and the use of the use of `cleancache` cmd are gated behind use of go1.24?

hdonnay:
Yeah, the command and the relevant 1.24-new bits are behind build tags.

Issue #1419, opened 2024-09-25
Claircore does not fetch Oracle ELSAs from the current year


RTann:
As of at least v1.5.31, Claircore does not fetch Oracle Linux vulnerabilities from the current year. See https://github.com/quay/claircore/blob/v1.5.31/oracle/updaterset.go#L13

I fixing this brings an opportunity to improve the Oracle Linux support:

* https://linux.oracle.com/security/oval/ provides version-specific OVAL files. Perhaps Claircore can switch to those
* If switching to those, there will no longer be a need for hardcoding releases in https://github.com/quay/claircore/blob/v1.5.31/oracle/releases.go. Instead, Oracle Linux vuln updater can essentially mimic something like Alpine: https://github.com/quay/claircore/blob/v1.5.31/alpine/updater.go (have an UpdaterSet Factory and search for versions from 6 - whatever)

crozzy:
Yeah that's weird, nothing in the commit adding the `UpdaterSet` alludes to purposefully missing the current year. I think the approach seems reasonable (we'd need to workout deleting old vulns as I think the updater name would be different).

BradLugo:
https://github.com/quay/claircore/pull/1509 provides a rudimentary, short-term fix. We should still revisit @RTann's ideas for further improvements.

PR #1509, opened 2025-04-02
oracle: inclusive year limit in updater

BradLugo:
Fixes the Oracle updater to include the current year.

Proof:
<details>
<summary>
main.go
</summary>

```go
package main

import (
        "fmt"
        "time"
)

func main() {
        fmt.Println("Current loop:")
        for year, lim := 2007, time.Now().Year(); year != lim; year++ {
                fmt.Println(year)
        }

        fmt.Println()

        fmt.Println("Fixed loop:")
        for year, lim := 2007, time.Now().Year(); year <= lim; year++ {
                fmt.Println(year)
        }
}
```
</details>

<details>
<summary>
output
</summary>

```
Current loop:
2007
2008
2009
2010
2011
2012
2013
2014
2015
2016
2017
2018
2019
2020
2021
2022
2023
2024

Fixed loop:
2007
2008
2009
2010
2011
2012
2013
2014
2015
2016
2017
2018
2019
2020
2021
2022
2023
2024
2025
```
</details>

Related to https://github.com/quay/claircore/issues/1419.

### DIFF
--- a/oracle/updaterset.go
+++ b/oracle/updaterset.go
@@ -10,7 +10,7 @@ import (
 
 func UpdaterSet(_ context.Context) (driver.UpdaterSet, error) {
 	us := driver.NewUpdaterSet()
-	for year, lim := 2007, time.Now().Year(); year != lim; year++ {
+	for year, lim := 2007, time.Now().Year(); year <= lim; year++ {
 		u, err := NewUpdater(year)
 		if err != nil {
 			return us, fmt.Errorf("unable to create oracle updater: %v", err)


RTann (approved):
I'm ok with this for now, though I'd still love to see a change like what was proposed in https://github.com/quay/claircore/issues/1419 so we would not have to manually track Oracle Linux versions anymore.

Definitely wait for one more approval to make sure everyone it ok with this.

Thanks for doing this!

RTann:
I guess this does technically close https://github.com/quay/claircore/issues/1419, but I don't really want to lose track of the desire to stop hardcoding Oracle Linux versions. Would you mind making a followup issue to track that? Or perhaps we keep https://github.com/quay/claircore/issues/1419 around to track that

oracle/updaterset.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func UpdaterSet(_ context.Context) (driver.UpdaterSet, error) {
 	us := driver.NewUpdaterSet()
-	for year, lim := 2007, time.Now().Year(); year != lim; year++ {
+	for year, lim := 2007, time.Now().Year(); year <= lim; year++ {
```
RTann, line 13: From what I can tell, let's say it's January 1, 2026 and the 2026 OVAL file does not exist yet, we will still create the UpderSet, no problem, but we may see error messages for who knows how long. Is that ok?

BradLugo: I don't know enough about UpdaterSets to answer that. If we've solved this problem elsewhere, I'd be in favor of that solution for consistency. If we haven't, I need to better understand the intention of the failure mode of an UpdaterSet. My gut reaction is to log it as a warning. Or maybe even introduce a new error that indicates a 404, warn on that, and error in all other cases.

daynewlee: The change looks good to me. Typically, if we encounter some error messages but the overall Scanner V4 update workflow still succeeds, we rarely investigate individual errors or warnings unless the entire workflow fails.
It appears that in the Oracle update process, the libvuln update manager runs each updater independently. Even if one or more Oracle updaters fail, the rest continue executing.

daynewlee (approved):

oracle/updaterset.go
crozzy: Yep, confirming what @daynewlee said, the update process will carry on and errors will be aggregated and logged at the end.

crozzy (approved):

BradLugo:
rebasing

PR #1510, opened 2025-04-02
rpm: versioning fixes

hdonnay:
While working on #869, I got to a point where I wasn't sure if I could trust the test suite, which is an untenable state of affairs.

This is splitting out that work so it can be applied first.


internal/rpmver/rpmver_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+func TestRpmvercmp(t *testing.T) {
+	f, err := os.Open("testdata/rpmvercmp")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+	s := bufio.NewScanner(f)
+
+	exe, _ := exec.LookPath("rpm")
```
RTann, line 87: should we also ensure we are using a version which would even exist in RHEL? For example, I believe RHEL 9 is using 4.16. Not sure what 10 will use, and not sure what version we are on now, but probably something > 4.16

internal/rpmver/rpmver_test.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+
+			if crosscheck {
+				out, err := exec.Command(exe, `--eval`, fmt.Sprintf(`%%{lua: print(rpm.vercmp("%s", "%s"))}`, a, b)).Output()
+				if err != nil {
+					t.Fatalf("error exec'ing %q: %v\n%s", exe, err, string(out))
+				}
+
+				var n int
+				if _, err := fmt.Sscan(string(out), &n); err != nil {
```
RTann, line 135: TIL about `fmt.Sscan`

internal/rpmver/testdata/rpmvercmp
RTann, line 1: these tests only cover the version and release parts of the version, but they don't cover the name, epoch, nor arch. Can we add something for these?

internal/wart/pointer_slice.go
```diff
@@ -0,0 +1,34 @@
...
+
+// CollectPointer is the equivalent of "[slices.Collect]([AsPointer](seq))".
+func CollectPointer[T any](seq iter.Seq[T]) []*T {
+	backing := slices.Collect(seq)
+	ret := make([]*T, len(backing))
+	for i := range backing {
+		ret[i] = &backing[i]
+	}
+	return ret
+}
```
RTann, line 34: I think I asked this in the original PR, but not sure if it had a response: why not just do the following since it's equivalent?

```suggestion
// CollectPointer is the equivalent of "[slices.Collect]([AsPointer](seq))".
func CollectPointer[T any](seq iter.Seq[T]) []*T {
	return slices.Collect(AsPointer(seq))
}
```

This should use one fewer slice

rpm/native_db.go
```diff
@@ -393,14 +394,12 @@ func constructEVR(b *strings.Builder, info *Info) string {
 	return b.String()
 }
 
-func constructHint(b *strings.Builder, info *Info) string {
-	b.Reset()
+func constructHint(_ *strings.Builder, info *Info) string {
```
RTann, line 397: this may be mentioned elsewhere, but if we are removing the need of the builder, then why keep the parameter around?

rpm/native_db.go
```diff
@@ -412,22 +411,16 @@ func constructHint(b *strings.Builder, info *Info) string {
 				if p.SigType != 0 {
 					continue
 				}
-				if b.Len() != 0 {
-					b.WriteByte('|')
```
RTann, line 416: instead of appending, we will now overwrite the `key` if we get both `*packet.SignatureV3` and `*packet.Signature`. Is this possible, and if so, is overwriting expected?

RTann (commented):
I skipped everything under `test/` as well as `rpm/packagescanner_test.go`. I can go over `rpm/packagescanner_test.go` another time, if desired, but anything under `test/` is very unknown to me. I can look into it, if needed, otherwise I'll leave it for someone who knows more about that than I do

internal/rpmver/cmp_string_test.go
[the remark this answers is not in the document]
hdonnay: They're only ever printed in the tests.

internal/rpmver/rpmver.go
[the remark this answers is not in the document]
hdonnay: Yes, but I didn't want to roll that in here. I'll split off an issue.

internal/rpmver/testdata/rpmvercmp
hdonnay: I can, but they just use "normal" string comparison.

internal/wart/pointer_slice.go
hdonnay: My thinking was that at the cost of paying for one extra allocation, the GC only has two blocks to worry about: `[]*T` with pointers into `[]T`, rather than N+1 blocks: `[]*T` with pointers to N `T`.

It might be superstition, as I don't know how to check it other than "this seems to reduce time in GC with large slices-of-pointers".

internal/rpmver/rpmver_test.go
hdonnay: It shouldn't matter, as it's just using the built-in `lua` eval to call the comparison function. [The `rpm` test suite does this](https://github.com/rpm-software-management/rpm/blob/a333eaa3f0fbce0c77cea23015144e13d9c039e2/tests/rpmvercmp.at#L3-L7) (CW: m4), so it should be pretty stable.

internal/rpmver/rpmver.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// MarshalText implements [encoding.TextMarshaler].
+func (v *Version) MarshalText() ([]byte, error) {
+	return []byte(v.String()), nil
+}
+
+// IsZero reports true if the receiver is a zero-valued [Version].
+func (v *Version) IsZero() bool {
+	return v.Name == nil && v.Architecture == nil && v.Epoch == "" && v.Version == "" && v.Release == ""
```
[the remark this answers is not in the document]
hdonnay: I'd rather this panic if called on a `nil` pointer.

rpm/native_db.go
hdonnay: To minimize the code churn in this PR. #869 cleans this up when reworking the `rpm` package, so I didn't bother pulling that change "forward."

rpm/native_db.go
hdonnay: That should never happen because a given `rpm` version should only be producing one version of signature. If they ever do get mixed, I'm fine with last-one-wins.

internal/rpmver/rpmver.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+// Rpmvercmp compares RPM version strings.
+//
+// This is a port of the C version at https://github.com/rpm-software-management/rpm/blob/572844039a04846fe9e030cbacb6336e2240bd6f/rpmio/rpmvercmp.cc
+//
+//	 1: a is newer than b
+//	 0: a and b are the same version
+//	-1: b is newer than a
+func rpmvercmp(a, b string) int {
```
RTann, line 200: It's hard to tell, but I don't think this has changed since I last reviewed. When I last reviewed, this all looked good to me after I did a comparison between the C version and this version

RTann (approved):
From what I can tell, everything outside `test/` looks good. I haven't checked there, but I can if you need. Curious to see how this'll be used for the `dnf` PR

Issue #547, opened 2021-12-23
Clair incorrectly flags Oracle Linux ksplice userspace errata


jetsaredim:
Clair appears to be flagging Oracle Linux-based container images for errata that are only applicable to ksplice userspace packages.  These errata should never be applied to a container since ksplice userspace packages are not supported to be run within a container.  As such, these errata should never be flagged when scanning an Oracle Linux-based container.

floyd673:
hello you welcome how may we help you out

jetsaredim:
@floyd673 I used to be the maintainer of the code that generated the Oracle Linux OVAL files but I haven't been in that role for a couple years.  That said, from what I remember there needs to be code added to clair to handle OVAL errata definitions that are unique to Oracle Linux - specifically for things like UEK or ksplice that are slightly different from other distros - including RHEL.

hdonnay:
> hello you welcome how may we help you out

Changing the Oracle updater to omit the offending packages would be the way forward.

BradLugo:
> Changing the Oracle updater to omit the offending packages would be the way forward.

The change submitted in https://github.com/quay/claircore/pull/1511 omits the offending packages; however, I'm a bit concerned this might cause false negatives for bootc images (if one were to build such an image for Oracle Linux). I believe this is a moot point right now since we don't fully support bootc images today, but I felt it was worth noting.

hdonnay:
This is done, I think.

PR #1511, opened 2025-04-03
oracle: omit ksplice-related vulnerabilities

BradLugo:
Oracle's OVAL data includes ksplice userspace packages that aren't supported to run in containers.

This PR changes the parsing of this data to omit ksplice-related vulnerabilities by checking the affected CPEs list. If all CPEs are all related to ksplice, we'll omit it. If there is at least one CPE that is **not** related to ksplice, we'll add all the vulnerabilities anyway, which _may_ cause us to record false positives for that vulnerability, but that case seems very unlikely.

Related to https://github.com/quay/claircore/issues/547.

### DIFF
--- a/oracle/parser.go
+++ b/oracle/parser.go
@@ -12,6 +12,7 @@ import (
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/internal/xmlutil"
 	"github.com/quay/claircore/libvuln/driver"
+	"github.com/quay/claircore/pkg/cpe"
 	"github.com/quay/claircore/pkg/ovalutil"
 )
 
@@ -71,6 +72,50 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
 		if len(vs) == 0 {
 			return nil, fmt.Errorf("could not determine dist")
 		}
+
+		// Check if the vulnerability only affects a userspace_ksplice package.
+		//  These errata should never be applied to a container since ksplice
+		//  userspace packages are not supported to be run within a container.
+		// If there's at least one ksplice CPE and not all the affected CPEs
+		//  are ksplice related, this will cause false positives we can catch.
+		//  This should rarely happen; the most common case for this is if one
+		//  of the CPEs wasn't parseable.
+		kspliceCPEs := 0
+		cpes := len(def.Advisory.AffectedCPEList)
+		for _, affected := range def.Advisory.AffectedCPEList {
+			wfn, err := cpe.Unbind(affected)
+			if err != nil {
+				// Found a CPE but could not parse it. Log a warning and return
+				//  successfully.
+				zlog.Warn(ctx).
+					Str("def_title", def.Title).
+					Str("cpe", affected).
+					Msg("could not parse CPE: there may be a false positive match with a userspace_ksplice package")
+				return vs, nil
+			}
+			if wfn.Attr[cpe.Edition].V == "userspace_ksplice" {
+				kspliceCPEs++
+			}
+		}
+
+		switch diff := cpes - kspliceCPEs; {
+		case kspliceCPEs == 0:
+			// Continue if there are no ksplice CPEs.
+		case cpes == 0:
+			zlog.Warn(ctx).
+				Str("def_title", def.Title).
+				Msg("potential false positives: couldn't find CPEs to check for ksplice packages")
+		case diff == 0:
+			zlog.Debug(ctx).Msg("skipping userspace_ksplice vulnerabilities")
+			return nil, nil
+		case diff > 0:
+			zlog.Warn(ctx).
+				Str("def_title", def.Title).
+				Msg("potential false positives: OVAL may have a userspace_ksplice CPE which could not be skipped")
+		default:
+			panic("programmer error")
+		}
+
 		return vs, nil
 	}
 	vulns, err := ovalutil.RPMDefsToVulns(ctx, &root, protoVulns)
--- a/oracle/parser_test.go
+++ b/oracle/parser_test.go
@@ -2,12 +2,9 @@ package oracle
 
 import (
 	"context"
-	"encoding/xml"
 	"os"
 	"testing"
-	"time"
 
-	"github.com/quay/goval-parser/oval"
 	"github.com/quay/zlog"
 )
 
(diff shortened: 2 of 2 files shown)


hdonnay:
This looks right, although it should probably just omit advisories that are for ksplice instead of erroring.

BradLugo:
It appears one way to omit it is by erroring in the `protoVulns` callback:
https://github.com/quay/claircore/blob/178b6949de4c00b3688841ede73a29c2eb0145fc/pkg/ovalutil/rpm.go#L49-L56
I don't know which is more idiomatic to claircore. The other way is to log and return an empty list + nil error, I think.

oracle/parser.go
```diff
@@ -71,6 +72,44 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
...
+			}
+			if wfn.Attr[cpe.Edition].V == "userspace_ksplice" {
+				atLeastOneKsplice = true
+			} else {
+				isOnlyKsplice = false
+			}
+		}
+
+		if !isOnlyKsplice {
+			//  This should (almost) never happen.
```
crozzy, line 105: Nit: Which condition should (almost) never happen?

oracle/parser.go
```diff
@@ -71,6 +72,44 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
...
+		//  This should (almost) never happen.
+		atLeastOneKsplice := false
+		for _, affected := range def.Advisory.AffectedCPEList {
+			wfn, err := cpe.Unbind(affected)
+			if err != nil {
+				// Found a CPE but could not parse it. Assume it's not a
+				//  userspace_ksplice package.
+				zlog.Warn(ctx).Str("cpe", affected).Msg("could not parse CPE")
+				isOnlyKsplice = false
+				atLeastOneKsplice = true
```
crozzy, line 95: Do we want to break here? Doesn't look like anything will change in subsequent calls after these are set

oracle/parser_test.go
```diff
@@ -28,108 +25,7 @@ func TestParse(t *testing.T) {
 		t.Fatal(err)
 	}
 	t.Logf("found %d vulnerabilities", len(vs))
-	if got, want := len(vs), 3854; got != want {
+	if got, want := len(vs), 6021; got != want {
 		t.Fatalf("got: %d vulnerabilities, want: %d vulnerabilities", got, want)
 	}
 }
-
-var ovalDef = oval.Definition{XMLName: xml.Name{Space: "http://oval.mitre.org/XMLSchema/oval-definitions-5", Local: "definition"},
```
crozzy, line 36: Weird, this is just never used?

crozzy:
LGTM, just a few nits

oracle/parser.go
BradLugo: This is referring to the comment above: https://github.com/quay/claircore/pull/1511/files#diff-306dfd83be9b72dac194cdc4cc9022d33bc8f67b73567be2a23fa53df4b4412fR84-R86
Do you have any suggestions on how to make this clearer? I could copy+paste the whole comment here.

oracle/parser_test.go
BradLugo: My editor couldn't find any usages 🤷

oracle/parser.go
```diff
@@ -71,6 +72,45 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
...
+		//  userspace packages are not supported to be run within a container.
+		// If we couldn't find a CPE list, make sure to include the
+		//  vulnerability. We'd rather have false positives for
+		//  userspace_ksplice packages than have false negatives for
+		//  *everything*.
+		isOnlyKsplice := len(def.Advisory.AffectedCPEList) > 0
+		// If there's at least one ksplice CPE and not all the affected CPEs
+		//  are ksplice related, this will cause false positives we can catch.
+		//  This should (almost) never happen.
+		atLeastOneKsplice := false
```
RTann, line 87: I'm wondering if it may be cleaner to just count the number of `userspace_ksplice` CPEs and then compare this with the total number of CPEs after the loop

oracle/parser.go
```diff
@@ -71,6 +72,45 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
...
+		for _, affected := range def.Advisory.AffectedCPEList {
+			wfn, err := cpe.Unbind(affected)
+			if err != nil {
+				// Found a CPE but could not parse it. Assume it's not a
+				//  userspace_ksplice package.
+				zlog.Warn(ctx).Str("cpe", affected).Msg("could not parse CPE")
+				isOnlyKsplice = false
```
RTann, line 94: how do we know for a fact this is the case?

oracle/parser.go
```diff
@@ -71,6 +72,45 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
...
+				atLeastOneKsplice = true
```
RTann, line 95: The comment above says we are assuming this is not a ksplice package, but this is set to `true`. Is that intentional?

oracle/parser.go
```diff
@@ -71,6 +72,45 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
...
+			} else {
+				isOnlyKsplice = false
+			}
+		}
+
+		if !isOnlyKsplice {
+			//  This should (almost) never happen.
+			if atLeastOneKsplice {
+				zlog.Warn(ctx).Str("def.Title", def.Title).Msg("potential false positives: vulnerability has at least one unskippable ksplice match")
```
RTann, line 108: I think the logs typically do underscore instead of dot

```suggestion
				zlog.Warn(ctx).Str("def_title", def.Title).Msg("potential false positives: vulnerability has at least one unskippable ksplice match")
```

oracle/parser.go
```diff
@@ -71,6 +72,45 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
...
+			}
+		} else {
+			return nil, fmt.Errorf("skippable ksplice vulnerabilities")
```
RTann, line 111:
```suggestion
			return nil, errors.New("skippable ksplice vulnerabilities")
```

actually, how come we want to error here? Would it not be valid to `return nil, nil`, or do we prefer the signal an error message?

BradLugo: I thought it might be dangerous to return a nil pointer to the caller if there's no error, but I just tested that with a list of pointers in Go Playground, and it doesn't seem to be a problem. I'll add a debug log instead.

oracle/parser.go
BradLugo: I'm open to suggestions.

BradLugo:
@RTann I updated the comments in the changes. Let me know if they answer some of your review questions.

oracle/parser.go
```diff
@@ -71,6 +72,49 @@ func (u *Updater) Parse(ctx context.Context, r io.ReadCloser) ([]*claircore.Vuln
...
+			return nil, nil
+		}
+
+		if atLeastOneKsplice {
+			zlog.Warn(ctx).
+				Str("def_title", def.Title).
+				Msg("potential false positives: vulnerability has at least one unskippable ksplice match")
+		}
+
 		return vs, nil
```
RTann, line 118: how does something like this look? I was getting confused about some of the `isOnlyKsplice` and `atLeastOneKsplice` uses

Also, I realized we never really accounted for CPEs prior to this, but we do for RHEL. Should we be making an entry per CPE for this, too?

```suggestion
		kspliceCPEs := 0
		cpes := len(def.Advisory.AffectedCPEList)
		for _, affected := range def.Advisory.AffectedCPEList {
			wfn, err := cpe.Unbind(affected)
			if err != nil {
				// Found a CPE but could not parse it. Just log a warning and return successfully.
				zlog.Warn(ctx).
					Str("def_title", def.Title).
					Str("cpe", affected).
					Msg("could not parse CPE: there may be a false positive match with a userspace_ksplice package")
				return vs, nil
			}
			if wfn.Attr[cpe.Edition].V == "userspace_ksplice" {
				kspliceCPEs++
			}
		}

		switch diff := len(def.Advisory.AffectedCPEList) - kspliceCPEs; {
		case cpes == 0:
			// Just continue if there are no CPEs.
		case diff == 0:
			zlog.Debug(ctx).Msg("skipping userspace_ksplice vulnerabilities")
			return nil, nil
		case diff > 0:
			zlog.Warn(ctx).
				Str("def_title", def.Title).
				Msg("potential false positives: OVAL may have a userspace_ksplice CPE which could not be skipped")
		default:
			panic("programmer error")
		}

		return vs, nil
```

BradLugo: > Also, I realized we never really accounted for CPEs prior to this, but we do for RHEL. Should we be making an entry per CPE for this, too?

I'm not sure. There's probably an improvement we can make to make an entry per CPE, but I _think_ the rest of the OVAL data (criteria + rpm object lookups) exposes all the entries we need. I might be missing the point of your question, though (I'm still getting comfortable with CPEs and such).

RTann (approved):

BradLugo:
@crozzy @hdonnay, I think this is ready for final reviews.

BradLugo:
rebasing

PR #1514, opened 2025-04-09
indexer: align affectedManifests call with matcher definitions

crozzy:
This change attempts to make the matching process consistent across both the vulnerability matching and the manifest matching that is performed by the affectedManifests logic. It reduces the number of queries to 1 per matcher per vulnerability.


matchconstraint.go
```diff
@@ -0,0 +1,44 @@
...
+// MatchConstraint explains to the caller how a search for a package's vulnerability should
+// be constrained.
+//
+// for example if sql implementation encounters a DistributionDID constraint
+// it should create a query similar to "SELECT * FROM vulnerabilities WHERE package_name = ? AND distribution_did = ?"
+type MatchConstraint int
+
+//go:generate go run golang.org/x/tools/cmd/stringer -type MatchConstraint
+
+const (
```
crozzy, line 12: TODO: If these Constraints stay here, they need more specificity in their naming. `claircore.PackageName` has no indication that it's a contraint.

datastore/postgres/migrations/indexer/09-versionrange.sql
```diff
@@ -0,0 +1,6 @@
+DO $$
```
crozzy, line 1: TODO: Add comment about why this is needed and why it might already exist

pkg/omnimatcher/omnimatcher.go
```diff
@@ -24,6 +25,7 @@ var defaultOmniMatcher = []driver.Matcher{
 	&python.Matcher{},
 	&rhel.Matcher{},
 	&ubuntu.Matcher{},
+	rhcc.Matcher,
```
crozzy, line 28: :eyes:

datastore/postgres/affectedmanifest.go
```diff
@@ -18,336 +16,241 @@ import (
...
-		idStr := strconv.FormatInt(id, 10)
-		pkg.ID = idStr
-		if nKind != nil {
-			pkg.NormalizedVersion.Kind = *nKind
-			for i, n := range nVer.Elements {
-				pkg.NormalizedVersion.V[i] = n.Int
+		var ex goqu.Ex
+		switch m {
+		case claircore.PackageModule:
+			ex = goqu.Ex{"package.module": v.Package.Module}
```
crozzy, line 55: Needs a better explanation but essentially the RHEL matcher is doing something weird where it checks the package.module (and the package module is an empty string, all the other Matchers behave as you'd expect, ask to query on a column that actually have information about.

datastore/postgres/affectedmanifest.go
```diff
@@ -18,336 +16,241 @@ import (
...
-				Package:      &p,
-				Distribution: pr.Distribution,
-				Repository:   pr.Repository,
-			})
+	if vp.VersionAuthoritative != nil {
+		kind, lower, upper := rangefmt(v.Range)
+		if kind != nil {
+			exps = append(exps, goqu.And(
+				goqu.C("norm_kind").Eq(kind),
+				goqu.L("VersionRange('"+lower+"','"+upper+"') @> package.norm_version"),
```
crozzy, line 122: TODO: check the "version equals upper range" scenario

datastore/postgres/affectedmanifest_test.go
```diff
@@ -0,0 +1,411 @@
...
+			ctx:          ctx,
+			ir:           ir,
+			v:            tc.vuln,
+			isVulnerable: tc.isVulnerable,
+		}
+		t.Run(tc.name, e2e.Run)
+	}
+}
+
+func TestQueryBuilder(t *testing.T) {
```
crozzy, line 353: TODO: Probably delete these, the logic is tested via `store.AffectedManifests`

indexer/store.go
```diff
@@ -69,7 +69,7 @@ type Querier interface {
 	IndexReport(ctx context.Context, hash claircore.Digest) (*claircore.IndexReport, bool, error)
 	// AffectedManifests returns a list of manifest digests which the target vulnerability
 	// affects.
-	AffectedManifests(ctx context.Context, v claircore.Vulnerability, f claircore.CheckVulnernableFunc) ([]claircore.Digest, error)
+	AffectedManifests(ctx context.Context, v claircore.Vulnerability) ([]claircore.Digest, error)
```
crozzy, line 72: Change of the interface here

crozzy:
Execution Time not representative as it's from a small instance, but the output is still useful
```
Nested Loop  (cost=9.03..29.56 rows=1 width=528) (actual time=0.034..0.048 rows=4 loops=1)
  ->  Nested Loop Left Join  (cost=8.88..29.34 rows=1 width=504) (actual time=0.030..0.041 rows=4 loops=1)
        ->  Nested Loop  (cost=8.73..29.13 rows=1 width=256) (actual time=0.027..0.035 rows=4 loops=1)
              ->  Nested Loop  (cost=8.59..28.44 rows=3 width=136) (actual time=0.022..0.027 rows=6 loops=1)
                    ->  Bitmap Heap Scan on package  (cost=4.30..9.77 rows=2 width=120) (actual time=0.014..0.015 rows=2 loops=1)
                          Recheck Cond: ((name = 'pcre2'::text) AND (module = ''::text))
                          Heap Blocks: exact=1
                          ->  Bitmap Index Scan on package_unique_idx  (cost=0.00..4.29 rows=2 width=0) (actual time=0.011..0.012 rows=2 loops=1)
                                Index Cond: ((name = 'pcre2'::text) AND (module = ''::text))
                    ->  Bitmap Heap Scan on manifest_index  (cost=4.29..9.32 rows=2 width=32) (actual time=0.003..0.004 rows=3 loops=2)
                          Recheck Cond: (package_id = package.id)
                          Heap Blocks: exact=2
                          ->  Bitmap Index Scan on manifest_index_unique  (cost=0.00..4.29 rows=2 width=0) (actual time=0.001..0.001 rows=3 loops=2)
                                Index Cond: (package_id = package.id)
              ->  Index Scan using repo_pkey on repo  (cost=0.15..0.21 rows=1 width=136) (actual time=0.001..0.001 rows=1 loops=6)
                    Index Cond: (id = manifest_index.repo_id)
                    Filter: (key = 'rhel-cpe-repository'::text)
        ->  Index Scan using dist_pkey on dist  (cost=0.15..0.21 rows=1 width=264) (actual time=0.001..0.001 rows=1 loops=4)
              Index Cond: (id = manifest_index.dist_id)
  ->  Index Scan using manifest_pkey on manifest  (cost=0.15..0.22 rows=1 width=40) (actual time=0.001..0.001 rows=1 loops=4)
        Index Cond: (id = manifest_index.manifest_id)
Planning Time: 0.401 ms
Execution Time: 0.101 ms
```

datastore/postgres/migrations/indexer/09-versionrange.sql
```diff
@@ -0,0 +1,9 @@
+-- The indexer DB needs a VersionRange function for the affectedManifest
+-- method, it already exists in the matcher DB so the existence check is
+-- imperative.
+DO $$
```
hdonnay: Trailing whitespace

hdonnay:
What's the thinking on splitting the tests across two different packages?

crozzy:
> What's the thinking on splitting the tests across two different packages?

The idea is that any tests that call "meta" code should be at the root level `/test/` directory. These are the kind of tests (that have been know as e2e tests) which aren't directly running the code in their packages, but indirectly running it by using parent functions (i.e. calling `github.com/quay/claircore/internal/matcher.Match()` instead of for example using the debian matcher directly). This was already done for the `rhcc` package so [this commit](https://github.com/quay/claircore/pull/1514/commits/892a624f5f4fb7924a05a1dd209f4adf78a3bada) was following that pattern.

test/rhel/rhcc_matcher_integration_test.go
```diff
@@ -21,14 +21,7 @@ import (
...
 )
 
-func TestMain(m *testing.M) {
-	var c int
-	defer func() { os.Exit(c) }()
-	defer integration.DBSetup()()
-	c = m.Run()
-}
-
-func TestMatcherIntegration(t *testing.T) {
```
hdonnay, line 31: Why is the `TestMain` removed? It should still be needed.

Can this be split into a commit each for the `debian` and `rhel` changes?

test/mock/datastore/mocks.go
hdonnay, line 1: I'd like the changes to the mocks that seem to be (are?) just from running a new version of the tool split out into a new commit.

hdonnay (changes requested):

test/mock/indexer/mocks.go
```diff
@@ -42,18 +43,18 @@ func (m *MockStore) EXPECT() *MockStoreMockRecorder {
 }
 
 // AffectedManifests mocks base method.
-func (m *MockStore) AffectedManifests(arg0 context.Context, arg1 claircore.Vulnerability, arg2 claircore.CheckVulnernableFunc) ([]claircore.Digest, error) {
+func (m *MockStore) AffectedManifests(ctx context.Context, v claircore.Vulnerability) ([]claircore.Digest, error) {
```
crozzy, line 46: This signature has also changed

crozzy: Which IIRC is why the mocks were recreated

test/rhel/rhcc_matcher_integration_test.go
crozzy: split, and added back

datastore/postgres/affectedmanifest.go
```diff
@@ -4,340 +4,271 @@ import (
...
 		prometheus.HistogramOpts{
 			Namespace: "claircore",
-			Subsystem: "indexer",
-			Name:      "affectedmanifests_duration_seconds",
+			Subsystem: "datastore",
+			Name:      "getaffectedmanifests_duration_seconds",
 			Help:      "The duration of all queries issued in the AffectedManifests method",
 		},
 		[]string{"query"},
 	)
```
hdonnay, line 40: Changing the exported names on these is going to break dashboards.

It's fine as along as it's intended.

crozzy: I think it was expected considering we don't use it in our own dashboard and we consider metrics and non-API service. Will add a note in the commit message.

PR #1516, opened 2025-04-11
vex: allow timeout to pull down VEX archive to be configurable

crozzy:
The timeout was hardcoded to 2m, this remains the default value but users have the option to configure it to a different value using `updaters.config.rhel-vex.compressed_file_timeout`.

### DIFF
--- a/duration.go
+++ b/duration.go
@@ -1,0 +1,27 @@
+package claircore
+
+import (
+	"errors"
+	"time"
+)
+
+// Duration is a serializeable [time.Duration].
+type Duration time.Duration
+
+// UnmarshalText implements [encoding.TextUnmarshaler].
+func (d *Duration) UnmarshalText(b []byte) error {
+	dur, err := time.ParseDuration(string(b))
+	if err != nil {
+		return err
+	}
+	*d = Duration(dur)
+	return nil
+}
+
+// MarshalText implements [encoding.TextMarshaler].
+func (d *Duration) MarshalText() ([]byte, error) {
+	if d == nil {
+		return nil, errors.New("cannot marshal nil duration")
+	}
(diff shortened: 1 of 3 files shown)


duration.go
```diff
@@ -0,0 +1,21 @@
+package claircore
```
crozzy, line 1: I have to believe there's a better place for this

duration.go
```diff
@@ -0,0 +1,21 @@
...
+	if err != nil {
+		return err
+	}
+	*d = Duration(dur)
+	return nil
+}
+
+// MarshalText implements [encoding.TextMarshaler].
+func (d *Duration) MarshalText() ([]byte, error) {
+	return []byte(time.Duration(*d).String()), nil
```
RTann, line 20: do we want to consider a `nil` check here?

crozzy: I think it'd be pretty difficult to arrive there (particularly with the current way of instantiating) but possible no the less.

duration.go
```diff
@@ -0,0 +1,27 @@
...
+// Duration is a serializeable [time.Duration].
+type Duration time.Duration
+
+// UnmarshalText implements [encoding.TextUnmarshaler].
+func (d *Duration) UnmarshalText(b []byte) error {
+	dur, err := time.ParseDuration(string(b))
+	if err != nil {
+		return err
+	}
+	*d = Duration(dur)
```
RTann, line 17: I guess a `nil` check would make sense here, too, but I'm looking at https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/time/time.go;l=1637 and even the standard library doesn't bother `nil` checking. Up to you

RTann (approved):

PR #1517, opened 2025-04-11
rpm: update package detector versions

crozzy:
After merging 1a72e8e2ab846a4ef08a4e4ba7829161db466be0 the package detector versions should have iterated, this commit corrects that. (Java omitted as it has been updated since).

### DIFF
--- a/nodejs/packagescanner.go
+++ b/nodejs/packagescanner.go
@@ -12,9 +12,9 @@ import (
 	"runtime/trace"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"github.com/quay/zlog"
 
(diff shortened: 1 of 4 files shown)


nodejs/packagescanner.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/quay/zlog"
```
RTann, line 16: Should this newline be deleted?

RTann (dismissed):
LGTM with one comment to consider

nodejs/packagescanner.go
crozzy: Yes, I guess if you're going to reorg imports you should go all the way

PR #1522, opened 2025-04-14
chore: update gitignore to exclude pg files

crozzy:
The claircore unittests will use embedded postgres instance, this leaves a few non-vsc-interesting files around the codebase. This change adds a pattern so git ignores those files.


hdonnay:
These shouldn't exist anymore, [database files are created under the user's cache directory](https://github.com/quay/claircore/blob/98d3e3fd6b115d49f5f62c6413685e7907c2587d/test/integration/engine.go#L38).

crozzy:
Nice, must have been left in the dir from before

PR #1523, opened 2025-04-15
chore: consume layers in reverse order for rpm spotchecks

crozzy:
It turns out that some images are now upgrading packages in subsequent layers, this meant our process of overwriting packages with the same name should have worked but it looks like the layers are ordered in reverse when returned via the api/containers/v1/repositories/registry API.


crozzy:
Closes: https://github.com/quay/claircore/issues/1521

crozzy:
https://github.com/quay/claircore/actions/runs/14478198608

hdonnay (approved):
LGTM

PR #1527, opened 2025-04-24
rpm: optimize file checking operation

crozzy:
While load testing the latest version it was found that the indexing latency was drastically increase (7x was observed). After some profile/blocking analysis it was found that the volume of calls to the `FileInstalledByRPM()` function were causing lock contention due to sheer number of paths the language detectors were checking. This change addresses that issue by creating `FileChecker` object before path iteration and then using it for the subsequent paths. After this change, no perceivable latency was added as a result of the RPM file checking.

### DIFF
--- a/gobin/gobin.go
+++ b/gobin/gobin.go
@@ -84,6 +84,11 @@ func (Detector) Scan(ctx context.Context, l *claircore.Layer) ([]*claircore.Pack
 	//
 	// Only create a single spool file per call, re-use for every binary.
 	var spool spoolfile
+	fc, err := rpm.NewFileChecker(ctx, l)
+	if err != nil {
+		return nil, fmt.Errorf("gobin: unable to check RPM db: %w", err)
+	}
+
 	walk := func(p string, d fs.DirEntry, err error) error {
 		ctx := zlog.ContextWithValues(ctx, "filename", d.Name())
 
@@ -108,11 +113,7 @@ func (Detector) Scan(ctx context.Context, l *claircore.Layer) ([]*claircore.Pack
 			return nil
 		}
 
-		isRPM, err := rpm.FileInstalledByRPM(ctx, l, p)
-		if err != nil {
-			return err
-		}
-		if isRPM {
+		if fc.IsRPM(p) {
 			zlog.Debug(ctx).
 				Str("path", p).
 				Msg("file path determined to be of RPM origin")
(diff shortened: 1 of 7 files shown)


gobin/gobin.go
```diff
@@ -84,6 +84,11 @@ func (Detector) Scan(ctx context.Context, l *claircore.Layer) ([]*claircore.Pack
 	//
 	// Only create a single spool file per call, re-use for every binary.
 	var spool spoolfile
+	fc, err := rpm.NewFileChecker(ctx, l)
+	if err != nil {
+		return nil, fmt.Errorf("ruby: unable to check RPM db: %w", err)
+	}
+
```
hdonnay, line 91: Copy-pasted error string?

java/packagescanner.go
```diff
@@ -146,13 +146,13 @@ func (s *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircor
 	ck := make([]byte, sha1.Size)
 	doSearch := s.root != nil
 	defer putBuf(buf)
+	fc, err := rpm.NewFileChecker(ctx, layer)
+	if err != nil {
+		return nil, fmt.Errorf("ruby: unable to check RPM db: %w", err)
+	}
```
hdonnay, line 152: Copy-pasted error string?

nodejs/packagescanner.go
```diff
@@ -92,12 +92,13 @@ func (s *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircor
 
 	ret := make([]*claircore.Package, 0, len(pkgs))
 	var invalidPkgs []string
+	fc, err := rpm.NewFileChecker(ctx, layer)
+	if err != nil {
+		return nil, fmt.Errorf("ruby: unable to check RPM db: %w", err)
+	}
+
```
hdonnay, line 99: Copy-pasted error string?

python/packagescanner.go
```diff
@@ -79,12 +79,12 @@ func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*clairco
 		return nil, fmt.Errorf("python: failed to find delicious egg: %w", err)
 	}
 	var ret []*claircore.Package
+	fc, err := rpm.NewFileChecker(ctx, layer)
+	if err != nil {
+		return nil, fmt.Errorf("ruby: unable to check RPM db: %w", err)
+	}
```
hdonnay, line 85: Copy-pasted error string?

crozzy: durrr

hdonnay (approved):
LGTM

PR #1528, opened 2025-04-24
datastore: gc enrichments

RTann:
I have StackRox installed using Scanner V4, which is based on Claircore. Here's the `EXPLAIN ANALYZE` for the deletion queries:

```
# EXPLAIN ANALYZE DELETE FROM vuln v1 USING
vuln v2
LEFT JOIN uo_vuln uvl
ON v2.id = uvl.vuln
WHERE uvl.vuln IS NULL
AND v2.updater = 'rhel-vex'
AND v1.id = v2.id;
                                                                      QUERY PLAN                                                                       
-------------------------------------------------------------------------------------------------------------------------------------------------------
 Delete on vuln v1  (cost=1068367.20..1887805.36 rows=0 width=0) (actual time=18785.634..18785.638 rows=0 loops=1)
   ->  Hash Join  (cost=1068367.20..1887805.36 rows=407282 width=18) (actual time=18785.631..18785.635 rows=0 loops=1)
         Hash Cond: (v1.id = v2.id)
         ->  Seq Scan on vuln v1  (cost=0.00..795983.61 rows=5168461 width=14) (actual time=0.035..0.036 rows=1 loops=1)
         ->  Hash  (cost=1063276.18..1063276.18 rows=407282 width=20) (actual time=18784.195..18784.198 rows=0 loops=1)
               Buckets: 524288  Batches: 1  Memory Usage: 4096kB
               ->  Hash Anti Join  (cost=200433.67..1063276.18 rows=407282 width=20) (actual time=18784.194..18784.196 rows=0 loops=1)
                     Hash Cond: (v2.id = uvl.vuln)
                     ->  Seq Scan on vuln v2  (cost=0.00..808904.76 rows=1395657 width=14) (actual time=535.456..13101.478 rows=1387909 loops=1)
                           Filter: (updater = 'rhel-vex'::text)
                           Rows Removed by Filter: 3771885
                     ->  Hash  (cost=101633.63..101633.63 rows=5683763 width=14) (actual time=1862.736..1862.737 rows=6723341 loops=1)
                           Buckets: 1048576  Batches: 16  Memory Usage: 27863kB
                           ->  Seq Scan on uo_vuln uvl  (cost=0.00..101633.63 rows=5683763 width=14) (actual time=0.055..721.353 rows=6723341 loops=1)
 Planning Time: 2.595 ms
 Execution Time: 18785.752 ms
(16 rows)

# EXPLAIN ANALYZE DELETE FROM enrichment e1 USING
enrichment e2
LEFT JOIN uo_enrich uen
ON e2.id = uen.enrich
WHERE uen.enrich IS NULL
AND e2.updater = 'nvd'
AND e1.id = e2.id;
                                                                  QUERY PLAN                                                                   
-----------------------------------------------------------------------------------------------------------------------------------------------
 Delete on enrichment e1  (cost=24746.16..88408.00 rows=0 width=0) (actual time=1076.427..1076.434 rows=0 loops=1)
   ->  Nested Loop  (cost=24746.16..88408.00 rows=19071 width=18) (actual time=1076.423..1076.430 rows=0 loops=1)
         ->  Hash Anti Join  (cost=24745.74..68030.52 rows=19071 width=20) (actual time=1076.421..1076.427 rows=0 loops=1)
               Hash Cond: (e2.id = uen.enrich)
               ->  Seq Scan on enrichment e2  (cost=0.00..36912.85 rows=204714 width=14) (actual time=0.962..640.351 rows=196623 loops=1)
                     Filter: (updater = 'nvd'::text)
                     Rows Removed by Filter: 291609
               ->  Hash  (cost=12517.55..12517.55 rows=703455 width=14) (actual time=274.576..274.578 rows=700355 loops=1)
                     Buckets: 1048576  Batches: 2  Memory Usage: 24600kB
                     ->  Seq Scan on uo_enrich uen  (cost=0.00..12517.55 rows=703455 width=14) (actual time=0.018..89.865 rows=700355 loops=1)
         ->  Index Scan using enrichment_pkey on enrichment e1  (cost=0.42..1.07 rows=1 width=14) (never executed)
               Index Cond: (id = e2.id)
 Planning Time: 0.452 ms
 Execution Time: 1079.150 ms
(14 rows)
```

Tested on a local installation with nothing else running, and it took 6 - 8 minutes (sorry I forget). I had someone run this in an active cluster, and it took 7 minutes. All of these were on SSD. I ran it again on an inactive installation (ie one with no other activity happening) and it took 50 minutes on HDD.

### DIFF
--- a/datastore/postgres/gc.go
+++ b/datastore/postgres/gc.go
@@ -15,6 +15,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/quay/zlog"
 	"golang.org/x/sync/semaphore"
+
+	"github.com/quay/claircore/libvuln/driver"
 )
 
 var (
@@ -54,6 +56,8 @@ const (
 // If a full GC is required run this method until the returned int64 value
 // is 0.
 func (s *MatcherStore) GC(ctx context.Context, keep int) (int64, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "datastore/postgres/GC")
+
 	// obtain update operations which need deletin'
 	ops, totalOps, err := eligibleUpdateOpts(ctx, s.pool, keep)
 	if err != nil {
@@ -81,20 +85,32 @@ func (s *MatcherStore) GC(ctx context.Context, keep int) (int64, error) {
 	cpus := int64(runtime.GOMAXPROCS(0))
 	sem := semaphore.NewWeighted(cpus)
 
-	errC := make(chan error, len(updaters))
+	errC := make(chan error, len(updaters[driver.VulnerabilityKind])+len(updaters[driver.EnrichmentKind]))
 
-	for _, updater := range updaters {
-		err = sem.Acquire(ctx, 1)
-		if err != nil {
-			break
+	for kind, us := range updaters {
+		var cleanup cleanupFunc
+		switch kind {
+		case driver.VulnerabilityKind:
+			cleanup = vulnCleanup
+		case driver.EnrichmentKind:
+			cleanup = enrichmentCleanup
+		default:
+			zlog.Error(ctx).Str("kind", string(kind)).Msg("unknown updater kind; skipping cleanup")
+			continue
 		}
-		go func(u string) {
-			defer sem.Release(1)
-			err := vulnCleanup(ctx, s.pool, u)
+		for _, u := range us {
+			err = sem.Acquire(ctx, 1)
 			if err != nil {
-				errC <- err
+				break
 			}
-		}(updater)
+			go func(cleanup cleanupFunc, u string) {
+				defer sem.Release(1)
+				err := cleanup(ctx, s.pool, u)
+				if err != nil {
+					errC <- err
+				}
+			}(cleanup, u)
+		}
 	}
 
 	// unconditionally wait for all in-flight go routines to return.
@@ -116,11 +132,11 @@ func (s *MatcherStore) GC(ctx context.Context, keep int) (int64, error) {
 
 // distinctUpdaters returns all updaters which have registered an update
 // operation.
-func distinctUpdaters(ctx context.Context, pool *pgxpool.Pool) ([]string, error) {
+func distinctUpdaters(ctx context.Context, pool *pgxpool.Pool) (map[driver.UpdateKind][]string, error) {
 	const (
 		// will always contain at least two update operations
 		selectUpdaters = `
-SELECT DISTINCT(updater) FROM update_operation;
+SELECT DISTINCT(updater), kind FROM update_operation;
 `
 	)
 	rows, err := pool.Query(ctx, selectUpdaters)
@@ -129,17 +145,20 @@ SELECT DISTINCT(updater) FROM update_operation;
 	}
 	defer rows.Close()
 
-	var updaters []string
+	updaters := make(map[driver.UpdateKind][]string)
 	for rows.Next() {
-		var updater string
-		err := rows.Scan(&updater)
+		var (
+			updater string
+			kind    driver.UpdateKind
+		)
+		err := rows.Scan(&updater, &kind)
 		switch err {
 		case nil:
 			// hop out
 		default:
 			return nil, fmt.Errorf("error scanning updater: %v", err)
 		}
-		updaters = append(updaters, updater)
+		updaters[kind] = append(updaters[kind], updater)
 	}
 	if rows.Err() != nil {
 		return nil, rows.Err()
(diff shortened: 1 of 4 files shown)


datastore/postgres/migrations/matcher/15-uo_enrich-fkey-index.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX ON uo_enrich (enrich);
+CREATE INDEX ON uo_enrich (uo);
```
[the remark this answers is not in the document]
RTann: This second one is actually unrelated to the deletion. I added it to match the `uo_vuln` table, honestly, but I think I can see the value:

https://github.com/quay/claircore/blob/main/datastore/postgres/migrations/matcher/05-uo_enrich-fkey.sql
```
ALTER TABLE uo_enrich
...
ADD CONSTRAINT uo_enrich_uo_fkey FOREIGN KEY (uo) REFERENCES update_operation (id) ON DELETE CASCADE,
...
```

This should improve the cascading delete operations (I think), but that being said, it wasn't there before, so perhaps we can continue to live without it. I'm open to removing it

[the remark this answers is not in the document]
crozzy: OK, it's a shame the `EXPLAIN` output doesn't include the DELETE CASCADE output because it's really hard to grasp how these indexes are being used. I'm fine if it's there to match the `uo_vuln` table, presumably they will help out the CASCADE DELETES.

crozzy (dismissed):

RTann:
@crozzy thanks for the approval. I want to do one more test prior to merge:

Deploy an older database with >25 million rows in the enrichment table, hit Indexer/Matcher with a bunch of scan requests, and then update to a version with the changes from this PR so I can see how long it takes to delete everything while under load. If that goes well, I'll merge

RTann:
@crozzy I ran the same test where I installed StackRox Scanner V4 DB with CPU limit of 2 and memory limit of 4GiB. I then gave it almost 2000 images to scan (15 at a time), and I found that the deletion took 45 - 48 minutes (ran it twice). So that's a nice speed up. I imagine more CPU will also work well. If you are ok with these changes, I'm ready to merge

datastore/postgres/gc.go
```diff
@@ -116,35 +132,37 @@ func (s *MatcherStore) GC(ctx context.Context, keep int) (int64, error) {
 
 // distinctUpdaters returns all updaters which have registered an update
 // operation.
-func distinctUpdaters(ctx context.Context, pool *pgxpool.Pool) ([]string, error) {
+func distinctUpdaters(ctx context.Context, pool *pgxpool.Pool) (int, map[string][]string, error) {
```
crozzy: Nit: I don't think it's worth returning the combined length of both maps, it should be O(1) to get the length of a map and can be done in the `GC` function.

crozzy: Why not make the key to the map `driver.UpdateKind`?

RTann: > Nit: I don't think it's worth returning the combined length of both maps, it should be O(1) to get the length of a map and can be done in the `GC` function.

Because I map `kind -> updaters`, I can't just do `len(updaters)` since that'd just be 2. I'd need something like `len(slices.Collect(maps.Values(updaters)))` (or I can just `for` loop it myself). I could switch it to `updater -> kind`, but I liked the idea of only deciding between which cleanup function to do just once instead of for each updater. I'm open to change, though

> Why not make the key to the map `driver.UpdateKind`?

Thought about it, but I think I'd have to do more type casting. Here I only have to do it once. I'm thinking of keeping it as-is unless you prefer otherwise

crozzy: You should be able to scan directly to the type no?
```go
func distinctUpdaters(ctx context.Context, pool *pgxpool.Pool) (map[driver.UpdateKind][]string, error) {
	const (
		// will always contain at least two update operations
		selectUpdaters = `
SELECT DISTINCT(updater), kind FROM update_operation;
`
	)
	rows, err := pool.Query(ctx, selectUpdaters)
	if err != nil {
		return nil, fmt.Errorf("error selecting distinct updaters: %v", err)
	}
	defer rows.Close()

	updaters := make(map[driver.UpdateKind][]string)
	for rows.Next() {
		var (
			updater string
			kind    driver.UpdateKind
		)
		err := rows.Scan(&updater, &kind)
		switch err {
		case nil:
			// hop out
		default:
			return nil, fmt.Errorf("error scanning updater: %v", err)
		}
		updaters[kind] = append(updaters[kind], updater)
	}
	if rows.Err() != nil {
		return nil, rows.Err()
	}
	return updaters, nil
}
```
and yes, I was talking about this, this way you are sure of number of potential errors rather than relying on a function that may be modified
```
errC := make(chan error, len(updaters[driver.VulnerabilityKind])+len(updaters[driver.EnrichmentKind]))
```

RTann: I think I was worried about

```
var (
	updater string
	kind    driver.UpdateKind
)
```

but it does seem to work, so I'll switch. Thanks

PR #1534, opened 2025-05-08
java: remove header check

RTann:
Some JAR files start off with a bash script, so the header may not quite match the ZIP magic number(s). This is ok, and it's still a valid JAR. The Go standard library accounts for this, too, and it can still parse these JARs. So, this PR removes the header check, and it just lets the standard library do its thing.

It'd be great to not copy/paste this stuff in both pacakgescanner.go and jar.go. Where would the best place be to track removing this duplication?

Tested with a JAR file which starts with a bash script, and we can parse it now.

### DIFF
--- a/java/jar/jar.go
+++ b/java/jar/jar.go
@@ -45,13 +45,6 @@ import (
 	"github.com/quay/zlog"
 )
 
-// Header is the magic bytes at the beginning of a jar.
-//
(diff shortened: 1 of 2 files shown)


BradLugo (approved):
Nice find. LGTM.

crozzy (approved):
LGTM

PR #1535, opened 2025-05-09
rpm: continue after finding nonexistent rpm sqlite DB

crozzy:
It is possible that a symlink can point to a nonexisting target (possibly from another layer) and it can cause the RPM package scanner to error out. This change checks for that error, logs it, and continues.


RTann (commented):
Do any package scanners which rely on this need to be bumped?

crozzy:
> Do any package scanners which rely on this need to be bumped?

I don't believe so because manifests just aren't indexed at the moment if they include these danglers

rpm/native_db.go
```diff
@@ -44,8 +45,13 @@ func getDBObjects[T ObjectResponse](ctx context.Context, sys fs.FS, db foundDB,
 	switch db.Kind {
 	case kindSQLite:
 		r, err := sys.Open(path.Join(db.Path, `rpmdb.sqlite`))
-		if err != nil {
-			return nil, fmt.Errorf("rpm: error reading sqlite db: %w", err)
+		switch {
+		case errors.Is(err, nil):
+		case errors.Is(err, fs.ErrNotExist):
+			zlog.Warn(ctx).Err(err).Msg("rpm: unable to open sqlite db")
```
RTann, line 51: can you confirm you see this logged when you run the test? Just want to make sure this is the expected error (not exist)

RTann (approved):
LGTM. just one comment

PR #1536, opened 2025-05-09
Load latest repo2cpe/container mapping files prior to embedded

daynewlee:
Currently ACS Scanner indexer will load the 'embedded' repository-to-cpe.json & container-name-repos-map.json and use them for 24 hours before attempting to pull the latest from Central/live URL.

This will lead to 'scanning discrepancies' when an indexer first starts up (such as when the HPA adds new replicas).

### DIFF
--- a/rhel/internal/common/updater.go
+++ b/rhel/internal/common/updater.go
@@ -38,10 +38,6 @@ func NewUpdater(url string, init interface{}) *Updater {
 		reqRate: rate.NewLimiter(interval, 1),
 	}
 	u.value.Store(init)
-	// If we were provided an initial value, pull the first token.
-	if !reflect.ValueOf(init).IsNil() {
-		u.reqRate.Allow()
-	}
 	return &u
 }
 


rhel/internal/common/updater.go
```diff
@@ -38,6 +39,7 @@ func NewUpdater(url string, init interface{}) *Updater {
 		reqRate: rate.NewLimiter(interval, 1),
 	}
 	u.value.Store(init)
+	u.firstUse.Store(true)
 	// If we were provided an initial value, pull the first token.
 	if !reflect.ValueOf(init).IsNil() {
 		u.reqRate.Allow()
```
crozzy, line 45: Couldn't this same functionality be achieved by not pulling a token here? Am I missing something obvious?

RTann (approved):
Seems good to me, but definitely wait for another review

hdonnay:
This is a very small change, and neither behavior is obviously wrong, so LGTM. Needs to be rebased to a single commit.

hdonnay (changes requested):
Need commits squashed

hdonnay (approved):

hdonnay:
Unrelated CI failure is fixed in `main` now.

PR #1540, opened 2025-05-12
test: discount packages without source information in rpm-manifest

crozzy:
/api/containers/v1/images/id/{img_id}/rpm-manifest now returns a lot of packages in the response for different architectures. These packages don't have any source information. This was causing test errors when trying to convert them into `claircore.Package`s. This patch ignores pacakges that don't contain full source information.

### DIFF
--- a/test/rpmtest/manifest.go
+++ b/test/rpmtest/manifest.go
@@ -45,6 +45,12 @@ func PackagesFromManifest(t *testing.T, items iter.Seq[catalog.RpmsItems]) iter.
 				t.Errorf("#%03d: (%#v) unable to determine version: %v", n, it, err)
 				continue
 			}
+
+			if it.SrpmName == "" && it.SrpmNEVRA == "" {
+				// This is a binary package with no source information. This can be discounted.
+				continue
+			}
+
 			// Newer images produced from Konflux shove all the source information
 			// into the SourceName and omit the SourceNEVRA. Try both.
 			//


crozzy:
https://github.com/quay/claircore/actions/runs/14983915737

crozzy:
Closes https://github.com/quay/claircore/issues/1538

hdonnay:
I tackled this by changing [this `cmp.Option`](https://github.com/quay/claircore/blob/a285e89f1887a50fcdf135ee02f5473090f9b058/test/rpmtest/options.go#L98) to also ignore the `.Source` field.

crozzy:
> I tackled this by changing [this `cmp.Option`](https://github.com/quay/claircore/blob/a285e89f1887a50fcdf135ee02f5473090f9b058/test/rpmtest/options.go#L98) to also ignore the `.Source` field.

But won't the test fail before it even gets to comparing if the source info isn't in the response, e.g:

```
manifest.go:54: #007: (catalog.RpmsItems{Architecture:"aarch64", GPG:"199e2f91fd431d51", Name:"binutils", NVRA:"binutils-2.35.2-63.el9.aarch64", Release:"63.el9", SrpmName:"", SrpmNEVRA:"", Summary:"A GNU collection of binary utilities", Version:"2.35.2", Epoch:"", Module:"", RepositoryHint:""}) unable to determine source version: rpmver: : missing separators
```

hdonnay (approved):

PR #1544, opened 2025-05-19
rhcc: account for labels.json file

crozzy:
Red Hat's new container image builder system does not include a
dockerfile at a known path in the image, this is replaces with a file
called labels.json file at a known location that contains the
information that the rhcc package and repository scanners need. This
change adds another rhcc package and repository scanner that read this
labels.json file. The dockerfile versions still exist but will be
eventually deprecated and removed.

### DIFF
--- a/rhel/rhcc/detector.go
+++ b/rhel/rhcc/detector.go
@@ -1,0 +1,176 @@
+package rhcc
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"strconv"
+	"time"
+
+	"github.com/quay/zlog"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/indexer"
+	"github.com/quay/claircore/pkg/rhctag"
+	"github.com/quay/claircore/toolkit/types/cpe"
+)
+
+var (
+	_ indexer.PackageScanner    = (*detector)(nil)
+	_ indexer.RepositoryScanner = (*repoDetector)(nil)
+
+	labelsFilepath = "root/buildinfo/labels.json"
+)
+
+type detector struct{}
+
+// Name implements [indexer.VersionedScanner].
+func (s *detector) Name() string { return "rhel_package_container_detector" }
+
+// Version implements [indexer.VersionedScanner].
+func (s *detector) Version() string { return "1" }
+
+// Kind implements [indexer.VersionedScanner].
+func (s *detector) Kind() string { return "package" }
+
+// Scan implements [indexer.PackageScanner].
+//
+// It performs a package scan on the given layer and returns all the RHEL
+// container identifying metadata by using the labels.json file.
+func (s *detector) Scan(ctx context.Context, l *claircore.Layer) ([]*claircore.Package, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "rhel/rhcc/detector.Scan")
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+	sys, err := l.FS()
+	if err != nil {
+		return nil, fmt.Errorf("rhcc: unable to open layer: %w", err)
+	}
+
+	labels, err := findJSONLabels(ctx, sys)
+	switch {
+	case errors.Is(err, nil):
+	case errors.Is(err, errNotFound):
+		return nil, nil
+	default:
+		return nil, err
+	}
+
+	// Check if required fields are present
+	if labels.Name == "" || labels.Architecture == "" || labels.Created.IsZero() {
+		zlog.Warn(ctx).Msg("required labels not found in labels.json")
+		return nil, nil
+	}
+
+	vr := strconv.FormatInt(labels.Created.Unix(), 10)
+	rhctagVersion, err := rhctag.Parse(vr)
+	if err != nil {
+		// A unix timestamp should always parse, so this is unexpected.
+		zlog.Warn(ctx).Err(err).Str("version", vr).Msg("failed to parse rhctag version")
+		return nil, nil
+	}
+
+	normVer := rhctagVersion.Version(true)
+	src := claircore.Package{
+		Kind:              claircore.SOURCE,
+		Name:              labels.Name,
+		Version:           vr,
+		NormalizedVersion: normVer,
+		PackageDB:         labelsFilepath,
+		Arch:              labels.Architecture,
+		RepositoryHint:    `rhcc`,
+	}
+	pkgs := []*claircore.Package{&src}
+
+	pkgs = append(pkgs, &claircore.Package{
+		Kind:              claircore.BINARY,
+		Name:              labels.Name,
+		Version:           vr,
+		NormalizedVersion: normVer,
+		Source:            &src,
+		PackageDB:         labelsFilepath,
+		Arch:              labels.Architecture,
+		RepositoryHint:    `rhcc`,
+	})
+	return pkgs, nil
+}
+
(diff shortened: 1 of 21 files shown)


rhel/rhcc/scanner.go
crozzy, line 1: This file is totally TBD until we get a concrete example of what the `labels.json` will look like in the wild. At that point we can start to refine this better.

crozzy: The eventual schema should be:
```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "type": "object",
    "properties": {
        "name": {
            "type": "string"
        },
        "org.opencontainers.image.created": {
            "type": "string",
            "format": "date-time"
        },
        "cpe": {
            "type": "string",
            "format": "uri"
        },
        "architecture": {
            "type": "string"
        }
    },
    "required": [
        "name",
        "org.opencontainers.image.created",
        "cpe",
        "architecture"
    ],
    "additionalProperties": {
        "type": "string"
    }
}
```

crozzy:
This should be ready for a once over

rhel/vex/parser.go
```diff
@@ -128,18 +128,19 @@ func newRepoCache() *repoCache {
...
 
-// Get attempts to find a repo in the cache identified by a WFN. If
-// it isn't found a repo is created and returned.
-func (rc *repoCache) Get(cpe cpe.WFN) *claircore.Repository {
-	if r, ok := rc.cache[cpe.String()]; ok {
+// Get attempts to find a repo in the cache identified by a WFN and
+// the repoKey. If it isn't found a repo is created and returned.
+func (rc *repoCache) Get(cpe cpe.WFN, repoKey string) *claircore.Repository {
+	cacheKey := cpe.String() + "|" + repoKey
+	if r, ok := rc.cache[cacheKey]; ok {
```
hdonnay: Should probably make this a `cacheKey` type and use a `unique.Handle[cacheKey]`: that'll avoid creating another big string to achieve the same result.

rhel/rhcc/rhcc.go
```diff
@@ -5,12 +5,24 @@
...
+// the image has been build via the legacy Red Hat build system. With newer images, reliable
+// repository CPEs are available and can be used in lieu of the GoldRepo.
+var (
+	GoldRepo = claircore.Repository{
+		Name: "Red Hat Container Catalog",
+		URI:  `https://catalog.redhat.com/software/containers/explore`,
+		Key:  RepositoryKey,
+	}
+	errNotFound = errors.New("not found")
+)
```
hdonnay: This should be a separate var block so that the docs render "right".

rhel/rhcc/matcher.go
```diff
@@ -26,16 +28,31 @@ func (*matcher) Name() string { return "rhel-container-matcher" }
...
+		if err != nil {
+			zlog.Warn(ctx).
+				Str("vulnerability name", vuln.Name).
+				Err(err).
+				Msg("unable to unbind repo CPE")
+			return false, nil
+		}
+		if !cpe.Compare(vuln.Repo.CPE, record.Repository.CPE).IsSuperset() && !rhel.IsCPESubstringMatch(record.Repository.CPE, vuln.Repo.CPE) {
+			return false, nil
+		}
```
hdonnay, line 54: This smells backwards, but I guess it's not: If this isn't a superset and isn't a Red Hat substring, then it's not relevant. The arguments being swapped also is odd but makes sense after thinking about it.

rhel/rhcc/detector.go
```diff
@@ -0,0 +1,186 @@
...
+		Arch:              arch,
+		RepositoryHint:    `rhcc`,
+	})
+	return pkgs, nil
+}
+
+// findJSONLabels reads and parses root/buildinfo/labels.json
+// Schema: see testdata/labels.schema.json
+func findJSONLabels(ctx context.Context, sys fs.FS) (map[string]string, error) {
+	labels := make(map[string]string)
```
hdonnay: This'll be a proper type later, right?

rhel/vex/parser.go
crozzy: yeah, `unique.Handle` is a good call but why would creating a custom type perform better than `unique.Handle[string]` and doing `k := unique.Make(cpe.String() + "|" + repoKey)`?

rhel/rhcc/detector.go
crozzy: Yeah, I will go ahead and make it one now

rhel/vex/parser.go
hdonnay: The string concatenation needs to allocate the length of the string and copy into it, a custom type (even composed of strings) is copying just the string headers.

rhel/rhcc/scanner.go
crozzy: checked in the schema

rhel/rhcc/detector.go
```diff
@@ -0,0 +1,173 @@
...
+	// Check if required fields are present
+	if labels.Name == "" || labels.Architecture == "" || labels.Created.IsZero() {
+		zlog.Warn(ctx).Msg("required labels not found in labels.json")
+		return nil, nil
+	}
+
+	vr := strconv.FormatInt(labels.Created.Unix(), 10)
+	rhctagVersion, err := rhctag.Parse(vr)
+	if err != nil {
+		// A unix timestamp should always parse, so this is unexpected.
```
hdonnay, line 70: Add a log, because this is going to be a pain to debug when this doesn't hold.

version.go
```diff
@@ -17,6 +17,8 @@ type Version struct {
 	V    [10]int32
 }
 
+var UnmatchableKind = "unmatchable"
```
hdonnay, line 22: Needs doc comment.

rhel/rhcc/detector.go
```diff
@@ -0,0 +1,173 @@
...
+// Version implements [indexer.VersionedScanner].
+func (s *detector) Version() string { return "1" }
+
+// Kind implements [indexer.VersionedScanner].
+func (s *detector) Kind() string { return "package" }
+
+// Scan performs a package scan on the given layer and returns all
+// the RHEL container identifying metadata by using the labels.json file.
+
+// Scan implements [indexer.PackageScanner].
```
hdonnay: Is this supposed to be one comment? It reads like a doc comment.

rhel/rhcc/detector.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	labels, err := findJSONLabels(ctx, sys)
+	switch {
+	case errors.Is(err, nil):
+	case errors.Is(err, errNotFound):
+		return nil, nil
+	default:
+		return nil, err
+	}
+	if labels.CPE != "" {
```
hdonnay: nit: invert the conditional, because that's house style.

crozzy:
I think there is a race condition in the tests that is triggered when a fixture is pulled from the cache (or reused in testing) https://github.com/quay/claircore/actions/runs/16357519449/job/46219135496

crozzy:
> I think there is a race condition in the tests that is triggered when a fixture is pulled from the cache (or reused in testing) https://github.com/quay/claircore/actions/runs/16357519449/job/46219135496

I think the issue here was that subtests (from different Test functions but from the same file) had the same name, so the fixture cache path was the same, causing a race condition `github.com_quay_claircore/rhel_rhcc/LabelsTest`, I modified the name of the subtest but maybe I should also go back and update the `GenerateFixture` to create a more specific cache path (in a separate PR).

hdonnay (approved):

PR #1546, opened 2025-05-20
dockerfile: add unquote tests

hdonnay:
I had this test file hanging around, so here it is cleaned up and committed.


crozzy (approved):
LGTM

PR #1547, opened 2025-05-22
internal: cache package

hdonnay:
This is a generic'd version of the `rpm.PathSet` cache.

Once merged, that implementation and the refcounting scheme in the fetcher can be replaced with `cache.Live`.

The package is named `cache` because we may want different cache strategies in the future, so we can just add more types here (e.g. `S3Fifo`, `LRU`).

### DIFF
--- a/internal/cache/live.go
+++ b/internal/cache/live.go
@@ -1,0 +1,96 @@
+//go:build go1.24
+
+package cache
+
+import (
+	"context"
+	"runtime"
+	"sync"
+	"weak"
+
+	"github.com/quay/claircore/internal/singleflight"
+)
+
+// Live is a cache that keeps a cached copy as long as the go runtime determines
+// the value is live.
+//
+// The Create member can be populated to simplify a call site, ala
+// [sync.Pool.New].
+// The zero value is safe to use.
+//
+// See also: [weak.Pointer].
+type Live[K comparable, V any] struct {
+	Create CreateFunc[K, V]
+	m      sync.Map
+	sf     singleflight.Group[K, *V]
+}
+
+// Get returns a pointer to the value associated with the key, calling the
+// "Create" function if populated and the "create" argument is nil.
+//
+// This function will panic if neither function is provided.
+func (c *Live[K, V]) Get(ctx context.Context, key K, create CreateFunc[K, V]) (*V, error) {
+	var fn CreateFunc[K, V]
+	switch {
+	case create != nil:
+		fn = create
+	case c.Create != nil:
+		fn = c.Create
+	default:
+		panic("programmer error: missing create function")
+	}
+	for {
+		// Try to load an existing value out of the cache.
+		value, ok := c.m.Load(key)
+		if !ok {
+			// No value found. Create a new value.
+			fn := func() (*V, error) {
+				// Eagerly check the Context so that every create function
+				// doesn't need the preamble.
+				//
+				// Do this because this goroutine may have gone around the loop
+				// multiple times and found entries in the map that had
+				// invalidated weak pointers, so the context may have expired.
+				if ctx.Err() != nil {
+					return nil, context.Cause(ctx)
+				}
+				v, err := fn(ctx, key)
+				if err != nil {
+					return nil, err
+				}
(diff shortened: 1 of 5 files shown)


hdonnay:
Undrafting this because I'd like to use it in the fetcher.

internal/singleflight/singleflight_test.go
```diff
@@ -0,0 +1,424 @@
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Based on https://github.com/golang/sync/commit/913fb63af28f446cd10c684ee847b5606cf328f7
```
crozzy, line 5: I think this link may be wrong?

internal/singleflight/singleflight.go
```diff
@@ -0,0 +1,209 @@
...
+		recovered = true
+	}
+}
+
+// Forget tells the singleflight to forget about a key.  Future calls
+// to Do for this key will call the function rather than waiting for
+// an earlier call to complete.
+func (g *Group[K, V]) Forget(key K) {
+	g.mu.Lock()
+	delete(g.m, key)
```
crozzy, line 207: More of a discussion, but a different generic-singleflight implementation adds a `forgotten` boolean field that is set here and checked during `doCall()`, I think that is to check if `Forget` was called while the same key was in-flight. This isn't in the mainline, so I think it makes sense to not include it, just wondered your views as you've been staring at it more.

crozzy (commented):

internal/singleflight/singleflight_test.go
BradLugo: Maybe this?
```suggestion
// Based on https://github.com/golang/sync/tree/913fb63af28f446cd10c684ee847b5606cf328f7/singleflight
```
or the specific file?
```suggestion
// Based on https://github.com/golang/sync/blob/913fb63af28f446cd10c684ee847b5606cf328f7/singleflight/singleflight_test.go
```
You could click the "Browse files" button on the commit page, though.

internal/singleflight/singleflight.go
BradLugo: It was removed from the mainline. I don't completely understand it yet, but here's the context: https://redirect.github.com/golang/go/issues/55343

internal/cache/live.go
BradLugo, line 1: How might we test this?

internal/singleflight/singleflight_test.go
hdonnay: Yeah, I just copied the link to the commit; I'll swap this to a nicer link.

internal/cache/live.go
hdonnay: The checked-in test not enough, you think?

internal/singleflight/singleflight.go
hdonnay: 🤷 I'm just going with what's in the repo.

internal/cache/live.go
BradLugo: For checking the correctness of `Live` in particular (e.g., tracking weak pointers, expected deletion), I don't think so. But maybe the implementation speaks for itself? I'm not sure. What'd you think?

hdonnay: ah, I see

internal/cache/live_test.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	wp := weak.Make(b)
+
+	t.Logf("a: %p, b: %p", a, b)
+	if a != b {
+		t.Fail()
+	}
+	a, b = nil, nil
+
+	for range 2 { // Need two cycles:
```
BradLugo, line 39: Do you know the reason?

hdonnay: Not entirely; I assume it has to do with marking the pointers dead, then running any cleanups

BradLugo: I double checked this out of curiosity, and it looks like a single `runtime.GC()` call works on `go version go1.25.0 darwin/arm64`

PR #1555, opened 2025-05-29
rhel: account for the rpmmod being included in the component's PURL

crozzy:
Red Hat prodsec have decided to move from declaring a component's RPM module in the VEX relationships to directly including the module string in the component's purl as a qualifier. This change accounts for that change (while still supporting the original logic path as it's currently split-brain).


hdonnay (approved):
LGTM

crozzy:
Another example: https://security.access.redhat.com/data/csaf/v2/vex/2023/cve-2023-45143.json

PR #1557, opened 2025-05-30
Add MultiCoalescer

RTann:
The new `(*language.coalesce).Coalesce` function does not work well when there are multiple packages under the same Package DB. This is common for Go and Java packages, so this solution is insufficient for those two ecosystems.

This change brings back the old, language-specific Coalesce method but puts it in the common `language` package, as all of the old coalescers for languages were essentially the same (see the change which added the `language` package [here](https://github.com/quay/claircore/pull/1275). This will just be used by Go and Java at this time.

Also, instead of mapping from package db -> package, `(*language.coalesce).Coalesce` now maps from filepath -> package. That's because the `Filepath` field is a more accurate representation of where the package is located than `PackageDB`. For example, Python packages have a `PackageDB` set to the directory in which the Egg or Wheel is located instead of the location of Egg/Wheel


crozzy:
I think this moot now and we should investigate another solution.

PR #1334, opened 2024-05-03
jsonblob: Support individual record iteration

jvdm:
Add a per-record iterator interface using rangefunc-style to `jsonblob`. This conveniently aligns `jsonblob` to be used with the Vulnerability and Enrichment datastore iterator interfaces.

PR #1559, opened 2025-05-31
jsonblob: implement Load as documented

BradLugo:
- Changes current `Load()` to `NewLoader()`
- Implements `Load()` as documented, i.e., will load provided `io.Reader` in a `Store`
- Moves `Loader` types and methods to a separate file
- Fixes a bug in `Store.UpdateEnrichments()` where `s.Enrichments` weren’t being set
- Fixes a bug in `Store.UpdateVulnerabilities()` where `s.Vuln` wasn’t being set

### DIFF
--- a/libvuln/jsonblob/jsonblob.go
+++ b/libvuln/jsonblob/jsonblob.go
@@ -45,85 +45,43 @@ type Store struct {
 }
 
 // Load reads in all the records serialized in the provided [io.Reader].
-func Load(ctx context.Context, r io.Reader) (*Loader, error) {
-	l := Loader{
-		dec: json.NewDecoder(r),
-		cur: uuid.Nil,
-	}
-	return &l, nil
-}
-
-// Loader is an iterator that returns a series of [Entry].
-//
-// Users should call [*Loader.Next] until it reports false, then check for
-// errors via [*Loader.Err].
-type Loader struct {
-	err error
-	e   *Entry
-
-	dec  *json.Decoder
-	next *Entry
-	de   diskEntry
-	cur  uuid.UUID
-}
-
-// Next reports whether there's an [Entry] to be processed.
-func (l *Loader) Next() bool {
-	if l.err != nil {
-		return false
+func Load(ctx context.Context, r io.Reader) (*Store, error) {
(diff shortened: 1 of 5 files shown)


libvuln/jsonblob/jsonblob.go
```diff
@@ -45,85 +45,43 @@ type Store struct {
...
-			l.next = &Entry{}
-			l.next.Updater = l.de.Updater
-			l.next.Fingerprint = l.de.Fingerprint
-			l.next.Date = l.de.Date
+	// TODO(DO NOT MERGE): ~~This implementation might be a bit naive. Currently,
+	//  it basically copies [OfflineImport]. We could probably do some custom
+	//  parsing such that it basically just decodes the json into the
+	//  appropriate [Store] fields.~~
+	//  Actually, this might be the way.
+	//  [OfflineImport]: https://github.com/quay/claircore/blob/126f688bb11220fb34708719be91952dc32ff7b1/libvuln/updates.go#L17-L74
```
BradLugo, line 64: Couple of other thoughts here - we could either:
1. Change the documentation to be specific to `Loader` instead of `Store`
2. Merge https://github.com/quay/claircore/pull/1334, then use `Unmarshal()` instead of using a `Loader` to load the data.

BradLugo, line 64: Couple of other thoughts here - we could either:
1. Change the documentation to be specific to `Loader` instead of `Store`
2. Merge https://github.com/quay/claircore/pull/1334, then use `Unmarshal()` instead of using a `Loader` to load the data.

BradLugo, line 64: Couple of other thoughts here - we could either:
1. Change the documentation to be specific to `Loader` instead of `Store`
2. Merge https://github.com/quay/claircore/pull/1334, then use `Unmarshal()` instead of using a `Loader` to load the data.

hdonnay (changes requested):
@BradLugo What do you want to do here? There's a `DO NOT MERGE` comment in here.

hdonnay:
is this still needed with #1986 merged?

PR #1561, opened 2025-06-02
rhel: remove containerapi usage

crozzy:
The RHEL repository scanner had logic to consult the Red Hat container API to get repository CPEs if it wasn't able to get them from the content-manifests files. Now we will have 2 other ways to find this information it is sub-optimal to use an external API to do this as well. Using an external API in the indexer process is a very maintenence heavy burden.


hdonnay (approved):
LGTM

rip bozo 💯💯

PR #1562, opened 2025-06-04
Revert: all: account for language package overwrites

crozzy:
This reverts commit ff0aeab9703937d4b2a76b23676e0ebcd77226b0.


RTann (approved):

RTann:
Reverts https://github.com/quay/claircore/pull/1275 as there was an issue where a PackageDB references multiple packages

crozzy:
For future people: The fact that the `PackageDB` is not unique across packages (particularly in the go and java ecosystems) meant that only one package would show up in index reports for packages that came from the same binary (as go deps for example).

PR #1563, opened 2025-06-04
kev: CISA KEV enricher

RTann:


enricher/kev/kev.go
```diff
@@ -0,0 +1,279 @@
...
+func (e *Enricher) Configure(_ context.Context, f driver.ConfigUnmarshaler, c *http.Client) error {
+	var cfg Config
+	e.c = c
+	if err := f(&cfg); err != nil {
+		return err
+	}
+	e.feed = defaultFeed
+	if cfg.Feed != nil {
+		if !strings.HasSuffix(*cfg.Feed, ".json") {
+			return fmt.Errorf("URL not pointing to JSON: %q", *cfg.Feed)
```
crozzy: Nit: it is possible to serve a JSON file with the `.json` extension right? Nitty enough to ignore probably

enricher/kev/kev.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	return nil
+}
+
+// Name implements driver.Enricher and driver.EnrichmentUpdater.
+func (*Enricher) Name() string { return name }
+
+// FetchEnrichment implements driver.EnrichmentUpdater.
+func (e *Enricher) FetchEnrichment(ctx context.Context, hint driver.Fingerprint) (io.ReadCloser, driver.Fingerprint, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "enricher/kev/Enricher/FetchEnrichment")
```
crozzy:
```suggestion
	ctx = zlog.ContextWithValues(ctx, "component", "enricher/kev/Enricher.FetchEnrichment")
```

enricher/kev/kev.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	if hint != "" {
+		// Note: Though the default URL returns an etag, the server does not seem to respond
+		// to the If-None-Match header. It seems like it does respond to If-Modified-Since, though,
+		// so the timestamp is used as the hint.
+		req.Header.Set("If-Modified-Since", string(hint))
+	}
+	res, err := e.c.Do(req)
+	if err != nil {
+		return nil, hint, err
```
crozzy: I think if we follow the typical updater flow these errors should be wrapped like: 
```go
return nil, hint, fmt.Errorf("kev: unable to do request: %w", err)
```
I get that that's not how the cvss enricher looks like now

enricher/kev/kev.go
```diff
@@ -0,0 +1,279 @@
...
+	var success bool
+	defer func() {
+		if !success {
+			if err := out.Close(); err != nil {
+				zlog.Warn(ctx).Err(err).Msg("unable to close spool")
+			}
+		}
+	}()
+
+	// When originally created, the file was around 1.1MB, so
```
crozzy: nit
```suggestion
	// The file at the time of writing was around 1.1MB, so
```

enricher/kev/kev.go
```diff
@@ -0,0 +1,279 @@
...
+	zlog.Debug(ctx).
+		Str("hint", string(hint)).
+		Msg("using new hint")
+
+	return out, hint, nil
+}
+
+// ParseEnrichment implements driver.EnrichmentUpdater.
+func (e *Enricher) ParseEnrichment(ctx context.Context, rc io.ReadCloser) ([]driver.EnrichmentRecord, error) {
+	ctx = zlog.ContextWithValues(ctx, "component", "enricher/kev/Enricher/ParseEnrichment")
```
crozzy:
```suggestion
	ctx = zlog.ContextWithValues(ctx, "component", "enricher/kev/Enricher.ParseEnrichment")
```

enricher/kev/kev.go
```diff
@@ -0,0 +1,279 @@
...
+
+	m := make(map[string][]json.RawMessage)
+	erCache := make(map[string][]driver.EnrichmentRecord)
+
+	for id, v := range r.Vulnerabilities {
+		t := make(map[string]struct{})
+		ctx := zlog.ContextWithValues(ctx, "vuln", v.Name)
+
+		for _, elem := range []string{
+			v.Description,
```
crozzy, line 209: Probably not a discussion for this PR but I really feel like we shouldn't be using `Description` anymore, the false positive possibilities seem too high for the potential benefits.

enricher/kev/kev.go
```diff
@@ -0,0 +1,279 @@
...
+			Tags:       []string{vuln.CVEID},
+			Enrichment: enrichment,
+		})
+	}
+
+	return recs, nil
+}
+
+// Enrich implements driver.Enricher.
+func (e *Enricher) Enrich(ctx context.Context, g driver.EnrichmentGetter, r *claircore.VulnerabilityReport) (string, []json.RawMessage, error) {
```
crozzy, line 198: Are the the enrichers basically duplicating this logic? Should we thing about making it generic instead of copying functionality?

crozzy (changes requested):
A few small things, two larger points:

- In the matcher v2 world I think this will need to be all reworked to fit into that model
- It's more off-by-default code (like the EPSS code), this has traditionally been shied away from but I think as ACS is exercising it, it probably allays that fear.

enricher/kev/kev.go
RTann: True. I guess you can also use lie and your `.json` file is not valid JSON. I'll remove this

enricher/kev/kev.go
RTann: Yeah pretty much. I can create a ticket to capture this

RTann: https://issues.redhat.com/browse/CLAIRDEV-172

enricher/kev/kev.go
RTann: I mentioned that in https://issues.redhat.com/browse/CLAIRDEV-172

enricher/kev/kev.go
```diff
@@ -0,0 +1,275 @@
...
+var (
+	_ driver.Enricher          = (*Enricher)(nil)
+	_ driver.EnrichmentUpdater = (*Enricher)(nil)
+
+	defaultFeed *url.URL
+)
+
+const (
+	// Type is the type of data returned from the Enricher's Enrich method.
+	Type = `message/vnd.clair.map.vulnerability; enricher=clair.kev schema=https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities_schema.json`
```
crozzy: Is this the schema that actually gets returned? It seems from the comments that the schema is something like `claircore.enricher.kev.Entry`?

crozzy (commented):
Looks good, just one more comment

enricher/kev/kev.go
RTann: I figured this had to be a link to something. Can I just say it's `claircore.enricher.kev.Entry`? Looks like for EPSS, we did `schema=none`. Which would you prefer?

RTann: Changed it to `schema=none` to match EPSS. I also consdiered just using the actual CISA KEV schema, but it's not quite what we want (we add metadata only found in the "root" of the data)

Issue #1567, opened 2025-06-10
java: maven search alternative


kamphaus:
The jar package scanner looks up jars in search.maven.org by their sha1 hash for better accuracy. However that site seems to be down for longer periods of time: https://status.maven.org/

Would it be possible to use https://central.sonatype.com/solrsearch/select instead of https://search.maven.org/solrsearch/select?
The parameters and responses used by java/packagescanner.go are the same.
The only incompatibility is that central.sonatype.com's q parameter does not support " quotes.

For example for a given hash 704ebb292d4b7caaaf3a02630ad4b64c9c2b0f9f the current default request would be
https://search.maven.org/solrsearch/select?q=1%3A%22704ebb292d4b7caaaf3a02630ad4b64c9c2b0f9f%22&wt=json finds 1 doc when search.maven.org is working.

However changing the URL to central.sonatype.com:
https://central.sonatype.com/solrsearch/select?q=1%3A%22704ebb292d4b7caaaf3a02630ad4b64c9c2b0f9f%22&wt=json does not find a document.

Without the quote both searches find the same document:
https://central.sonatype.com/solrsearch/select?q=1%3A704ebb292d4b7caaaf3a02630ad4b64c9c2b0f9f&wt=json finds 1 doc.
https://search.maven.org/solrsearch/select?q=1%3A704ebb292d4b7caaaf3a02630ad4b64c9c2b0f9f&wt=json finds 1 doc.

hdonnay:
We can try removing the quotes, as the API docs don't really say what the type of the `1` field is, and solr seems to have opinions on the difference between "text" and "string".

I'm wary of making the URL change, because the [docs](https://central.sonatype.org/search/rest-api-guide/) use `search.maven.org` and say nothing about `central.sonatype.com`.

kamphaus:
> We can try removing the quotes, as the API docs don't really say what the type of the `1` field is, and solr seems to have opinions on the difference between "text" and "string".
> 
> I'm wary of making the URL change, because the [docs](https://central.sonatype.org/search/rest-api-guide/) use `search.maven.org` and say nothing about `central.sonatype.com`.

That's fine.
Since removing the quotes seems to work for both URLs, users can then use the following existing config option to chose whether to use the default `search.maven.org` or `central.sonatype.com`: https://github.com/quay/claircore/blob/cc6d0a8ecec55c21146b842a414b0a60db6a833d/java/packagescanner.go#L55

PR #1568, opened 2025-06-10
java: change search query

hdonnay:
This change constructs the solr search as a string field instead of a text field. Other implementations of the Cental search API seem to like it this way instead of the other.

See-also: quay/claircore#1567


RTann (commented):
do we have any testing for this to confirm the query returns some kind of valid result?

PR #1572, opened 2025-06-16
java: add doc about alternate URL

hdonnay:
Closes: #1567

### DIFF
--- a/java/packagescanner.go
+++ b/java/packagescanner.go
@@ -42,6 +42,9 @@ var (
 // DefaultSearchAPI is a maven-like REST API that may be used to do
 // reverse lookups based on an archive's sha1 sum.
 //
+// Experimentally, the host "central.sonatype.com" also works. This URL is not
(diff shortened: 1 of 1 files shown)


java/packagescanner.go
```diff
@@ -42,6 +42,9 @@ var (
 // DefaultSearchAPI is a maven-like REST API that may be used to do
 // reverse lookups based on an archive's sha1 sum.
 //
+// Experimentally, the host "central.sonatype.com" also works. This URL is not
+// mentioned in the Maven Central documentation, though.
```
RTann, line 46: when I looked up Maven Central, I ended up getting https://central.sonatype.org/ and https://central.sonatype.com/

RTann (approved):

java/packagescanner.go
hdonnay: Yeah the HTTP redirects, but they're different hosts.

PR #1588, opened 2025-07-21
dnf: skip checking DNF DB if from_dnf_hint isn't true

crozzy:
For older RH images that include content_sets but don't have the from_dnf_hint we cannot rely on the repoids that are extracted from the DNF history DB. This change accounts for that logic.


rhel/repositoryscanner.go
```diff
@@ -326,52 +332,17 @@ func bugURL(id string, err error) string {
 }
 
 // RepoidsFromContentSets returns a slice of repoids, as discovered by examining
-// information contained within the container.
+// information contained within the container. Found repoids will need to be translated
+// using a mapping file provided by Red Hat's PST team.
```
crozzy, line 335: Kind of tempted to nix this function altogether but it does some subtle logic

rhel/repositoryscanner.go
```diff
@@ -199,6 +199,14 @@ func (r *RepositoryScanner) Scan(ctx context.Context, l *claircore.Layer) ([]*cl
 		return nil, fmt.Errorf("rhel: unable to open layer: %w", err)
 	}
 
+	useDNFData := false
+	man, err := getContentManifest(ctx, sys)
+	if err != nil {
+		return nil, fmt.Errorf("rhel: unable to get content manifest: %w", err)
+	}
+	useDNFData = man == nil || man.FromDNFHint
+	fmt.Println("useDNFData", useDNFData)
```
hdonnay: Leftover `fmt`

hdonnay (changes requested):

PR #1589, opened 2025-07-23
postgres: remove batching helper

hdonnay:
This removes the batching helper, which still had a timeout in it. This does batching directly in the database methods, with flushing for the ones that are expected to be large sets.


datastore/postgres/indexpackage.go
```diff
@@ -98,72 +99,39 @@ func (s *IndexerStore) IndexPackages(ctx context.Context, pkgs []*claircore.Pack
...
-
-	skipCt = 0
-	// make package scan artifacts
-	mBatcher = microbatch.NewInsert(tx, 500, time.Minute)
-
-	start = time.Now()
 	for _, pkg := range pkgs {
 		if pkg.Name == "" {
 			skipCt++
+			// Original code has this not continue, but that seems wrong...
```
crozzy, line 122: Nit: I'd remove this comment as it seems more like a review comment (I also agree, it seems like we should `continue` in this case and not attempt to queue the insert).

crozzy: Not really a part of this change but do you think we should log here? I guess the question is; what would we log?

datastore/postgres/indexpackage.go
```diff
@@ -98,72 +99,39 @@ func (s *IndexerStore) IndexPackages(ctx context.Context, pkgs []*claircore.Pack
...
-	if err != nil {
-		return fmt.Errorf("failed to create statement: %w", err)
-	}
+	ctx = zlog.ContextWithValues(ctx, "component", "datastore/postgres/IndexerStore.IndexPackages")
 
+	var batch, assocBatch pgx.Batch
 	skipCt := 0
-
-	start := time.Now()
-	mBatcher := microbatch.NewInsert(tx, 500, time.Minute)
```
crozzy, line 121: I suppose we're assuming in the new code that most likely there will always be less than 500 packages, so there is no need to manually flush during the queuing?

datastore/postgres/updatevulnerabilities.go
```diff
@@ -270,31 +276,29 @@ func (s *MatcherStore) updateVulnerabilities(ctx context.Context, updater string
...
-		if err != nil {
-			err = fmt.Errorf("failed to queue vulnerability: %w", err)
-			return false
-		}
+		batch.Queue(assoc, hashKind, hash, uoID)
 
-		err = mBatcher.Queue(ctx, assoc, hashKind, hash, uoID)
-		if err != nil {
-			err = fmt.Errorf("failed to queue association: %w", err)
+		if ct := batch.Len(); ct < 1000 {
```
crozzy, line 289: Was there an intuition that 2000 was too large?

Would this make sense as a configured value? (could be annoying to thread through, but it is sort of dependent on instance size).

crozzy (commented):
Looks good, definitely an improvement code-wise, just a handful of

datastore/postgres/indexpackage.go
hdonnay: Yeah, I don't know what to do: the tests seem happy.

datastore/postgres/indexpackage.go
hdonnay: that's my thinking.

datastore/postgres/updatevulnerabilities.go
hdonnay: The value in the previous version was `2000` but then also queuing 2 statements per loop, which got me confused about how many statements that _actually_ represented, so I just whacked 1000 in here.

crozzy (approved):

PR #1593, opened 2025-07-29
rhel: remove remaining OVAL code

crozzy:
These remnants were left over from the earlier cleanup work.


test/ovaldebug/main.go
```diff
@@ -1,95 +0,0 @@
-// Ovaldebug is a helper for debugging the ovalutil package.
```
crozzy, line 1: I was tempted to just remove the RHEL part of this, but without that, it essentially does nothing

PR #1597, opened 2025-08-08
cvss: switch to NVD 2.0 JSON feeds

RTann:


enricher/cvss/testdata/feed.json
RTann, line 1: I made this a simple 3 item file with only two relevant CVEs. The third is "Rejected".

datastore/postgres/migrations/matcher/16-delete-cvss-enrichments.sql
```diff
@@ -0,0 +1,15 @@
...
+    enrichment e2
+    LEFT JOIN uo_enrich ue
+        ON e2.id = ue.enrich
+    WHERE ue.enrich IS NULL
+    AND e2.updater = 'clair.cvss'
+    AND e1.id = e2.id; 
+
+-- Clean up after deletions
+ANALYZE enrichment;
+ANALYZE uo_enrich;
```
crozzy: This is actually quite important, the subsequent updates took ~30mins when the tables weren't reanalyzed (as opposed to ~1 min when they are)

enricher/cvss/cvss.go
```diff
@@ -32,15 +32,15 @@ var (
 
 const (
 	// Type is the type of data returned from the Enricher's Enrich method.
-	Type = `message/vnd.clair.map.vulnerability; enricher=clair.cvss schema=https://csrc.nist.gov/schema/nvd/feed/1.1/cvss-v3.x.json`
+	Type = `message/vnd.clair.map.vulnerability; enricher=clair.cvss schema=https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema`
```
RTann, line 35: The new schema did not seem to have a separate file to reference just the CVSS aspect, so I just linked to the single file that has everything. Not sure if that's desired or not. Just wanted to call that out

crozzy: Seems ok to me, I don't think anyone is using it programmatically

datastore/postgres/enrichment.go
```diff
@@ -114,7 +124,6 @@ VALUES
 			WHERE
 				hash_kind = $1
 				AND hash = $2
-				AND updater = $3
```
RTann, line 117: I believe you said this is ok because `(hash_kind, hash)` is already unique. Makes sense to me

datastore/postgres/enrichment.go
```diff
@@ -21,6 +21,16 @@ import (
 	"github.com/quay/claircore/libvuln/driver"
 )
 
+// Tuneable knobs for bulk enrichment processing.
```
RTann: is there a desire to make this configurable (perhaps in another PR)?

RTann:
@crozzy since I'm the original author I can't approve, so feel free to approve and merge once you are ready

datastore/postgres/enrichment.go
crozzy: Not at the moment, I think I'll remove this comment

datastore/postgres/enrichment.go
crozzy: yeah, should be using `enrichment_hash_kind_hash_idx` index which is unique.

crozzy (dismissed):
LGTM but I also wrote most of it

enricher/cvss/feed.go
```diff
@@ -61,14 +66,36 @@ func (f *itemFeed) WriteCVSS(ctx context.Context, w io.Writer) error {
...
+			if cvss.Type != "Primary" {
+				continue
+			}
+			enrichment = cvss.CVSS
+			break
+		}
+		if enrichment == nil {
+			// Did not find a CVSS 3.x from the primary source.
 			skip++
 			continue
```
hdonnay: This could be a single loop over a sequence of `Metric`s instead of two for loops that will always be at least partially evaluated.

```
seq := func(yield func(metric) bool) {
    for _, m := range v.CVE.Metrics.V31{
        if !yield(m){
            return
        }
    }
    for _, m := range v.CVE.Metrics.V30{
        if !yield(m){
            return
        }
    }
}
var enrichment json.RawMessage
for m := range seq {
	if cvss.Type == "Primary" {
    	enrichment = cvss.CVSS
    	break
	}
}
if enrichment == nil {
    // ...
}
```

crozzy: updated

enricher/cvss/feed.go
```diff
@@ -4,44 +4,48 @@ import (
...
-	Count int             `json:"CVE_data_numberOfCVEs,string"`
-	Items json.RawMessage `json:"CVE_Items"`
+type vulnFeed struct {
+	Total int             `json:"totalResults"`
+	Vulns json.RawMessage `json:"vulnerabilities"`
 }
 
-// This is an envelope type so we can get at the cvssv3 object way in there.
-type cve struct {
+// This is an envelope type so we can get at the cvssv3 objects way in there.
```
hdonnay:
```suggestion
```

enricher/cvss/feed.go
```diff
@@ -61,14 +65,41 @@ func (f *itemFeed) WriteCVSS(ctx context.Context, w io.Writer) error {
...
+			}
+		}
+		var enrichment json.RawMessage
+		seq(func(m metric) bool {
+			if m.Type == "Primary" {
+				enrichment = m.CVSS
+				return false
+			}
+			return true
+		})
```
hdonnay:
```suggestion
		for m := range seq {
			if m.Type == "Primary" {
				enrichment = m.CVSS
				break
			}
		}
```

hdonnay: Unless there's some reason to not use a `for` loop, this should just use a `for` loop.

hdonnay (approved):
LGTM

PR #1602, opened 2025-08-15
vex: update module parsing

crozzy:
DO NOT MERGE: this is best guess attempt at how modules are represented in VEX data, but it just an attempt made by looking at the current data and not based on any type of guidelines.


crozzy:
closing in favor of https://github.com/quay/claircore/pull/1629

PR #1611, opened 2025-08-28
test: fix rhel updater check to deal with DeltaParse

crozzy:
The VEX updater exposes a DeltaParse() method that does all the work but the test just called Parse() for all updaters. This change checks if the DeltaParse() method is implemented and uses it, otherwise it uses Parse().

### DIFF
--- a/.github/workflows/updater-check.yml
+++ b/.github/workflows/updater-check.yml
@@ -22,7 +22,7 @@ jobs:
           check-latest: true
           go-version-file: ./go.mod
       - name: Check
-        run: go test ./test/periodic -enable
+        run: go test -timeout=20m ./test/periodic -enable
       - name: Create URL
         id: url
         if: failure()


.github/workflows/updater-check.yml
```diff
@@ -16,4 +16,4 @@ jobs:
         with:
           check-latest: true
           go-version-file: ./go.mod
-      - run: go test ./test/periodic -enable
+      - run: go test -timeout=20m ./test/periodic -enable
```
hdonnay: Is this really needed? It takes over 10 minutes?

hdonnay (approved):
SGTM

.github/workflows/updater-check.yml
crozzy: https://github.com/quay/claircore/actions/runs/17303309277/job/49119753908 yeah :(

hdonnay (approved):

PR #1620, opened 2025-09-02
all: modernize

hdonnay:
Run `golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize` across the project.

### DIFF
--- a/updater/osv/osv_test.go
+++ b/updater/osv/osv_test.go
@@ -33,7 +33,7 @@ func TestFetch(t *testing.T) {
 	ctx := zlog.Test(context.Background(), t)
 
 	f := Factory{}
-	cfgFunc := func(v interface{}) error {
+	cfgFunc := func(v any) error {
 		cfg := v.(*FactoryConfig)
 		cfg.URL = srv.URL
 		return nil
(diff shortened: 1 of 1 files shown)


hdonnay:
done via
```
go run github.com/quay/clair-workflows/cmd/modernize@latest
```

repository.go
```diff
@@ -8,5 +8,5 @@ type Repository struct {
 	Name string  `json:"name,omitempty"`
 	Key  string  `json:"key,omitempty"`
 	URI  string  `json:"uri,omitempty"`
-	CPE  cpe.WFN `json:"cpe,omitempty"`
```
BradLugo, line 11: Saw this a couple of times, so I think it's intended, but it's the only thing that caught my eye as a potential issue

hdonnay: It is; I read what that pass was doing and it made sense at the time. I'll go double-check and add a PR comment.

package.go
hdonnay, line 1: These `omitempty` tags never actually did anything because the tag only causes the field to be omitted when the struct value is zero-sized. This might be replaced with `omitzero`? I'm inclined to not bother, because out current thinking is that we don't actually want the JSON representation of our types to be used directly.

repository.go
hdonnay: See [this comment](https://github.com/quay/claircore/pull/1620/files#r2341948989).

PR #1629, opened 2025-09-10
Update RH VEX parsing to account for new module logic

crozzy:
The Red Hat security data team are updating how modules are represented
in VEX files, this change accounts for that update. Specifically, module
relationships are no longer formally expressed through VEX
relationships but rather as PURL qualifiers.


rhel/vex/testdata/cve-2024-24786-1.json
crozzy, line 1: This is purportedly the new format for modules

rhel/vex/parser_test.go
```diff
@@ -23,6 +23,35 @@ import (
...
+func TestComponentPURLToModuleName(t *testing.T) {
+	testcases := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{name: "old version", in: "pkg:rpmmod/redhat/postgresql@13:8060020240903094008:ad008a3a", want: ""},
+		{name: "simple", in: "pkg:rpm/redhat/postgresql-upgrade-devel-debuginfo@15.8-1.module%2Bel9.4.0%2B22208%2Bcd6b92ff?arch=s390x&rpmmod=postgresql:15:9040020240812115436:rhel9", want: "postgresql:15"},
+		{name: "short version", in: "pkg:rpm/redhat/golang-race?rpmmod=go-toolset:rhel8", want: "go-toolset:rhel8"},
+		{name: "empty rpmmod", in: "pkg:rpm/redhat/golang-race?rpmmod=", want: ""},
```
hdonnay: Maybe this should error? I think we should probably err on the side of being as well-defined as possible.

crozzy: Hmm, I'd agree but I think `packageurl.FromString` actually strips any empty qualifiers so in the test and in the actual code the scenario is never hit. I could manually define the purl in the tests, but that seems less representative of the code, what do you think?

crozzy: From the purl spec:
> A value MUST NOT be an empty string: a key=value pair with an empty value is the same as if no key=value pair exists for this key.

So seems like `packageurl.FromString` is doing the correct thing

crozzy: Well updated the tests (and code) any way to trip an error when the `rpmmod` is empty

hdonnay (approved):
LGTM

PR #1630, opened 2025-09-10
cicd: apply a bunch of lints to workflow files

hdonnay:
This is primarily to add permissions objects everywhere that GitHub and zizmor are complaining about. It also recommended not triggering the "Updater Check Failure" workflow the way it was written, so this way should do equivalently.

We may need to keep an eye on the CI for a week or so to see if the permissions are too tight.


.github/workflows/updater-check.yml
```diff
@@ -10,10 +10,45 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh run view "${GITHUB_RUN_ID}" --json jobs --jq '
+            .jobs[] | select(.name == "Periodic") |
+            (
+              "LOGS_URL=" +
+              .url +
+              (.steps[] | select(.name == "Check") | "#step:\(.number):0")
+            )' >> "$GITHUB_OUTPUT"
```
hdonnay, line 38: This is actually kind of neat, eg:
```
% gh run view "17619280136" --json jobs --jq '
  .jobs[] | select(.name == "Tests (stable)") |
  (
    .url +
    (.steps[] | select(.name == "Tests") |"#step:\(.number):0")
  )'
https://github.com/quay/claircore/actions/runs/17619280136/job/50060578250#step:6:0
```

PR #1631, opened 2025-09-11
updates: fix panic in test logs

hdonnay:
I'm taking this opportunity to restructure the tests to make them (IMO) easier to follow.

This seems to only trigger a race on go1.25.1, but prints after tests are done are against the `testing.T` contract.

See also: https://github.com/quay/claircore/actions/runs/17627587190/job/50088118987#step:6:65

### DIFF
--- a/libvuln/updates/locks_test.go
+++ b/libvuln/updates/locks_test.go
@@ -8,103 +8,126 @@ import (
 	"testing"
 )
 
-func TestLocalLockUnlock(t *testing.T) {
-	ctx := context.Background()
-	l := NewLocalLockSource()
-
-	t.Log("lock")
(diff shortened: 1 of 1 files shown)


libvuln/updates/locks_test.go
```diff
@@ -8,103 +8,126 @@ import (
...
+		}()
+
+		// Attempt to lock after the first goroutine releases it.
+		go func() {
+			defer wg.Done()
+			<-doStart
+
+			<-released
+			lc, done := locks.TryLock(ctx, key)
+			t.Logf("try: %v", lc.Err())
```
crozzy, line 83: Nit but might be better if this logged `try error:` or `error trying to lock:`

PR #1640, opened 2025-09-16
rpm: prefix PackageDB with kind

RTann:
Prior to https://github.com/quay/claircore/pull/869, the PackageDB for RPMs were prefixed with the DB kind (sqlite, bdb, or ndb). With that change, the prefix was removed. I believe this was an accident. This PR adds it back


crozzy (approved):
LGTM

PR #1641, opened 2025-09-16
rpm: return error upon failure

RTann:
When testing Claircore v1.5.40, I found the RPM packagescanner does not return an error if the layer does not exist. That's because the error is currently masked. It looks like the idea was to append this error to the output in the `defer`, but this requires us to use named return variables, which we do not at this time. This PR adds named return variables so the error is propagated correctly


rpm/packagescanner_test.go
```diff
@@ -201,3 +201,23 @@ func TestLayer(t *testing.T) {
...
+
+	l := &claircore.Layer{}
+	err := l.Init(t.Context(), desc, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var s Scanner
+	_, err = s.Scan(t.Context(), l)
+	if err == nil {
```
RTann, line 227: Prior to this change, the `err` was `nil`

rpm/packagescanner.go
```diff
@@ -50,7 +50,7 @@ func (*Scanner) Kind() string { return pkgKind }
 //
 // Deprecated: In-tree [indexer.PackageScanner] implementations should almost
 // certainly use the "internal/rpm" and "internal/dnf" packages.
-func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircore.Package, error) {
+func (ps *Scanner) Scan(ctx context.Context, layer *claircore.Layer) (out []*claircore.Package, err error) {
```
crozzy, line 53: Should the `err` inside the loop be changed so it's not shadowing the named return?

RTann: for readability? Sure I can do that

crozzy: Not just for readability but because it is overriding the value of `err` and I don't _think_ that is what is wanted in this case (because that error gets checked and returned immediately after. Or I'm reading this wrong

crozzy (approved):

rpm/packagescanner.go
RTann: I think it depends on which for loop you are talking about. I changed the only instances of shadowing I found. Are you referring to:

```
for db := range dbs {
		err = func() error {
			...
		}
}
```

If so, this is using `=` and not `:=`, so the `err` variable is set, as desired

RTann: Screenshot 2025-09-18 at 9 30 11 AM

Not sure what dev environment you are using, but GoLand will change the color of shadowed variables. This screenshot is the original (ie without me changing them to `dbErr` and `pkgErr`. These instances of `err` were ok, as they are returned by the `func() err`. You can also see the `err` in `err = func() err` is not highlighted, which tells us it is the same `err` as before (not shadowed)

PR #1642, opened 2025-09-17
rpm: fix use of `unique.Handle` pinning `fs.FS`

hdonnay:
This attempt to make the API impossible to misuse had some horrific, unintended side-effects for memory consumption, so remove it.


crozzy (approved):
LGTM!

PR #1644, opened 2025-09-19
dnf: add test and fixes for `dnf5`

hdonnay:


internal/dnf/sql.go
```diff
@@ -2,21 +2,63 @@ package dnf
...
+
+// Queries holds the sql needed to examine the dnf database.
+type queries struct {
+	AllRepoids       string
+	RepoidForPackage string
+}
+
+// SqlForVersion associates a version string returned from executing
+// [getDBVersion] to the SQL queries for that version of the database schema.
+var sqlForVersion = map[string]queries{
```
crozzy: Annoyingly I think there are some cases where it's dnf4 and 1.1:

- DNF4: https://github.com/rpm-software-management/libdnf/blob/master/libdnf/transaction/sql/create_tables.sql#L108
- DNF5: https://github.com/rpm-software-management/dnf5/blob/main/libdnf5/transaction/db/sql/create_tables.sql#L157

Having said that I have no idea where 1.2 comes from :thinking:

hdonnay: Yeah, I updated this with a check based on what tables are present.

crozzy (approved):
Nice, LGTM

PR #1654, opened 2025-10-01
ctxlock: remove v1 code

crozzy:
When the code was changed to use pgxv5 the ctxlock package grew a v2 to allow clients to be transferred without affecting the previous exported API. This change deprecates the old v1 functionality making the exported API return deprecation errors.


pkg/ctxlock/ctxlock.go
```diff
@@ -10,172 +10,25 @@ package ctxlock
...
-	}()
-	select {
-	case <-ready:
-	case <-ctx.Done():
-		l.Close(ctx)
-		return nil, ctx.Err()
-	}
-	return l, nil
+// Deprecated: Use ctxlock/v2 instead.
+func New(ctx context.Context, p *pgxpool.Pool) (*Locker, error) {
```
hdonnay: Could we change the argument type to `any`? I think it's _technically_ a breaking change, but isn't a source-level change.

crozzy: oh yeah good call, then the dep can actually go away

crozzy: Actually going to remove the comments too, they don't really make sense if the funcs don't do anything

hdonnay: The `Deprecated:` comments get special treatment in docs, I'd keep them.

crozzy: But like, the comments that describe what the function is doing?

hdonnay: I'd just leave the comments on the exported types and add the `Deprecated: ...` line.

go.mod
```diff
@@ -51,10 +50,9 @@ require (
 	github.com/jackc/pgpassfile v1.0.0 // indirect
 	github.com/jackc/pgproto3/v2 v2.3.3 // indirect
 	github.com/jackc/pgservicefile v0.0.0-20240606120523-5a60cdf6a761 // indirect
-	github.com/jackc/pgtype v1.14.2 // indirect
-	github.com/jackc/puddle v1.3.0 // indirect
 	github.com/jackc/puddle/v2 v2.2.2 // indirect
 	github.com/kylelemons/godebug v1.1.0 // indirect
+	github.com/lib/pq v1.10.2 // indirect
```
hdonnay, line 55: ???

hdonnay:
```
% go mod why github.com/lib/pq
# github.com/lib/pq
github.com/quay/claircore/datastore/postgres
github.com/doug-martin/goqu/v8
github.com/doug-martin/goqu/v8.test
github.com/lib/pq
```

PR #1657, opened 2025-10-01
all: slog conversion

hdonnay:
Does what it says on the tin.

~This is probably blocked on requiring go1.25 for the `(testing.TB).Output` method.~
The log output is done via buffering in-memory on go1.24.


alpine/distributionscanner.go
```diff
@@ -164,9 +161,9 @@ func readIssue(ctx context.Context, sys fs.FS) (*claircore.Distribution, error)
 			PrettyName: fmt.Sprintf(`Alpine Linux v%s`, v),
 		}, nil
 	case errors.Is(err, fs.ErrNotExist):
-		zlog.Debug(ctx).
-			Str("path", issuePath).
-			Msg("file doesn't exist")
+		slog.DebugContext(ctx,
+			"file doesn't exist",
+			"path", osrelease.Path)
```
crozzy:
```suggestion
			"path", issuePath)
```

crozzy (changes requested):
This all LGTM, just one minor, it really feels like Van Halen's brown M&M

PR #1660, opened 2025-10-08
update/osv: remove custom `cvss` implementation

hdonnay:
I don't _think_ this will affect the mapping, but did not check this.


updater/osv/osv.go
```diff
@@ -514,13 +515,23 @@ func (e *ecs) Insert(ctx context.Context, skipped *stats, name string, a *adviso
...
-			proto.NormalizedSeverity, err = fromCVSS3(ctx, s.Score)
+			var v cvss.V3
+			v, err = cvss.ParseV3(s.Score)
+			q = cvss.QualitativeScore(&v)
 		case `CVSS_V2`:
 			proto.Severity = s.Score
-			proto.NormalizedSeverity, err = fromCVSS2(s.Score)
+			var v cvss.V2
+			v, err = cvss.ParseV2(s.Score)
+			q = cvss.QualitativeScore(&v)
```
hdonnay, line 534: Could make this special-special like the previous code?

hdonnay: To expand on that, the code being removed did something ad-hoc for v2, whereas now it uses the v3/v4 "qualitative" spec.

Issue #1659, opened 2025-10-06
The `MediaType: "application/vnd.claircore.filesystem"` layer type can't handle inaccessible files and directories


evgenyz:
See log here: https://github.com/ComplianceAsCode/cvetool/issues/9. The same goes for special 'files' that are not accessible for reading even for root.

Since the layer is designed to read real file systems, it should be able work around these problems. What are the options here?

hdonnay:
Yeah, it's unexpected to attempt to read files that the process does not have permissions for.

I think wrapping the `fs.FS` returned from `os.Dir` with something that prevents walks from handling files where there's insufficient permissions.

evgenyz:
Makes sense. On the other hand, RPM scanner being unable to read `/boot/efi` is one thing that could easily be ignored, but  e.g. inaccessible `/var/lib/rpm/rpmdb.sqlite` is a problem that at least deserves a warning.

PR #1665, opened 2025-10-15
Gracefully handle access problems in layer FS

evgenyz:
An idea for fixing https://github.com/quay/claircore/issues/1659.

@hdonnay @BradLugo WDYT?

### DIFF
--- a/internal/filterfs/fs.go
+++ b/internal/filterfs/fs.go
@@ -1,0 +1,74 @@
+package filterfs
+
+import (
+	"io/fs"
+	"path"
+)
+
+// FS wraps an fs.FS and hides inaccessible files and directories
+// by returning appropriate "not found" errors or filtering them from listings.
+type FS struct {
+	fsys fs.FS
+}
+
+// New creates a new FS wrapper around the provided filesystem.
+func New(fsys fs.FS) *FS {
+	return &FS{fsys: fsys}
+}
+
+// Open opens the named file for reading and if file is a directory
+// wraps it with a ReadFileDir instance that filters out access errors
+func (f *FS) Open(name string) (fs.File, error) {
+	file, err := f.fsys.Open(name)
+	if err != nil {
+		return file, err
+	}
+
+	fi, err := file.Stat()
+	if err != nil {
+		return file, err
+	}
+
+	if fi.IsDir() {
+		return &DirFile{
+			fsys:    f,
+			fdir:    file,
+			pos:     0,
+			name:    name,
+			entries: nil,
+		}, nil
+	}
+
+	return file, nil
+}
+
+// ReadDir reads and returns directory entries, filtering out inaccessible items.
+func (f *FS) ReadDir(name string) ([]fs.DirEntry, error) {
+	if _, err := fs.Stat(f.fsys, name); err != nil {
+		return nil, fs.SkipDir
+	}
+
+	entries, err := fs.ReadDir(f.fsys, name)
+	if err != nil {
+		return nil, fs.SkipDir
+	}
+
+	var filtered []fs.DirEntry
+	// Filter out entries that are inaccessible
+	for _, entry := range entries {
+		// Try to stat the entry to check if it's accessible
+		path := path.Join(name, entry.Name())
+		fi, err := fs.Stat(f.fsys, path)
+		if err == nil {
+			// Accessible, include it
+			if fi.Mode().IsRegular() || fi.Mode().IsDir() {
+				if file, err := f.fsys.Open(path); err == nil {
+					file.Close()
+					filtered = append(filtered, entry)
+				}
+			}
+		}
+	}
+
+	return filtered, nil
+}
--- a/layer.go
+++ b/layer.go
@@ -2,15 +2,18 @@ package claircore
 
 import (
 	"bytes"
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
 	"io"
 	"io/fs"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
 
+	"github.com/quay/claircore/internal/filterfs"
 	"github.com/quay/claircore/pkg/tarfs"
 )
 
@@ -116,7 +119,14 @@ func (l *Layer) Init(ctx context.Context, desc *LayerDescription, r io.ReaderAt)
 		if desc.URI == "" {
 			return fmt.Errorf("claircore: layer %v: unable to create fs.FS: no URI provided", desc.Digest)
 		}
-		sys := os.DirFS(desc.URI)
+		u, err := url.Parse(desc.URI)
+		if err != nil {
+			return fmt.Errorf("claircore: layer %v: unable to create fs.FS: %w", desc.Digest, err)
+		}
+		if u.Scheme != "file" {
+			return fmt.Errorf("claircore: layer %v: unable to create fs.FS: unsupported URI scheme %v", desc.Digest, u.Scheme)
+		}
+		sys := filterfs.New(os.DirFS(cmp.Or(u.EscapedPath(), u.Opaque)))
 		l.sys = sys
 		l.rd = nil // The reader cannot be used for a filesystem
 	default:
--- a/internal/filterfs/file.go
+++ b/internal/filterfs/file.go
@@ -1,0 +1,49 @@
+package filterfs
+
+import (
+	"io"
+	"io/fs"
+)
+
+// A ReadDirFile wrapper for fs.File
+type DirFile struct {
+	fdir    fs.File
+	fsys    *FS
+	name    string
+	entries []fs.DirEntry
+	pos     int
+}
+
+func (d *DirFile) Close() error               { d.entries = nil; return d.fdir.Close() }
+func (d *DirFile) Read(_ []byte) (int, error) { return 0, io.EOF }
+func (d *DirFile) Stat() (fs.FileInfo, error) { return d.fdir.Stat() }
+
+func (d *DirFile) ReadDir(n int) ([]fs.DirEntry, error) {
+	if d.entries == nil {
+		es, err := d.fsys.ReadDir(d.name)
+		if err != nil {
+			return es, err
+		}
+		d.entries = es
+	}
+	es := d.entries[d.pos:]
+	if len(es) == 0 {
+		if n == -1 {
+			return nil, nil
+		}
+		return nil, io.EOF
+	}
+	end := min(len(es), n)
+	if n == -1 {
+		end = len(es)
+	}
+	d.pos += end
+	return es[:end], nil
+}
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
--- a/internal/filterfs/fs_test.go
+++ b/internal/filterfs/fs_test.go
@@ -1,0 +1,32 @@
+package filterfs
+
+import (
+	"os"
+	"testing"
+	"testing/fstest"
+)
+
+func TestCurrentDir(t *testing.T) {
+	fileset := []string{
+		"fs.go",
+		"fs_test.go",
+	}
+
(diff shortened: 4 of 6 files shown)


evgenyz:
I'd be then rather difficult to explain to the user why the scanner is not working correctly in a structured way on the application side.

evgenyz:
Okay, here's my another stab on it.

layer_fs_wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+// FilterFS wraps an fs.FS and hides inaccessible files and directories
+// by returning appropriate "not found" errors or filtering them from listings.
+type FilterFS struct {
+	fsys fs.FS
+	ctx  context.Context
+}
+
+// New creates a new FilterFS wrapper around the provided filesystem.
```
BradLugo: nit:
```suggestion
// NewFilterFS creates a new FilterFS wrapper around the provided filesystem.
```

layer_fs_wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+	"syscall"
+
+	"github.com/quay/zlog"
+)
+
+// FilterFS wraps an fs.FS and hides inaccessible files and directories
+// by returning appropriate "not found" errors or filtering them from listings.
+type FilterFS struct {
+	fsys fs.FS
+	ctx  context.Context
```
BradLugo: Looks like this Context is only used for logging. I wonder if there's another way to handle this? Passing it around in the struct doesn't feel great to me, but it definitely makes sense.

layer_fs_wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// New creates a new FilterFS wrapper around the provided filesystem.
+func NewFilterFS(fsys fs.FS, ctx context.Context) *FilterFS {
```
BradLugo: If we keep the Context in the struct, should it come first in this function? It _usually_ comes first, but perhaps having it come second would signal to the caller that we're using it a bit differently here.

BradLugo (commented):
Took a first pass at the review. The core logic looks good and is what we were asking for. Left some notes on the cosmetics. @hdonnay, what'd you think?

layer_fs_wrapper.go
evgenyz: Noted, will fix.

layer_fs_wrapper.go
evgenyz: I have no idea, I'm coming from a non-Go world. Whatever you can suggest fits me.

layer_fs_wrapper.go
evgenyz: I've swapped the parameters.

layer_fs_wrapper.go
evgenyz: Fixed.

internal/rpmver/rpmver.go
```diff
@@ -128,6 +128,7 @@ func Parse(v string) (Version, error) {
 // and what's just another version segment.
 var architectures = map[string]struct{}{
 	".aarch64": {},
+	".i386":    {},
```
hdonnay: This seems off-topic, and possibly unwanted? I don't think there are any non-EOL distributions that build for `i386`.

internal/dnf/dnf.go
```diff
@@ -264,6 +264,14 @@ func (h *historyDB) AddRepoid(ctx context.Context, pkg *claircore.Package) error
...
 	}
 
+	if ver.Architecture == nil {
+		zlog.Warn(ctx).
+			Err(err).
+			Str("version", v).
+			Msg("unable to parse arch")
+		return nil
+	}
+
```
hdonnay, line 273: This shouldn't log IMO, and definitely shouldn't log above "debug". Architecture missing is not something that's _always_ unexpected or requires notifying the operator.

layer_fs_wrapper.go
hdonnay:
```suggestion
```

Go idiom is not to capture `context.Context` values.

layer_fs_wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Open opens the named file for reading. If the file is inaccessible,
+// it returns fs.ErrNotExist to hide its existence.
+func (f *FilterFS) Open(name string) (fs.File, error) {
+	file, err := f.fsys.Open(name)
+	if err != nil {
+		// Convert permission errors to not found to hide existence
+		if os.IsPermission(err) {
+			zlog.Debug(f.ctx).Str("path", name).Err(err).Msg("access error on Open")
```
hdonnay: Don't log and return. <sup>[doc](https://quay.github.io/claircore/contributor/logging.html#dont-log-and-return)</sup>

If you want to keep the original error, you could return something like:
```
fmt.Errorf("filterfs: path %q: %w", name, errors.Join(err, fs.ErrNotExist))
```

layer_fs_wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+	return file, nil
+}
+
+// ReadDir reads and returns directory entries, filtering out inaccessible items.
+func (f *FilterFS) ReadDir(name string) ([]fs.DirEntry, error) {
+	entries, err := fs.ReadDir(f.fsys, name)
+	if err != nil {
+		if os.IsPermission(err) || errors.Is(err, syscall.EINVAL) {
+			zlog.Debug(f.ctx).Str("path", name).Err(err).Msg("access error on ReadDir")
+			return nil, fs.SkipDir
```
hdonnay: Don't log and return. <sup>[doc](https://quay.github.io/claircore/contributor/logging.html#dont-log-and-return)</sup>

If you want to keep the original error, you could return something like:
```
fmt.Errorf("filterfs: path %q: %w", name, errors.Join(err, fs.SkipDir))
```

layer_fs_wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		return nil, err
+	}
+
+	// Filter out entries that are inaccessible
+	var filtered []fs.DirEntry
+	for _, entry := range entries {
+		// Try to stat the entry to check if it's accessible
+		path := filepath.Join(name, entry.Name())
```
hdonnay: This sort of path manipulation should use `path` instead of `filepath` because the `fs.FS` interface defines all paths to be [slash-separated](https://pkg.go.dev/io/fs#ValidPath). Using `filepath` means the behavior will vary by platform.

layer_fs_wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// WalkDir walks the directory tree rooted at dir, skipping inaccessible entries.
+func (f *FilterFS) WalkDir(name string, fn fs.WalkDirFunc) error {
+	return fs.WalkDir(f.fsys, name, func(path string, d fs.DirEntry, err error) error {
+		// Hide inaccessible entries by skipping them silently
+		if err != nil {
+			if os.IsPermission(err) || errors.Is(err, syscall.EINVAL) {
+				zlog.Debug(f.ctx).Str("path", path).Err(err).Msg("access error on WalkDir")
+				return fs.SkipDir
```
hdonnay: Don't log and return. <sup>[doc](https://quay.github.io/claircore/contributor/logging.html#dont-log-and-return)</sup>

If you want to keep the original error, you could return something like:
```
fmt.Errorf("filterfs: path %q: %w", path, errors.Join(err, fs.SkipDir))
```

hdonnay (changes requested):
1. I think the changes to the `dnf` and `rpmver` packages should be dropped or split into separate PRs.
1. The `context.Context` should be removed from the `FilterFS` struct. A consequence of that is that the logging has to be removed, but that's just the way it goes.
1. This `FilterFS` type should either be unexported (renamed to `filterFS`) or moved to a separate package (so renamed+moved to something like `./internal/filterfs.FS`).
1. I'd like to see a simple test using `testing/fstest.TestFS` to wrap the `FilterFS` around a `testing/fstest.MapFS`.

I think that implementing the `fs.StatFS` interface would get `fs.WalkDir` to work "for free", but implementing `fs.WalkDirFS` is fine.

internal/rpmver/rpmver.go
evgenyz: You'd be surprised what one could find on a real workstation system: https://www.brother.co.uk/en/support/mfc9140cdn/downloads.

layer_fs_wrapper.go
evgenyz: Fixed.

layer_fs_wrapper.go
evgenyz: Fixed.

layer_fs_wrapper.go
evgenyz: Fixed.

layer_fs_wrapper.go
evgenyz: Fixed.

layer_fs_wrapper.go
evgenyz: Done.

internal/dnf/dnf.go
evgenyz: Kept as `Debug` level in a separate PR: https://github.com/quay/claircore/pull/1674.

internal/rpmver/rpmver.go
evgenyz: The `i386` arch is now gone.

internal/filterfs/fs.go
```diff
@@ -0,0 +1,68 @@
...
+	"errors"
+	"io/fs"
+	"os"
+	"path"
+	"syscall"
+)
+
+// filterfs.FS wraps an fs.FS and hides inaccessible files and directories
+// by returning appropriate "not found" errors or filtering them from listings.
+type FS struct {
```
hdonnay:
```suggestion
// FS wraps an fs.FS and hides inaccessible files and directories
// by returning appropriate "not found" errors or filtering them from listings.
type FS struct {
```

internal/filterfs/fs.go
```diff
@@ -0,0 +1,68 @@
...
+	fsys fs.FS
+}
+
+// FilterFS creates a new filterfs.FS wrapper around the provided filesystem.
+func FilterFS(fsys fs.FS) *FS {
```
hdonnay:
```suggestion
// New creates a new filterfs.FS wrapper around the provided filesystem.
func New(fsys fs.FS) *FS {
```

internal/filterfs/fs.go
```diff
@@ -0,0 +1,68 @@
...
+	return &FS{fsys: fsys}
+}
+
+// Stat just funnels the request to fs.Stat, no need to filter errors here
+func (f *FS) Stat(name string) (fs.FileInfo, error) {
+	return fs.Stat(f.fsys, name)
+}
```
hdonnay: If this doesn't do anything, then it can just be omitted. The `fs.Stat` function will call `Open` and `Stat` on the returned `File`.

evgenyz: You're right. Got rid of it.

layer.go
```diff
@@ -116,7 +117,7 @@ func (l *Layer) Init(ctx context.Context, desc *LayerDescription, r io.ReaderAt)
 		if desc.URI == "" {
 			return fmt.Errorf("claircore: layer %v: unable to create fs.FS: no URI provided", desc.Digest)
 		}
-		sys := os.DirFS(desc.URI)
+		sys := filterfs.New(os.DirFS(desc.URI))
```
hdonnay:
```suggestion
		u, err := url.Parse(desc.URI)
		if err != nil {
		// ...
		}
		if u.Scheme != "file" {
		// return error
		}
		sys := filterfs.New(os.DirFS(cmp.Or(u.EscapedPath(), u.Opaque)))
```

evgenyz: Done.

PR #1666, opened 2025-10-15
rhel: account for multiple cpes per repoid

crozzy:
This PR has two distinct concerns:
* Update the RHEL repositoryScanner to utilize the URI field for saving the repoid metadata
* Use that information in the RHEL coalescer to make decisions about which repositories are associated to which packages.

Here is a brief logic explanation:
**No DNF repoid:**
* If the content_sets have the use_dnf key, don’t associate the package to any repo
* If the content_sets don’t have the use_dnf key, associate the package with the content_sets’ repos (i.e every CPE that each content_set repoid refers to in the repository_to_cpe map).

**DNF repoid exists:**
* If the repoid exists in the repo_to_cpe map then associate package to repos for all the mapped CPEs.
* If the repoid doesn’t exist in the repo_to_cpe map then don’t associate the package to any repo

_Note_: it isn’t possible to currently save multiple repos for a repoid (even if the coalescer was updated) because of the unique index on the repo table: 
```sql
CREATE UNIQUE INDEX repo_unique_idx ON public.repo USING btree (name, key, uri)
```
means that only one repo was ever persisted to the DB for each repoid.

### DIFF
--- a/rhel/coalescer.go
+++ b/rhel/coalescer.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/indexer"
+	"github.com/quay/zlog"
 )
 
 // Coalescer takes individual layer artifacts and coalesces them into a full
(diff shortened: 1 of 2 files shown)


rhel/coalescer.go
```diff
@@ -63,12 +64,9 @@ func (*Coalescer) Coalesce(ctx context.Context, artifacts []*indexer.LayerArtifa
 	}
 	// This dance with copying the product information in both directions means
 	// that if Red Hat product information is found, it "taints" all the layers.
-
-	// Break the key-by-ID convention because we need to talk about the repoid
-	// in the package matching step.
 	for _, a := range artifacts {
 		for _, repo := range a.Repos {
-			ir.Repositories[repo.Name] = repo
+			ir.Repositories[repo.ID] = repo
```
hdonnay, line 69: Why change this back to using the `ID`?

crozzy: It's a field that uniquely identifies the repository.

I think if it's `Name` and it holds the string representation of a CPE then that seems like it'd be unique, but that CPE might have come from 2 different repoids and when trying to match a certain package's repoid it might not be the one we've persisted. I guess there is a 3rd option that is make the repository scanner append more repoids to the `repoid=` if we see it multiple times for different repoids.

PR #1670, opened 2025-10-23
datastore/postgres: lock around manifest delete

hdonnay:
Should be straightforward. I had to pull the key making function into a common spot, but that's the only reposurgery.

### DIFF
--- a/datastore/postgres/deletemanifests.go
+++ b/datastore/postgres/deletemanifests.go
@@ -11,6 +11,7 @@ import (
 	"github.com/quay/zlog"
 
 	"github.com/quay/claircore"
+	"github.com/quay/claircore/internal/dblock"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


hdonnay:
LMK if you think this needs a test specifically.

datastore/postgres/deletemanifests.go
```diff
@@ -53,23 +61,40 @@ func (s *IndexerStore) DeleteManifests(ctx context.Context, ds ...claircore.Dige
 			l.id = $1
 			AND ml.layer_id IS NULL
 		);`
+		// This is equivalent to the ctxlock command except that it's requesting
+		// a transaction-scoped lock.
+		tryLock = `SELECT lock FROM pg_try_advisory_xact_lock($1) lock WHERE lock = true;`
```
crozzy, line 66: I think this makes sense, it's a silent error if more than one manifest is being deleted but it's really the only "good" outcome when being told to do two contradictory things.

crozzy: The indexing will queue behind any deletions which should be fine (unless something gets out of whack with the indexes the deletions use and it causes a slowdown in indexing, again, very corner-casey but worth having in the back of our minds that the two process are now connected).

hdonnay: Right; partial progress isn't an error condition, but no progress is.

PR #1681, opened 2025-11-14
photon: add dynamic distribution discovery

crozzy:
The photon ecosystem still had statically defined distributions which meant that it has gotten stale over time. This change add the ability to dynamically discover those distributions.
<a data-ca-tag href="https://codeapprove.com/pr/quay/claircore/1681">Review on CodeApprove</a>

### DIFF
--- a/photon/distributionscanner.go
+++ b/photon/distributionscanner.go
@@ -12,9 +12,6 @@ import (
 	"github.com/quay/claircore/indexer"
 )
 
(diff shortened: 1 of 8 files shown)


crozzy:
This came up as I was doing the PURL stuff, I'm not sure if anyone uses this but it's simple enough to update

hdonnay (approved):

PR #1694, opened 2025-12-03
rpm: close sqlite resources on the error path

crozzy:
It was observed in production that there was a goroutine leak, this was traced back to a database/sql.(*DB).connectionOpener+0x86 call and the rpm sqlite support. There is a case where the DB is open and the Ping() fails. This means there is no Close() issued, no spool file clean up and no finalizer supported added, so it's not caught in a panic.
<a data-ca-tag href="https://codeapprove.com/pr/quay/claircore/1694">Review on CodeApprove</a>

### DIFF
--- a/internal/rpm/find.go
+++ b/internal/rpm/find.go
@@ -218,6 +218,12 @@ func OpenDB(ctx context.Context, sys fs.FS, found FoundDB) (*Database, error) {
 	case kindSQLite:
 		sdb, err := sqlite.Open(spool.Name())
 		if err != nil {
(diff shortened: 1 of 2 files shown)


hdonnay (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/quay/claircore/1694)


<a href="https://codeapprove.com/pr/quay/claircore/1694">Approved on CodeApprove</a>
✔️ Approved

🤦

<hr />

👀 @crozzy it's your turn please take a look

PR #1730, opened 2026-01-13
vex: don't error when fetching non-existent VEX files.

crozzy:
There is a case during a period when there are lots of deletes happening in the VEX data that there are files deleted after we fetch the changes.csv file and before we actually try and read the changes in the referenced VEX file.
<a data-ca-tag href="https://codeapprove.com/pr/quay/claircore/1730">Review on CodeApprove</a>

### DIFF
--- a/rhel/vex/fetcher.go
+++ b/rhel/vex/fetcher.go
@@ -337,9 +337,17 @@ func (u *Updater) processChanges(ctx context.Context, w io.Writer, fp *fingerpri
 				return fmt.Errorf("error making advisory request %w", err)
 			}
 			defer res.Body.Close()
-			err = httputil.CheckResponse(res, http.StatusOK)
-			if err != nil {
-				return fmt.Errorf("unexpected response: %w", err)
+
+			switch res.StatusCode {
+			case http.StatusOK:
+				break
+			case http.StatusNotFound:
+				// We don't want to fail the entire fetch if an advisory is not found.
+				// It can happen if the advisory has just been deleted.
+				slog.DebugContext(ctx, "advisory not found", "url", advisoryURI)
+				return nil
+			default:
+				return fmt.Errorf("unexpected response: %s", res.Status)
 			}
 
 			// Add compacted JSON to buffer.


rhel/vex/fetcher.go
```diff
@@ -337,9 +337,17 @@ func (u *Updater) processChanges(ctx context.Context, w io.Writer, fp *fingerpri
...
-			err = httputil.CheckResponse(res, http.StatusOK)
-			if err != nil {
-				return fmt.Errorf("unexpected response: %w", err)
+
+			switch res.StatusCode {
+			case http.StatusOK:
+				break
+			case http.StatusNotFound:
+				// We don't want to fail the entire fetch if an advisory is not found.
+				// It can happen if the advisory has just been deleted.
```
jvdm, line 346: "Just been deleted" seems to include planned deletions but also unexpected deletions (eg. security data errors), right? *Not failing* here takes care of both scenarios?

crozzy: It would account for both scenarios, the caveat being what Hank said: if it wasn't a legit delete that change wouldn't be represented in the DB (and even if the file is added back and is different from the original, those differences won't be represented in the DB until (or if) it's changed again).

jvdm: Yes. And for use cases where the data is being added "from scratch" (eg. air-gapped envs, or exports of vulnerabilities) this would not be a problem as added back entries would mean recovery, right?

crozzy: Right, in those cases we don't have to "keep state" we just have to represent current state.

jvdm: The problem is if the deletion is invalid (not "legit").

When consuming a feed, if part of the feed payload is invalid according to the spec, the consumer should ignore the invalid part, still ingest the rest, and still report the feed as "updated"?

hdonnay:
I guess I'm trying to guard against the case where the `csv` files are changed in the "incorrect" order w/r/t the deletion of the VEX file. As it stands, there's 3 logical events:
1. `changes.csv` is updated
2. `deletions.csv` is updated
3. VEX file is deleted

Because we process `changes` *and* fetch VEX files _then_ process `deletions`, we assume that these have to happen in the order above, but that's not specified anywhere. Using the csvs to generate one "view" of the state of the VEX files might make it more resilient?

I dunno, this gets very distsys very fast. It'd be much nicer if they atomically swapped between revisions of the VEX corpus.

---

Another, perhaps worse, approach would be to restart the (little-p) process when we detect an inconsistency.

crozzy:
> I guess I'm trying to guard against the case where the `csv` files are changed in the "incorrect" order w/r/t the deletion of the VEX file. As it stands, there's 3 logical events:
> 
>     1. `changes.csv` is updated
> 
>     2. `deletions.csv` is updated
> 
>     3. VEX file is deleted
> 
> 
> Because we process `changes` _and_ fetch VEX files _then_ process `deletions`, we assume that these have to happen in the order above, but that's not specified anywhere. Using the csvs to generate one "view" of the state of the VEX files might make it more resilient?
> 
> I dunno, this gets very distsys very fast. It'd be much nicer if they atomically swapped between revisions of the VEX corpus.
> 
> Another, perhaps worse, approach would be to restart the (little-p) process when we detect an inconsistency.

The problem is, even if prodsec did atomically swap all the data at once (which I suspect is not the case), we still have a race condition where we've requested the changes.csv file (and potentially the deletions.csv) before processing it, so there is always the potential that things have changed/disappeared in that delta. If we started the process again it might work but it also might not again if deletions fall into the deadly void.

It's like we'd want to transactional-ize something that is cannot logically fit in a transaction (the retrieval of what has changed and what those changes are).

PR #1733, opened 2026-01-14
rpm: check more paths for rpm-installed files

crozzy:
The function that checks if files were installed via RPMs gates the paths it considers with a regex, this regex was missing some binary paths that (when evaluated) don't contain a preceding slash. This also adds a path where the golang binaries are installed in some RH images.
<a data-ca-tag href="https://codeapprove.com/pr/quay/claircore/1733">Review on CodeApprove</a>

### DIFF
--- a/internal/rpm/info.go
+++ b/internal/rpm/info.go
@@ -223,15 +223,16 @@ var (
 	// FilePatterns is a regular expression for *any* file that may need to be
 	// recorded alongside a package.
 	//
-	// The tested strings are absolute paths.
+	// The tested strings are paths relative to the rpm root (almost always "/").
 	filePatterns = sync.OnceValue(func() *regexp.Regexp {
 		pat := []string{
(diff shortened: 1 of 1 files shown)


hdonnay:
Oh, yeah, that's what the comment says, huh. Then yeah the regexp is just wrong :+1:

crozzy:
Sorry, should have updated before but it just occurred to me

internal/rpm/info.go
```diff
@@ -223,15 +223,16 @@ var (
 	// FilePatterns is a regular expression for *any* file that may need to be
 	// recorded alongside a package.
 	//
-	// The tested strings are absolute paths.
+	// The tested strings are absolute paths but don't contain leading slashes.
```
hdonnay: So, not absolute.
```suggestion
	// The tested strings are paths relative to the rpm root (almost always "/").
```

hdonnay (changes requested):

hdonnay (approved):

PR #1739, opened 2026-01-20
bdb: fix (add) in-line value support

hdonnay:
This adds support for cases where the rpm header blobs are small enough to be stored in-line instead of in dedicated "overflow" pages.

See-also: https://issues.redhat.com/browse/CLAIRDEV-229


internal/rpm/bdb/bdb.go
```diff
@@ -93,183 +96,318 @@ type meta struct {
...
+						}
+						var buf bytes.Buffer
+						buf.Grow(int(itemLen))
+						if _, err := io.CopyN(&buf, pg, itemLen); err != nil {
+							if !yield(nil, fmt.Errorf("bdb: error reading hash entry: %w", err)) {
+								return
+							}
+							continue HandleEntry
+						}
+						buf.ReadByte() // Skip over "type"
```
crozzy: Should we check for an error here?

internal/rpm/bdb/bdb.go
```diff
@@ -93,183 +96,318 @@ type meta struct {
...
+	return ov, nil
+}
+
+// Overflow returns a [rope] reading the data from one or more Overflow pages.
+func (db *PackageDB) overflow(start uint32) (*rope, error) {
+	var r rope
+	pgno := start
+	for pgno != 0 {
+		pg := db.page(pgno)
+		ov, err := db.readOverflowpage(pg)
```
crozzy, line 411: Should this validate that it's actually a `PageTypeOverflow      // P_OVERFLOW` (or the `readOverflowpage` check it before it's returned)? I'd assume the data was corrupted in some way if it wasn't so I guess; is it overkill to check?

crozzy (commented):
Nice! 2 nits

crozzy (approved):
LGTM

PR #1741, opened 2026-01-23
fetcher: give tmpfile ability to create

crozzy:
When forcing the linux fallback (no O_TMPFILE) it triggers an error:
encountered errors: openat fetcher.xxxxxxxx: no such file or directory,
when adding the O_CREATE flag the fallback's error was fixed. This also
adds the flags to the unix and windows variants.

### DIFF
--- a/libindex/tempfile_linux.go
+++ b/libindex/tempfile_linux.go
@@ -25,7 +25,7 @@ func canTmp(dir *os.Root) bool {
 	tmpMap.RUnlock()
 	// want the default inverted:
 	if !loaded {
-		ok = true
+		ok = tryTMPFILE
 	}
 	return ok
 }
--- a/libindex/fetcher_test.go
+++ b/libindex/fetcher_test.go
@@ -49,15 +49,26 @@ func (tc fetchTestcase) Run(ctx context.Context) func(*testing.T) {
 }
 
 func TestFetchSimple(t *testing.T) {
-	ctx := context.Background()
-	tt := []fetchTestcase{
-		{N: 1},
-		{N: 4},
-		{N: 32},
-	}
+	runTCs := func(t *testing.T) {
+		ctx := t.Context()
+		tt := []fetchTestcase{
+			{N: 1},
+			{N: 4},
+			{N: 32},
+		}
 
-	for _, tc := range tt {
-		t.Run(strconv.Itoa(tc.N), tc.Run(ctx))
+		for _, tc := range tt {
+			t.Run(strconv.Itoa(tc.N), tc.Run(ctx))
+		}
+	}
+	runTCs(t)
+
+	if runtime.GOOS == "linux" {
+		t.Run("NoTMPFILE", func(t *testing.T) {
+			tryTMPFILE = false
+			t.Cleanup(func() { tryTMPFILE = true })
+			runTCs(t)
+		})
 	}
 }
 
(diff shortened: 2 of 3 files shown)


libindex/tempfile_linux.go
```diff
@@ -46,6 +46,7 @@ Loop:
 		flag := os.O_WRONLY
 		if !tmpOK {
 			name = fetchFilename()
+			flag |= os.O_CREATE | os.O_EXCL // Create new file, fail if exists
```
crozzy, line 49: Before the change I was manually forcing the fallback here and could trigger the error

crozzy: I mounted a fat32 filesystem and used it as the tmpfile to test this.

libindex/tempfile_unix.go
```diff
@@ -20,9 +19,8 @@ func openTemp(dir *os.Root) (f *os.File, err error) {
...
 	}
-	if err != nil {
-		return nil, err
-	}
-	runtime.AddCleanup(f, func(n string) { dir.Remove(n) }, f.Name())
-	return f, nil
+	// Files are cleaned up when the arena is closed or the process exits.
+	// We cannot use runtime.AddCleanup here because reopen() needs to open
+	// the file by name, and the cleanup would race with that.
+	return f, err
```
hdonnay: I think this needs the `AddCleanup`, because otherwise the process will never remove the files and just fill disk. The file should remain reachable as needed because of the closure [here](https://github.com/quay/claircore/blob/93df178179b263f7acf65bb0f13fa302138c4ff8/libindex/fetcher.go#L129-L138).

crozzy: OK, that seems sound

hdonnay: Yeah, I don't know how to cross-link that sort of tricky interaction. You can only know this by reading the arena code first.

crozzy: I updated now

libindex/tempfile_linux.go
hdonnay, line 49: 🤦

libindex/tempfile_unix.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func openTemp(dir *os.Root) (f *os.File, err error) {
-	const flag = os.O_WRONLY
+	const flag = os.O_WRONLY | os.O_CREATE | os.O_EXCL
```
hdonnay, line 13: 🤦

libindex/tempfile_windows.go
```diff
@@ -12,7 +12,7 @@ import (
 func openTemp(dir *os.Root) (f *os.File, err error) {
 	// Copied out of golang.org/x/sys/windows
 	const FILE_FLAG_DELETE_ON_CLOSE = 0x04000000
-	const flag = os.O_WRONLY | FILE_FLAG_DELETE_ON_CLOSE
+	const flag = os.O_WRONLY | os.O_CREATE | os.O_EXCL | FILE_FLAG_DELETE_ON_CLOSE
```
hdonnay, line 15: 🤦

libindex/reopen_other.go
```diff
@@ -2,8 +2,11 @@
...
 
-import "os"
+import (
+	"os"
+	"path/filepath"
+)
 
 func reopen(dir *os.Root, f *os.File) (*os.File, error) {
-	return dir.OpenFile(f.Name(), os.O_RDONLY, 0o644)
+	return dir.OpenFile(filepath.Base(f.Name()), os.O_RDONLY, 0o644)
```
hdonnay, line 11: 🤦

hdonnay (approved):

PR #1746, opened 2026-01-28
sbom/spdx: add SPDX decoder

BradLugo:
Adds a Decoder interface to the sbom package and implements an SPDX JSON decoder that converts SPDX documents back to IndexReport format. There's also an encoder refactoring patch to make the decoder implementation a bit nicer.

Requires https://github.com/quay/claircore/pull/1744.


sbom/spdx/encoder.go
```diff
@@ -11,30 +11,18 @@ import (
 	"strings"
 	"time"
 
+	"github.com/quay/claircore/sbom"
```
hdonnay: ???

sbom/spdx/spdx.go
```diff
@@ -0,0 +1,13 @@
+package spdx
+
+// Format describes the data format for the SPDX document.
+type Format string
+
+const JSONFormat Format = "json"
```
hdonnay:
```suggestion
const FormatJSON Format = "json"
```

Doing it this way makes autocomplete easier and multiple formats group better in documentation output, assuming there's ever another `Format`.

sbom/sbom.go
```diff
@@ -11,3 +11,8 @@ import (
 type Encoder interface {
 	Encode(ctx context.Context, w io.Writer, ir *claircore.IndexReport) error
 }
+
+// Decoder is an interface to convert an encoded SBOM from r into a claircore.IndexReport
```
hdonnay:
```suggestion
// Decoder is an interface to convert an encoded SBOM into a [claircore.IndexReport].
```

Full sentence with proper markup. If you want the documentation to say something specific about a method, put the documentation on that method.

sbom/spdx/encoder.go
```diff
@@ -11,30 +11,18 @@ import (
...
 )
 
-// Format describes the data format for the SPDX document.
-type Format string
-
-const JSONFormat Format = "json"
-
-// Option is a type for setting optional fields for the Encoder.
-type Option func(*Encoder)
+// EncoderOption is a type for setting optional fields for the Encoder.
```
hdonnay:
```suggestion
// EncoderOption is a type for configuring an [Encoder].
```

sbom/spdx/decoder.go
```diff
@@ -0,0 +1,190 @@
...
+func WithDecoderPURLConverter(registry purl.Converter) DecoderOption {
+	return func(d *Decoder) {
+		d.PURLConverter = registry
+	}
+}
+
+// Decode decodes an SPDX document from r and returns an IndexReport.
+//
+// Known limitations:
+// - Only package indexing via PURL ExternalRefs is supported.
```
hdonnay:
```suggestion
//   - Only package indexing via PURL ExternalRefs is supported.
```
Needs two spaces of indent to render correctly.

sbom/spdx/decoder.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/package-url/packageurl-go"
+	"github.com/quay/claircore/sbom"
+	spdxjson "github.com/spdx/tools-golang/json"
+	"github.com/spdx/tools-golang/spdx/v2/v2_3"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/purl"
+)
+
+// DecoderOption is a type for setting optional fields for the Decoder.
```
hdonnay:
```suggestion
// DecoderOption is a type for configuring a [Decoder].
```

sbom/spdx/decoder.go
```diff
@@ -0,0 +1,190 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"strconv"
+
+	"github.com/package-url/packageurl-go"
+	"github.com/quay/claircore/sbom"
```
hdonnay: Wrong grouping

sbom/spdx/decoder.go
```diff
@@ -0,0 +1,190 @@
...
+				var unhandled purl.ErrUnhandledPurl
+				if errors.As(err, &unhandled) {
+					slog.WarnContext(ctx, "unregistered PURL type",
+						"purl", ref.Locator,
+						"type", unhandled.Type,
+						"namespace", unhandled.Namespace)
+				} else {
+					slog.WarnContext(ctx, "failed to parse PURL to IndexRecords",
+						"purl", ref.Locator,
+						"error", err)
```
hdonnay:
```suggestion
						"reason", err)
```
House style is to try to avoid "error" so that a grep for errors reveals fatal errors. This happens a few times in the log lines.

sbom/spdx/decoder.go
```diff
@@ -0,0 +1,190 @@
...
+	ir := &claircore.IndexReport{
+		State:         "IndexFinished",
+		Success:       true,
+		Packages:      make(map[string]*claircore.Package),
+		Distributions: make(map[string]*claircore.Distribution),
+		Repositories:  make(map[string]*claircore.Repository),
+		Environments:  make(map[string][]*claircore.Environment),
+	}
+
+	if d.PURLConverter == nil {
```
hdonnay, line 92: Why does this behave differently to the `Encoder`?

hdonnay (changes requested):
Some documentation fixes/nits. Want to see:

- [ ] fix import grouping
- [ ] use `"reason"` key for logging errors instead of `"error"`
- [ ] why is the default behavior different between `Encoder` and `Decoder` w/r/t the converter?

sbom/spdx/decoder.go
BradLugo: You could, in theory, convert a `claircore.IndexRecord` into an SPDX package without a PURL, but we wouldn't be able to read it back. I can add that requirement to the encoder if you'd like (it should _probably_ be a hard requirement at some point anyway), though it might be some ACS functionality.

hdonnay: Yeah, my preference would be to have a default converter the way the `Encoder` does.

BradLugo: I'm thinking of two options here:

1. We change the encoder to require a PURL.Converter
2. We don't skip+warn if no PURLConverter is set on the decoder, but we'll still skip during the parsing: https://github.com/BradLugo/claircore/blob/467fce6c520b15b1dd08d88493d31721afea3c44/sbom/spdx/decoder.go#L108-L111

Is your preference <span>#</span>2?

hdonnay: Okay, after reading _and comprehending_ the code, the Encoder and Decoder _do_ have the same behavior. So I think this is ok.

PR #1749, opened 2026-02-04
postgres: streamline migrations

hdonnay:
This should make it easier to run the migrations and make sure that no connections from before the migrations are run could possibly be around to be used by accident.

See-also: https://issues.redhat.com/browse/CLAIRDEV-232


datastore/postgres/types/version.go
```diff
@@ -37,6 +37,7 @@ func registerVersionRange(ctx context.Context, c *pgx.Conn) error {
 		if pgErr.Code == "42704" { // OK: "no such type"
 			break
 		}
+		c.Exec(ctx, `DO $$ BEGIN RAISE NOTICE USING MESSAGE = 'type error expected', DETAIL = 'running migrations should create needed types'; END; $$;`)
```
BradLugo: Should the exec be in the pg err code check?
```suggestion
		if pgErr.Code == "42704" { // OK: "no such type"
			c.Exec(ctx, `DO $$ BEGIN RAISE NOTICE USING MESSAGE = 'type error expected', DETAIL = 'running migrations should create needed types'; END; $$;`)
			break
		}
```

hdonnay: yes

BradLugo (approved):

PR #1767, opened 2026-03-02
release: allow for branch protection in release steps

crozzy:
Since the introduction of the branch protection rule to stop users pushing directly to `main` the previous release steps started failing. This changes the instructions to require the user merges the changelog update before tagging the release.

### DIFF
--- a/.github/scripts/prepare-release
+++ b/.github/scripts/prepare-release
@@ -38,6 +38,7 @@ gh run watch "${id}"
 cat <<.
 Now, to merge and release:
 
+	go to github and approve the PR titled "${VERSION} Changelog Bump" and wait for it to be merged.
 	git checkout ${BRANCH}
 	git fetch ${REMOTE}
 	git merge --ff-only ${REMOTE}/ready-${VERSION}


.github/scripts/prepare-release
```diff
@@ -38,10 +38,11 @@ gh run watch "${id}"
...
 Now, to merge and release:
 
+	go to github and approve the PR titled "${VERSION} Changelog Bump" and wait for it to be merged.
 	git checkout ${BRANCH}
 	git fetch ${REMOTE}
-	git merge --ff-only ${REMOTE}/ready-${VERSION}
+	git merge --ff-only ${REMOTE}/${BRANCH}
 	git tag -s ${VERSION}
-	git push ${REMOTE} ${BRANCH} :ready-${VERSION} tag ${VERSION}
+	git push :ready-${VERSION} tag ${VERSION}
```
hdonnay: Why are these changes needed?

crozzy: 44. you need to merge the changelog update that has just been merged upstream into your local `main` before cutting the ~branch~ tag (I think)
46. you can't `git push upstream main` anymore

hdonnay: I think the only change that needs to be made is "go approve the PR." You can still "merge" via the command line, because that's what our `fast-forward` command does under the hood.

crozzy: Oh, so the last command should work because `main` will not have to be modified?

hdonnay: Yeah, it should work because the commit that's added in that push is "approved" on the PR.

PR #1781, opened 2026-03-19
claircore: change `Package.Kind` type and add new values

hdonnay:
This was largely an automated rewrite, but does touch a lot of files.

### DIFF
--- a/datastore/postgres/querybuilder.go
+++ b/datastore/postgres/querybuilder.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/datastore"
+	"github.com/quay/claircore/internal/wart"
 	"github.com/quay/claircore/libvuln/driver"
 )
 
@@ -26,7 +27,7 @@ func buildGetQuery(record *claircore.IndexRecord, opts *datastore.GetOpts) (stri
 	}
 	packageQuery := goqu.And(
 		goqu.Ex{"package_name": record.Package.Name},
-		goqu.Ex{"package_kind": record.Package.Kind},
+		goqu.Ex{"package_kind": wart.StringFromPackageKind(record.Package.Kind)},
 	)
 	exps = append(exps, packageQuery)
 
@@ -34,7 +35,7 @@ func buildGetQuery(record *claircore.IndexRecord, opts *datastore.GetOpts) (stri
 	if record.Package.Source.Name != "" {
 		sourcePackageQuery := goqu.And(
 			goqu.Ex{"package_name": record.Package.Source.Name},
-			goqu.Ex{"package_kind": record.Package.Source.Kind},
+			goqu.Ex{"package_kind": wart.StringFromPackageKind(record.Package.Source.Kind)},
 		)
 		or := goqu.Or(
 			packageQuery,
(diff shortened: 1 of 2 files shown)


package.go
```diff
@@ -35,7 +40,47 @@ type Package struct {
...
+	//
+	// Conceptually, a "layer" package is a higher type that describes the
+	// container image rather than the contents.
+	PackageLayer // layer
+	// An "ancestry" package is used to refer to a container layer and all
+	// previous layers in an image.
+	//
+	// Conceptually, an "ancestry" package is a higher type that describes the
+	// container image rather than the contents.
+	PackageAncestry // ancestry
```
BradLugo: I'm not sure why, but this seems a bit unclear to me. Did you have some other ideas about this name? Or maybe we should explain why this value is important.

hdonnay:
```suggestion
	// An "ancestry" package is used to refer to a container layer and all
	// previous layers (a layer and its ancestors) in an image.
	//
	// Conceptually, an "ancestry" package is a higher type that describes the
	// container image rather than the contents.
	PackageAncestry // ancestry
```

package.go
```diff
@@ -35,7 +40,47 @@ type Package struct {
...
 }
 
+// PackageKind indicates what kind of package is being described.
+type PackageKind uint
+
+func (k PackageKind) MarshalText() ([]byte, error) {
+	return []byte(k.String()), nil
+}
+
+func (k *PackageKind) UnmarshalText(text []byte) error {
```
crozzy: If this is an empty string it will Unmarshal to `binary(1)` instead of `unknown(0)` is this expected?

package.go
crozzy, line 1: How does this fit with https://github.com/quay/claircore/blob/main/toolkit/types/package_kind.go? Is that replaced with this?

package.go
```diff
@@ -35,7 +40,47 @@ type Package struct {
...
+	pos := bytes.Index([]byte(_PackageKind_name), text)
+	x := slices.Index(_PackageKind_index[1:], uint8(pos))
+	x++
+	*k = PackageKind(x)
+	return nil
+}
+
 const (
```
crozzy: I'm not sure anyone is using `BINARY` and `SOURCE` outside of claircore, but is it worth adding aliases here?

package.go
hdonnay: Yeah, this is preserving the current behavior.

package.go
hdonnay: It probably needs to be updated in turn. Will do.

package.go
hdonnay: I could but given that the type also changed, I don't know how much use it'd be.

package.go
crozzy: This order is now different: Unknown/Binary/Source vs Unknown/Source/Binary, although it seems like the toolkit one isn't actually used anywhere, should we just delete it?

package.go
crozzy: ok, fair

package.go
hdonnay: I don't understand what you mean. If you mean different to the `claircore.PackageKind`, then yes but they're different types.

I think I had this in here to facilitate moving the opposite way I did (like with the `cpe` package). I think this should stay if only for actual API reasons. I could also have this be *the* type instead of a new one in `claircore`.

crozzy: My thought was just that it's confusing having 2 types to represent what I think is the same thing, but I get that it's annoying to not have it next to the Package type. My preference would be to remove it from the toolkit, if we believe it's being used by someone that seems like we'd want to deprecate it and encourage movement to the new type (which has different `Unmarshal` behaviour IIUC). I could be missing something

PR #1782, opened 2026-03-19
migrations: SQL formatting

hdonnay:
This change adds a `generate` directive that runs an SQL formatter via `go run`. It also runs the formatter over the existing SQL.

### DIFF
--- a/datastore/postgres/migrations/migrations.go
+++ b/datastore/postgres/migrations/migrations.go
@@ -101,5 +101,6 @@ func runMigrations(ctx context.Context, cfg *pgx.ConnConfig, table pgx.Identifie
 	return nil
 }
 
+//go:generate find . -name *.sql -exec go run github.com/wasilibs/go-sql-formatter/v15/cmd/sql-formatter@latest --language postgresql --fix {} ;
 //go:embed */*.sql
 var sys embed.FS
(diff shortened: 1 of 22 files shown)


datastore/postgres/migrations/migrations.go
```diff
@@ -101,5 +101,6 @@ func runMigrations(ctx context.Context, cfg *pgx.ConnConfig, table pgx.Identifie
 	return nil
 }
 
+//go:generate find . -name *.sql -exec go run github.com/wasilibs/go-sql-formatter/v15/cmd/sql-formatter@latest --language postgresql --fix {} ;
```
hdonnay, line 104: This is the only "real" change, everything else is whitespace changes as a result of running `go generate`.

dcaravel, line 104: For consideration - is it worth referencing a specific tag vs. latest here? Thinking of risk if repo is compromised, a tag + sumdb may help detect.

hdonnay: So, this tool recommends against using `go tool` for dependency tree reasons.
I'd be wary of pinning a tag because then there's no good way to keep it updated.
Considering that its use is completely offline (we're checking in and reviewing the output) and any exploit would need to be _in_ the wasm runtime or defeat the wasm sandbox, I think it's very low risk to just use `latest`.

dcaravel: > and any exploit would need to be in the wasm runtime or defeat the wasm sandbox, I think it's very low risk to just use latest.

Fair enough - counter-point: If that repo were compromised the attacker would likely remove the Wasm sandbox entirely and add malicious code, which would run directly on our machines during go generate.

PR #1783, opened 2026-03-19
claircore: aliases

hdonnay:
This creates the `Alias` type, adds it to the `Vulnerability` type, and adds relevant database tables.

It does not add any code that makes use of these types.

### DIFF
--- a/alias_test.go
+++ b/alias_test.go
@@ -1,0 +1,68 @@
+package claircore
+
+import (
+	"fmt"
+	"testing"
+	"unique"
+)
+
+func alias(space, name string) Alias {
+	return Alias{
+		Space: unique.Make(space),
+		Name:  name,
+	}
+}
+
+func ExampleAlias_uri() {
+	fmt.Println(alias("https://example.com/", "CVE-2014-0160"))
+	// Output:
+	// https://example.com/#CVE-2014-0160
+}
+
+func ExampleAlias_cve() {
+	fmt.Println(alias("CVE", "2014-0160"))
+	// Output:
+	// CVE-2014-0160
+}
+
+func ExampleAlias_Equal() {
+	a, b := alias("CVE", "2014-0160"), alias("CVE", "2014-0160")
+	fmt.Println("equal:", a.Equal(b))
+	// Output:
+	// equal: true
+}
+
+func TestAlias(t *testing.T) {
+	t.Run("Valid", func(t *testing.T) {
+		t.Run("OK", func(t *testing.T) {
+			a := Alias{Space: unique.Make("TEST"), Name: "1"}
+			if got, want := a.Valid(), true; got != want {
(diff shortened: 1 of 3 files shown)


vulnerability.go
```diff
@@ -38,6 +38,17 @@ type Vulnerability struct {
...
+	// Self is an Alias that is the "identity" for this Vulnerability.
+	//
+	// This should be a system-wide, external identifier.
+	Self Alias
+	// Aliases is a set of aliases for the same abstract software flaw.
+	//
+	// For example, GHSA advisories frequently also reference CVE identifiers.
+	//
+	// This may contain the "Self" alias.
+	Aliases []Alias
```
dcaravel, line 52: Were the json tags left out intentionally?

alias_test.go
dcaravel, line 1: nit: consider adding test for `Valid()`

datastore/postgres/migrations/matcher/17-aliases.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE IF NOT EXISTS alias_namespace (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  namespace TEXT UNIQUE NOT NULL
+);
+
+COMMENT ON TABLE alias_namespace IS 'Contains namespaces for aliases. Usually short IDs like "CVE" or "GHSA".';
+
+CREATE TABLE IF NOT EXISTS alias (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  namespace INTEGER REFERENCES alias_namespace (id) ON DELETE CASCADE,
```
dcaravel: Should `namespace` have `NOT NULL` ?

vulnerability.go
dcaravel, line 52: Would this be easier to digest if it was prescriptive? - ie: "will always contain self alias" or "will never contain self alias"

hdonnay: Yes.

I'm planning on landing code in `clair` to split apart its JSON representation from the autogenerated JSON based on the `claircore` types. Since it'll have its own formatting, I'm going to not bother with tags. The JSON formatting was never purposefully an API.

datastore/postgres/migrations/matcher/17-aliases.sql
hdonnay: Yeah I thought the foreign key implied that, let me double-check.

vulnerability.go
hdonnay: My thinking is that there's two use-cases here: produced by updater code and produced by the database layer. Picking one way means forcing some work onto one of those "users".

hdonnay: I documented how instances of this type coming from the database should behave.

dcaravel (approved):
LGTM

PR #1790, opened 2026-03-25
rhel: update backup content-manifest path

crozzy:
The backup content-manifest path was incorrect as it didn't include the content_manifests directory.

### DIFF
--- a/rhel/rhel.go
+++ b/rhel/rhel.go
@@ -16,24 +16,31 @@ import (
 	"log/slog"
 )
 
+var contentManifestBlobs = []string{
+	`root/buildinfo/content_manifests/*.json`,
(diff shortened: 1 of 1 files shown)


rhel/rhel.go
hdonnay: I think this is better written as a loop of globs that returns the first matched path.

hdonnay (changes requested):

rhel/rhel.go
crozzy: I added the `var` for readability

rhel/rhel.go
```diff
@@ -16,24 +16,31 @@ import (
 	"log/slog"
 )
 
+var contentManifestPaths = []string{
```
hdonnay:
```suggestion
var contentManifestGlobs = []string{
```

PR #1794, opened 2026-03-26
libindex: layer metrics

hdonnay:
This adds a set of metrics for tracking layer count, size, and request duration.

Additionally adds a package for helping with units, and fixes some lints flagged by `revive`.

### DIFF
--- a/libindex/doc.go
+++ b/libindex/doc.go
@@ -1,0 +1,5 @@
+// Package libindex is the claircore Indexer.
+//
+// The Indexer is the component that analyzes container images (or
+// container-like filesystems) and produces IndexReports.
+package libindex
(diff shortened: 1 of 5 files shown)


libindex/instrumentation.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// ArenaMetrics holds metrics for a single [RemoteFetchArena].
+//
+// The resulting metrics are scoped for the arena.
+type arenaMetrics struct {
+	sizes     metric.Int64Histogram
+	duration  metric.Float64Histogram
+	requests  metric.Int64Counter
+	cacheHits metric.Int64Counter
```
crozzy: I can't see `cacheHits` used anywhere

hdonnay: Ah yeah, it's just an attribute on `requests`. Will update.

PR #1796, opened 2026-03-27
claircore: add "Invert" field

hdonnay:
This adds a field that inverts the meaning of a `Vulnerability`, and the database modification to store it.

### DIFF
--- a/vulnerability.go
+++ b/vulnerability.go
@@ -50,6 +50,10 @@ type Vulnerability struct {
 	// For instances from a claircore "datastore" implementation, this will also
 	// include the "Self" alias.
 	Aliases []Alias
+
+	// Invert means this Vulnerability should be interpreted as an assertion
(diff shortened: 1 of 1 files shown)


vulnerability.go
```diff
@@ -50,6 +50,10 @@ type Vulnerability struct {
 	// For instances from a claircore "datastore" implementation, this will also
 	// include the "Self" alias.
 	Aliases []Alias
+
+	// Invert means this Vulnerability should be interpreted as an assertion
+	// that matched packages are NOT vulnerable to the indicated flaw.
+	Invert bool
```
crozzy, line 56: Should this either be explicitly omitted or given a json tag?

hdonnay: My thinking is that there's no canonical JSON encoding anymore, so we shouldn't add any tags.

PR #1808, opened 2026-04-01
fix(dpkg): handle missing files in distroless status.d gracefully

yuvalk:
## Summary
- The distroless dpkg scanner fails the entire scan when a file in `status.d/` cannot be opened (e.g., a dangling symlink to a `.list` file in `info/` that doesn't exist in the layer)
- Now gracefully skips missing files with a warning log instead of aborting, matching the error-handling pattern used elsewhere in the dpkg scanners
- Adds a test that constructs a synthetic layer with a dangling symlink in `status.d/` to verify the fix

## Test plan
- [x] New unit test `TestDistrolessMissingListFile` passes — constructs a layer with a dangling symlink and verifies the scan succeeds
- [ ] Verify against the real container image that originally triggered the bug

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dpkg/distroless_scanner_test.go
+++ b/dpkg/distroless_scanner_test.go
@@ -1,7 +1,14 @@
 package dpkg
 
 import (
+	"archive/tar"
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"fmt"
+	"io"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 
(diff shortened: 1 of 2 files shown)


dpkg/distroless_scanner_test.go
```diff
@@ -61,3 +68,79 @@ func TestDistrolessLayer(t *testing.T) {
...
+		Name:     "var/lib/dpkg/status.d/gcc-14-base.list",
+		Linkname: "../info/gcc-14-base.list",
+		Mode:     0644,
+		ModTime:  now,
+	}); err != nil {
+		t.Fatal(err)
+	}
+	if err := tw.Close(); err != nil {
+		t.Fatal(err)
+	}
```
hdonnay: This should use [`test.GenerateFixture`](https://pkg.go.dev/github.com/quay/claircore/test#GenerateFixture).

dpkg/distroless_scanner_test.go
```diff
@@ -61,3 +68,79 @@ func TestDistrolessLayer(t *testing.T) {
 		t.Fatal(cmp.Diff(ps, want))
 	}
 }
+
+func TestDistrolessMissingListFile(t *testing.T) {
+	ctx := context.Background()
```
hdonnay: This should use [`test.Logging`](https://pkg.go.dev/github.com/quay/claircore/test#Logging).

dpkg/distroless_scanner_test.go
```diff
@@ -61,3 +68,79 @@ func TestDistrolessLayer(t *testing.T) {
...
+	}); err != nil {
+		t.Fatal(err)
+	}
+	if err := tw.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	desc := claircore.LayerDescription{
+		URI:       "file:///dev/null",
+		Digest:    fmt.Sprintf("sha256:%x", h.Sum(nil)),
```
hdonnay: This doesn't need the "correct" sha256, just a valid sha256, so this could be replaced with `sha256:aaa...`.

hdonnay (changes requested):

PR #1822, opened 2026-04-08
Acceptance testing

crozzy:


test/acceptance/testdata/cve-2024-0727-openssl.json
crozzy, line 1: In two minds about adding local fixtures, they do test functionality and result correctness without needing external deps, but they are sort of a collection of real and fake data. I might just clean it up to a couple of examples.

test/acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,648 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("parse CSAF/VEX: %w", err)
+		}
+
+		trackingID := c.Document.Tracking.ID
+		slog.DebugContext(ctx, "parsing CSAF/VEX document", "tracking_id", trackingID, "vuln_count", len(c.Vulnerabilities))
+
+		// Detect RHEL VEX documents by publisher namespace and use the specialised parser.
+		// RHEL VEX has complex product relationships and CPE-based repository matching.
+		if c.Document.Publisher.Namespace == "https://www.redhat.com" {
```
crozzy, line 279: I think this is strong enough, but other suggestions welcome

test/acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,648 @@
...
+	for _, productID := range v.ProductStatus["fixed"] {
+		vs, err := a.createVulnerabilitiesFromProduct(ctx, c, productID, trackingID, desc, links, statusFixed, v)
+		if err != nil {
+			slog.WarnContext(ctx, "skipping product", "product_id", productID, "reason", err)
+			continue
+		}
+		vulns = append(vulns, vs...)
+	}
+
+	// TODO(crozzy): Process known_not_affected products here
```
crozzy, line 350: Might need to sync a little here with the not affected work

test/acceptance/acceptance.go
```diff
@@ -0,0 +1,128 @@
...
+	"testing"
+
+	"github.com/quay/claircore/toolkit/fixtures"
+)
+
+// Auditor is the interface that vulnerability auditors must implement.
+type Auditor interface {
+	// Audit analyses the image at ref using the provided CSAF documents
+	// and returns vulnerability findings.
+	Audit(ctx context.Context, t testing.TB, ref string, csafDocs [][]byte) ([]Result, error)
```
crozzy: I know we had some discussion around adding `testing.TB` to the interface method, it was to be able to use the in-built layer caching in claircore but open to other suggestions

test/acceptance/compare.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	// Check for missing expected results.
+	for _, e := range expected {
+		k := key{e.ID, e.Product}
+		if seen[k] {
+			continue
+		}
+		switch e.Status {
+		case fixtures.StatusUnknown:
```
crozzy: Honestly, I'm on the fence about `fixtures.StatusUnknown` being an actual status

acceptance/integration_test.go
```diff
@@ -0,0 +1,277 @@
+//go:build integration
```
hdonnay, line 1: Why is this build-tagged instead of using `integration.Skip`?

acceptance/main.go
hdonnay, line 1: This should probably be in a `test/acceptance/cmd/fixture` package so that `go run` and/or `go tool` can work.

acceptance/acceptance.go
```diff
@@ -0,0 +1,128 @@
...
+	"testing"
+
+	"github.com/quay/claircore/toolkit/fixtures"
+)
+
+// Auditor is the interface that vulnerability auditors must implement.
+type Auditor interface {
+	// Audit analyses the image at ref using the provided CSAF documents
+	// and returns vulnerability findings.
+	Audit(ctx context.Context, t testing.TB, ref string, csafDocs [][]byte) ([]Result, error)
```
hdonnay: I think if it uses the `testing` package in the exported API, it should be in the `test/` hierarchy. That's a simple rule that helps keep the `testing` package out of our "real" API surface.

acceptance/acceptance.go
```diff
@@ -0,0 +1,128 @@
...
+			m.TrackingID, m.ProductID, m.Actual, m.Expected)
+	}
+	for _, m := range cmp.Misses {
+		t.Errorf("MISSING %s/%s: expected %s", m.ID, m.Product, m.Status)
+	}
+
+	if len(cmp.Extras) > 0 {
+		t.Logf("note: scanner reported %d results not in fixture", len(cmp.Extras))
+		for _, e := range cmp.Extras {
+			t.Logf("extra: %s/%s (%s) [pkg: %s]", e.TrackingID, e.ProductID, e.Status, e.Package)
```
hdonnay, line 122:
```suggestion
			t.Logf("EXTRA %s/%s: %s (package: %s)", e.TrackingID, e.ProductID, e.Status, e.Package)
```

This message format should probably be similar to the other "reporting" messages.

acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,651 @@
...
+
+	// Set up indexer pool
+	indexerPool := cfg.IndexerPool
+	if indexerPool == nil {
+		indexerPool, err = pgxpool.New(ctx, cfg.IndexerDSN)
+		if err != nil {
+			return nil, fmt.Errorf("indexer db connect: %w", err)
+		}
+	}
+	indexerStore, err := postgres.InitPostgresIndexerStore(ctx, indexerPool, true)
```
hdonnay, line 110: Use the `migrations` package directly.

acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,651 @@
...
+
+	// Set up matcher pool
+	matcherPool := cfg.MatcherPool
+	if matcherPool == nil {
+		matcherPool, err = pgxpool.New(ctx, cfg.MatcherDSN)
+		if err != nil {
+			return nil, fmt.Errorf("matcher db connect: %w", err)
+		}
+	}
+	matcherStore, err := postgres.InitPostgresMatcherStore(ctx, matcherPool, true)
```
hdonnay, line 136: Use the `migrations` package directly.

acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,651 @@
...
+	}
+	if a.matcher != nil {
+		a.matcher.Close(ctx)
+	}
+	if a.cachedArena != nil {
+		if err := a.cachedArena.Close(ctx); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errors.Join(errs...)
```
hdonnay, line 173:
```suggestion
	errs := make([]error, 3)
	if a.indexer != nil {
		errs[0] = a.indexer.Close(ctx)
	}
	if a.matcher != nil {
		errs[1] = a.matcher.Close(ctx)
	}
	if a.cachedArena != nil {
		errs[2] = a.cachedArena.Close(ctx)
	}
	return errors.Join(errs...)
```

`Errors.Join` filters out `nil` entries, so it's simpler to just catch all the possible `error`s.

acceptance/acceptance.go
hdonnay, line 17: Why `[][]byte` instead of `[]io.Reader` or `iter.Seq` of `[]byte` or `io.Reader`?

acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,648 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("parse CSAF/VEX: %w", err)
+		}
+
+		trackingID := c.Document.Tracking.ID
+		slog.DebugContext(ctx, "parsing CSAF/VEX document", "tracking_id", trackingID, "vuln_count", len(c.Vulnerabilities))
+
+		// Detect RHEL VEX documents by publisher namespace and use the specialised parser.
+		// RHEL VEX has complex product relationships and CPE-based repository matching.
+		if c.Document.Publisher.Namespace == "https://www.redhat.com" {
```
hdonnay: Feels bad that we need special Red Hat handling :/

acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,651 @@
...
+		return claircore.Low
+	case "negligible", "none":
+		return claircore.Negligible
+	default:
+		return claircore.Unknown
+	}
+}
+
+// SeverityFromCVSSScore converts a CVSS base score to claircore.Severity.
+func severityFromCVSSScore(score float64) claircore.Severity {
```
hdonnay, line 517: This seems like it's basically duplicating [`cvss.QualitativeScore`](https://pkg.go.dev/github.com/quay/claircore/toolkit@v1.6.0/types/cvss#QualitativeScore)?

acceptance/compare.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	// Check for missing expected results.
+	for _, e := range expected {
+		k := key{e.ID, e.Product}
+		if seen[k] {
+			continue
+		}
+		switch e.Status {
+		case fixtures.StatusUnknown:
```
hdonnay: Yeah, I'm not totally clear if "Unknown" and "Extra" are different concepts. Isn't an "Unknown" status just an "Extra" package?

acceptance/compare_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+func TestCompare(t *testing.T) {
+	tests := []struct {
+		name     string
+		expected []fixtures.ManifestRecord
+		actual   []Result
+		want     *Comparison
+	}{
+		{
+			name: "all match affected",
```
hdonnay, line 19: The names of these subtests should be in PascalCase, so that `go test`'s `-run` flag is consistent.

rhel/vex/parser.go
```diff
@@ -116,6 +117,90 @@ func (u *Updater) DeltaParse(ctx context.Context, contents io.ReadCloser) ([]*cl
 	return vulns, deleted, nil
 }
 
+// Parser parses individual RHEL CSAF/VEX documents into claircore vulnerabilities.
+// It maintains internal caches for repository and product lookups, so reusing
+// the same Parser instance across multiple documents is more efficient than
+// creating a new one for each document.
+type Parser struct {
```
hdonnay, line 124: I think the docs on the `Parser` type needs some work to explicitly call out that's it's caching the _`claircore` representation_ (e.g. Vulnerabilities and Repositories) because I was briefly confused about caching CSAF objects, which are per-document.

hdonnay (changes requested):

test/acceptance/compare.go
crozzy: Yeah, just going to remove this `case`

test/acceptance/auditor_claircore.go
crozzy: yeah

acceptance/integration_test.go
crozzy: It was all outside of the existing test framework at first, I believe this is just a hangover of that, going to change

test/acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,651 @@
...
+
+	// Set up matcher pool
+	matcherPool := cfg.MatcherPool
+	if matcherPool == nil {
+		matcherPool, err = pgxpool.New(ctx, cfg.MatcherDSN)
+		if err != nil {
+			return nil, fmt.Errorf("matcher db connect: %w", err)
+		}
+	}
+	matcherStore, err := postgres.InitPostgresMatcherStore(ctx, matcherPool, true)
```
crozzy: I think I did this wrong, in theory with the all the testing DB setup logic, even with an external PG DB, it should be migrated before being used, ie setting the `POSTGRES_CONNECTION_STRING`. I think this manual setup is probably not needed.

test/acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,651 @@
...
+	}
+	if a.matcher != nil {
+		a.matcher.Close(ctx)
+	}
+	if a.cachedArena != nil {
+		if err := a.cachedArena.Close(ctx); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errors.Join(errs...)
```
crozzy: :facepalm:

test/acceptance/acceptance.go
crozzy: We have to read the doc twice (to see which way to parse it and there's a few times where we surface to the user how many docs there are (we can count after the seq is finished I suppose). 

What do you think? `iter.Seq` of `io.Reader`?

test/acceptance/auditor_claircore.go
```diff
@@ -0,0 +1,651 @@
...
+		return claircore.Low
+	case "negligible", "none":
+		return claircore.Negligible
+	default:
+		return claircore.Unknown
+	}
+}
+
+// SeverityFromCVSSScore converts a CVSS base score to claircore.Severity.
+func severityFromCVSSScore(score float64) claircore.Severity {
```
crozzy: Kind of but to use it we'd need to:
1) Parse the float as v2, v3, v4 etc
2) Call `cvss.QualitativeScore()`
3) Covert the  cvss.Qualitative score to claircore.Severity

Maybe the correct thing to do is add a cvss.Qualitative score to claircore.Severity function to the CVSS toolkit?

rhel/vex/parser.go
crozzy: will update :+1:

test/acceptance/cmd/fixture/main.go
crozzy: done

test/acceptance/auditor_claircore.go
hdonnay: Well, `claircore` can't import `toolkit/types`, so...

It's annoying that this is kind of duplicated, but I guess NBD.

docs/howto/acceptance_testing.md
```diff
@@ -0,0 +1,332 @@
...
+```
+
+## Media Types
+
+| Type | Media Type | Description |
+|------|------------|-------------|
+| VEX | `application/csaf+json` | CSAF/VEX vulnerability document |
+| Expected Results | `application/vnd.com.redhat.container.acceptancetest.v1+csv` | Test expectations CSV |
+
+Compressed variants (`.../+zstd`) are also supported.
```
hdonnay:
```suggestion
Compressed variants (`...+zstd`) are also supported.
```

I think inserting a slash is needlessly complicated.

docs/howto/acceptance_testing.md
```diff
@@ -0,0 +1,332 @@
...
+})
+```
+
+## Testing with Claircore
+
+Most teams will use the built-in claircore analyzer to test their images. Here's
+a complete, copy-paste-able test file:
+
+```go
+//go:build integration
```
hdonnay:
```suggestion
```

docs/howto/acceptance_testing.md
```diff
@@ -0,0 +1,332 @@
...
+    -tag multi-vex-fixture \
+    -vex vex1.json \
+    -vex vex2.json \
+    -vex vex3.json \
+    -manifest expected.csv \
+    -repo your.registry.io/your-repo
+
+# Verify the fixture was created
+go run -mod=mod main.go list -image your.registry.io/your-repo:my-fixture-name
+```
```
hdonnay: These should use command's path, now that it's split out.

hdonnay (changes requested):
Did a read-through of the docs.

test/acceptance/compare.go
```diff
@@ -0,0 +1,84 @@
...
+			} else {
+				cmp.Mismatches = append(cmp.Mismatches, Mismatch{
+					TrackingID: r.TrackingID,
+					ProductID:  r.ProductID,
+					Expected:   exp.Status,
+					Actual:     r.Status,
+				})
+			}
+		}
+	}
```
hdonnay:
```suggestion
		switch {
		case exp.Status == fixtures.StatusAbsent:
			// Expected NOT to be in results, but it is - failure.
			cmp.Mismatches = append(cmp.Mismatches, Mismatch{
				TrackingID: r.TrackingID,
				ProductID:  r.ProductID,
				Expected:   exp.Status,
				Actual:     r.Status,
			})
		case exp.Status == r.Status:
			cmp.Matches = append(cmp.Matches, Match{
				TrackingID: r.TrackingID,
				ProductID:  r.ProductID,
				Status:     r.Status,
			})
		default:
			cmp.Mismatches = append(cmp.Mismatches, Mismatch{
				TrackingID: r.TrackingID,
				ProductID:  r.ProductID,
				Expected:   exp.Status,
				Actual:     r.Status,
			})
		}
	}
```
I find this a more intuitive way to present this, but I love me a `switch`.

test/acceptance/integration_test.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		t.Fatalf("read VEX file %q: %v", name, err)
+	}
+	return data
+}
+
+// TestAcceptanceRun tests the full acceptance framework using acceptance.Run.
+func TestAcceptanceRun(t *testing.T) {
+	integration.Skip(t)
+	integration.NeedDB(t)
```
hdonnay, line 38: Having both of these means the test will only run if the `integration` tag is provided (or the `CI` environment variable is set).

I think this is fine for an initial inclusion, but ideally we'll look in a cache (somewhere) to see what's available to use.

test/acceptance/loader.go
```diff
@@ -0,0 +1,201 @@
...
+
+// FetchBlob retrieves a blob from the registry, decompressing if needed.
+func fetchBlob(ctx context.Context, rc *regclient.RegClient, r ref.Ref, desc descriptor.Descriptor, compressed bool) ([]byte, error) {
+	blob, err := rc.BlobGet(ctx, r, desc)
+	if err != nil {
+		return nil, err
+	}
+	defer blob.Close()
+
+	if compressed {
```
hdonnay, line 197: can't this be `if slices.Contains(compressedTypes, desc.ArtifactType) {}`?

test/acceptance/acceptance.go
hdonnay: I _think_ an `iter.Seq[io.Reader]` is just a better interface? Worst case we end up turning it into `[]io.Reader` right away and reuse a `bytes.Buffer` for the actual contents.

test/acceptance/cmd/fixture/main.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/regclient/regclient/types/platform"
+	"github.com/regclient/regclient/types/ref"
+
+	"github.com/quay/claircore/toolkit/fixtures"
+)
+
+const defaultRepo = "quay.io/projectquay/clair-fixtures"
+
+// TODO (crozzy): Add this somewhere common for all CLIs to use.
+type logHandler struct{}
```
hdonnay, line 29: Why a custom handler instead of a `slog.TextHandler`?

crozzy: It seemed more cli friendly to output something like: 
```
message key=blah something=algo
```
Rather than something like:
```
time=2024-01-01T12:00:00.000Z level=INFO msg=message key=blah something=algo
```

crozzy:
@hdonnay I updated to `iter.Seq[io.Reader]`, it seems clunky but the interface is nice so LMK

docs/howto/acceptance_testing.md
```diff
@@ -0,0 +1,335 @@
...
+|--------|---------|---------------|
+| `affected` | Product is vulnerable | MUST appear in analyzer results as affected |
+| `not_affected` | Product is known to not be affected | MUST appear in analyzer results as not-affected |
+| `absent` | Product should not appear (fixed/filtered) | MUST NOT appear in analyzer results |
+
+Lines starting with `#` are treated as comments.
+
+### Uploading Fixtures
+
+Use the `main.go` tool in the acceptance directory to create fixtures:
```
hdonnay:
```suggestion
Use the `test/acceptance/cmd/fixture` tool to create fixtures:
```

docs/howto/acceptance_testing.md
```diff
@@ -0,0 +1,335 @@
...
+
+```bash
+# Navigate to the acceptance directory
+cd acceptance
+
```
hdonnay:
```suggestion
```

docs/howto/acceptance_testing.md
```diff
@@ -0,0 +1,335 @@
...
+## Running Tests
+
+### Running Integration Tests
+
+The integration tests automatically download and run embedded PostgreSQL
+binaries - no database setup required:
+
+```bash
+cd acceptance
```
hdonnay:
```suggestion
cd test/acceptance
```

docs/howto/acceptance_testing.md
```diff
@@ -0,0 +1,335 @@
...
+6. Report any mismatches or missing vulnerabilities
+
+## Implementing a Custom Auditor
+
+If you're building your own vulnerability analyzer (not using claircore), implement
+the `Auditor` interface:
+
+```go
+type Auditor interface {
+    Audit(ctx context.Context, t testing.TB, ref string, csafDocs [][]byte) ([]Result, error)
```
hdonnay:
```suggestion
    Audit(ctx context.Context, t testing.TB, ref string, csafDocs iter.Seq[io.Reader]) ([]Result, error)
```

docs/howto/acceptance_testing.md
```diff
@@ -0,0 +1,335 @@
...
+4. Match vulnerabilities against the indexed packages
+5. Compare results to the expected results CSV
+6. Report any mismatches or missing vulnerabilities
+
+## Implementing a Custom Auditor
+
+If you're building your own vulnerability analyzer (not using claircore), implement
+the `Auditor` interface:
+
+```go
```
hdonnay: Might be better to replace this code block with a `pkg.go.dev` link, so that it doesn't rot.

docs/howto/acceptance_testing.md
```diff
@@ -0,0 +1,335 @@
...
+        vulns, err := a.parseCSAF(doc)
+        if err != nil {
+            return nil, err
+        }
+        // 3. Match packages against vulnerabilities
+    }
+    
+    // 4. Return results in the expected format, the auditor will do the comparison.
+    return results, nil
+}
```
hdonnay: This could be an actual go file and included, so that the interface implementation could get typechecked.

hdonnay (changes requested):

docs/howto/acceptance_testing.md
crozzy: Good idea

crozzy:
All code moved out of docs in order avoid rotting

hdonnay (approved):

PR #1832, opened 2026-04-13
chore: account for all modules in go mod tidy check

crozzy:
This will do `go mod tidy` for each module. This came up recently when the updater/driver's go.mod and go.sum had become stale.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,8 +48,9 @@ jobs:
       - name: Go Tidy
         if: ${{ !cancelled() && steps.checkout.conclusion == 'success' && steps.setupgo.conclusion == 'success' }}
         run: |
-          trap 'echo "::error file=go.mod,title=Tidy Check::Commit would leave go.mod untidy"' ERR
-          go mod tidy
+          trap 'echo "::error title=Tidy Check::Commit would leave go.mod/go.sum untidy"' ERR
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -48,8 +48,9 @@ jobs:
       - name: Go Tidy
         if: ${{ !cancelled() && steps.checkout.conclusion == 'success' && steps.setupgo.conclusion == 'success' }}
         run: |
-          trap 'echo "::error file=go.mod,title=Tidy Check::Commit would leave go.mod untidy"' ERR
-          go mod tidy
+          trap 'echo "::error title=Tidy Check::Commit would leave go.mod/go.sum untidy"' ERR
+          find . -name go.mod -not -path './vendor/*' -printf '%h\n' |
```
crozzy: Annoyingly couldn't get `go list` to work due to `..pkg/mod/github.com/quay/claircore/toolkit@v1.6.0` path when local replace isn't there.

crozzy: I could swap out the path from `go list` and remove the version, but at that point it feels like `find` is just easier.

hdonnay: This should have the same `find` invocation as the step below.

hdonnay (changes requested):

hdonnay (approved):

PR #1842, opened 2026-04-24
test: propagate Fail/Skip calls from RealizeLayers goroutines

hdonnay:
Found this because my local tests (sans `integration`) started failing.


crozzy (approved):
LGTM

Issue #1844, opened 2026-04-27
SQLite WAL/SHM spool side-files leak in os.TempDir() after RPM scan


zandwang:
## Summary

When `claircore`'s RPM scanner runs against an SQLite-backend RPM database
(Rocky 9 / RHEL 9 / Fedora 35+), it leaves orphaned `*-wal` and `*-shm` files
in `os.TempDir()` after every scan, e.g.:

```text
$ ls /tmp | grep rpm
rpm.package.sqlite.4257389849-shm
rpm.package.sqlite.4257389849-wal
```

The main spool file (`rpm.package.sqlite.4257389849`) is correctly unlinked,
but its SQLite companion files survive. They accumulate over the lifetime of
any long-running process (e.g. an agent that scans on a schedule).

This does **not** reproduce on CentOS 6/7 because those distros use the BDB
backend, which is parsed in-memory without opening a real SQLite connection.

## Root cause

In `internal/rpm/find.go`:

```go
func (c *databaseCleanup) Close() (err error) {
    if c.spool != nil {
        err = errors.Join(err, c.spool.Close(), os.Remove(c.spool.Name())) // (1) main file unlinked
    }
    if c.close != nil {
        err = errors.Join(err, c.close())                                  // (2) sqlite.Close runs *after*
    }
    return err
}
```

Two things go wrong, in order:

1. The main spool file is unlinked **before** `sdb.Close()` is invoked. By the
   time the SQLite driver (`modernc.org/sqlite`) gets a chance to perform its
   clean shutdown — which is what removes WAL/SHM — the main DB file has
   already been unlinked. SQLite's shutdown sequence assumes the main file is
   still present; some platform/filesystem combinations result in the side
   files being abandoned.
2. Even if SQLite did manage to unlink the WAL/SHM during shutdown, this
   ordering is fragile: SQLite's WAL checkpoint / shutdown semantics are
   documented around the main file existing on disk.

`Open` in `internal/rpm/sqlite/sqlite.go` opens the connection with
`_pragma=query_only(1)`, which makes the connection read-only at the SQL level,
but the driver still creates `-wal` / `-shm` companions on the filesystem.

PR #1845, opened 2026-04-27
rpm: prevent sqlite WAL/SHM spool files from leaking on Close

zandwang:
databaseCleanup.Close unlinked the spool file before closing the SQLite handle. SQLite's shutdown sequence (WAL checkpoint, -wal/-shm removal) relies on the main DB file still being on disk, so on some platform/filesystem combinations the -wal and -shm side files were abandoned in os.TempDir().

This was visible to long-running processes scanning RPM-on-SQLite databases (Rocky 9, RHEL 9, Fedora 35+, AlmaLinux 9), which accumulated rpm.package.sqlite.<n>-wal / -shm files indefinitely. CentOS 6/7 (BDB backend) was unaffected because those code paths never open a real SQLite connection.

Close the SQLite handle first, then unlink the spool. As a belt-and-suspenders measure, also remove the -wal and -shm siblings explicitly, ignoring ErrNotExist, in case a driver version skips them during clean shutdown.

### DIFF
--- a/internal/rpm/find.go
+++ b/internal/rpm/find.go
@@ -226,7 +226,7 @@ func OpenDB(ctx context.Context, sys fs.FS, found FoundDB) (*Database, error) {
 			return nil, fmt.Errorf("internal/rpm: unable to open sqlite db: %w", err)
 		}
 		db.headers = sdb
-		cleanup.close = sdb.Close
+		cleanup.close = sqliteCleanup(spool.Name(), sdb.Close)
 	case kindBDB:
 		var bpdb bdb.PackageDB
 		if err := bpdb.Parse(spool); err != nil {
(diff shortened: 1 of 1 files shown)


zandwang:
Fixes  #1844

internal/rpm/find.go
```diff
@@ -259,11 +259,23 @@ type databaseCleanup struct {
...
+	// platform/filesystem combinations.
 	if c.close != nil {
 		err = errors.Join(err, c.close())
 	}
+	if c.spool != nil {
+		name := c.spool.Name()
+		err = errors.Join(err, c.spool.Close(), os.Remove(name))
+		// Belt-and-suspenders: SQLite normally removes these during clean
+		// shutdown, but some driver versions leave them behind.
+		for _, suffix := range []string{"-wal", "-shm"} {
```
crozzy, line 281: I think this sqlite-specific code would probably fit better in the dedicated sqlite `cleanup.close()` function instead of the generic `Close()`

crozzy (changes requested):
Thanks for the PR, please see the comments

crozzy:
This PR needs to be signed off to pass the DCO check

internal/rpm/find.go
zandwang: I just pushed a new commit. I’m not sure I got your point, so feel free to close the PR and address the issue directly.

PR #1846, opened 2026-04-27
dnf: cleanup sqlite DBs

crozzy:
In the success path it was probable that the DNF sqlite DB was left of the filesystem after the DB was Closed (with Close()).


crozzy:
You can check this with:
```
go test ./internal/dnf/... -run 'TestFindRepoids|TestWrap' -count=1 -v
ls -l <your temp dir>
```

hdonnay (approved):
Ah, yeah, I bet I was thinking of the beautiful future when we don't need to link sqlite databases into the filesystem.

PR #1847, opened 2026-04-28
chore: add json tags to new vulnerabilityreport fields

BradLugo:
Without these tags, the unmarshaled json keys are Pascal case.

### DIFF
--- a/vulnerability.go
+++ b/vulnerability.go
@@ -42,18 +42,18 @@ type Vulnerability struct {
 	// Self is an Alias that is the "identity" for this Vulnerability.
 	//
 	// This should be a system-wide, external identifier.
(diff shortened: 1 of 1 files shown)


hdonnay:
The plan was that going forward, there's no canonical JSON representation of these objects provided by Claircore.

What problem is this causing?

BradLugo:
Nothing in particular. I was testing out a json representation to prototype in stackrox, but we could just as easily construct the claircore type instead of marshaling a string.

Should we add the `json:"-"` tag to these fields instead?

hdonnay:
Claircore depends on the objects being serialized internally (for now...), so we can't have them omitted.

PR #1848, opened 2026-04-28
filterfs: don't use /dev in test

hdonnay:
The `TestDev` test used the machine's `/dev` directory to test special files getting filtered out. However when running concurrently with tests starting and stopping PostgreSQL, the test would flap on `shm` files disappearing from underneath it.

This replaces the test with creating a fifo in a test-created directory, which will not have files change unexpectedly.

### DIFF
--- a/internal/filterfs/fs_test.go
+++ b/internal/filterfs/fs_test.go
@@ -1,7 +1,9 @@
 package filterfs
 
 import (
+	"net"
 	"os"
+	"path/filepath"
 	"testing"
 	"testing/fstest"
 )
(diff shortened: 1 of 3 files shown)


internal/filterfs/fs_test.go
```diff
@@ -19,12 +20,23 @@ func TestCurrentDir(t *testing.T) {
 	}
 }
 
-func TestDev(t *testing.T) {
+func TestPipe(t *testing.T) {
+	d := t.TempDir()
+
+	_, err := os.OpenFile(filepath.Join(d, "fifo"), os.O_CREATE|os.O_RDWR, os.ModeNamedPipe|0o755)
+	if err != nil {
+		t.Errorf("fifo: %v", err)
```
crozzy: Should this be a fatal?

internal/filterfs/fs_test.go
```diff
@@ -19,12 +20,23 @@ func TestCurrentDir(t *testing.T) {
...
+	}
+	_, err = os.Create(filepath.Join(d, "file"))
+	if err != nil {
+		t.Errorf("fiil: %v", err)
```
crozzy: Go on then, enlighten me

internal/filterfs/fs_test.go
```diff
@@ -19,12 +20,23 @@ func TestCurrentDir(t *testing.T) {
...
+
 	fileset := []string{
-		"cpu",
+		"file",
 	}
 
-	sys := New(os.DirFS("/dev"))
+	sys := New(os.DirFS(d))
 
 	if err := fstest.TestFS(sys, fileset...); err != nil {
```
crozzy: Should this test explicitly check if `fifo` has been omitted?

hdonnay:
Ah, yeah, I'm not sure this is testing the right thing now. Will rework a bit.

hdonnay:
Updated:

- use `net.Listen` to create a unix socket: The previous invocation was not actually creating a "special" file.
- use `slices.DeleteFunc`: The ruby arrow was bugging me.
- misc doc touchups

hdonnay:
@crozzy another look, please

BradLugo (commented):
LGTM, but I'd like to hear what Crozzy thinks. One minor thing: is it possible to log when an entry is filtered out? If so, what would you think about adding that?

hdonnay:
It'd be odd because we'd have to smuggle a `context.Context` in the `FS` struct. Given that this is kind of a hack with the current API, I'd rather not bother.

BradLugo:
Yeah... that's crossed my mind as well. I think it might be nice to plumb `context.Context` through to this function, so a caller could cancel this process, but maybe this is a discussion for a different PR (might need a totally different API - I don't really want a `context.Context` field in `FS` struct).

crozzy (approved):
It lgtm, `slices.DeleteFunc` is a nice fit. Logging removed files could be nice for debugging but can be discussed/implemented outside of this PR

PR #1849, opened 2026-04-29
rpm: fold WAL/SHM cleanup into sqlite close closure

### LINKED ISSUE #1844: SQLite WAL/SHM spool side-files leak in os.TempDir() after RPM scan

crozzy:
Add SQLite-specific teardown inside the kindSQLite branch via a closure.

### DIFF
--- a/internal/rpm/find.go
+++ b/internal/rpm/find.go
@@ -226,7 +226,18 @@ func OpenDB(ctx context.Context, sys fs.FS, found FoundDB) (*Database, error) {
 			return nil, fmt.Errorf("internal/rpm: unable to open sqlite db: %w", err)
 		}
 		db.headers = sdb
-		cleanup.close = sdb.Close
+		spoolPath := spool.Name()
+		cleanup.close = func() error {
+			err := sdb.Close()
+			// Belt-and-braces: SQLite normally removes these during clean
+			// shutdown, but some driver versions leave them behind.
+			for _, suffix := range [2]string{"-wal", "-shm"} {
+				if rmErr := os.Remove(spoolPath + suffix); rmErr != nil && !errors.Is(rmErr, os.ErrNotExist) {
+					err = errors.Join(err, rmErr)
+				}
+			}
+			return err
+		}
 	case kindBDB:
 		var bpdb bdb.PackageDB
 		if err := bpdb.Parse(spool); err != nil {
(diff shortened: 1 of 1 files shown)


crozzy:
Based on https://github.com/quay/claircore/pull/1845 (a community PR)

crozzy:
fixes https://github.com/quay/claircore/issues/1844

internal/rpm/find.go
```diff
@@ -226,7 +226,19 @@ func OpenDB(ctx context.Context, sys fs.FS, found FoundDB) (*Database, error) {
 			return nil, fmt.Errorf("internal/rpm: unable to open sqlite db: %w", err)
 		}
 		db.headers = sdb
-		cleanup.close = sdb.Close
+		spoolPath := spool.Name()
+		cleanup.close = func() error {
+			var err error
+			err = errors.Join(err, sdb.Close())
```
hdonnay:
```suggestion
			err := sdb.Close()
```

hdonnay (changes requested):
Small simplification

internal/rpm/find.go
crozzy: :facepalm:

hdonnay (approved):

PR #1850, opened 2026-04-29
claircore: add "not vulnerable" lookup to `VulnerabiltityReport`

hdonnay:
This is to be the API that's actually exposed for this information.

I think this is all that's needed on this end.


crozzy (approved):
LGTM

hdonnay:
will merge with green CI

PR #1860, opened 2026-04-30
cicd: allow deploy-docs to trigger from comments

hdonnay:
Considering just letting the crypto miners have at it if it means that the documentation will actually get built.

### DIFF
--- a/.github/workflows/fast-forward.yml
+++ b/.github/workflows/fast-forward.yml
@@ -31,7 +31,7 @@ jobs:
           merge: true
           comment: on-error
   deploy-docs:
-    if: ${{ success() && github.base_ref == 'main' }}
+    if: ${{ success() && github.event.issue.pull_request }}
     permissions:
       contents: write
     needs: fast-forward


BradLugo:
Thoughts on making a separate workflow?

hdonnay:
The issue is that workflows can't implicitly trigger other workflows. There's already a "push to main" trigger.

.github/workflows/fast-forward.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           merge: true
           comment: on-error
   deploy-docs:
-    if: ${{ success() && github.base_ref == 'main' }}
+    if: ${{ success() && github.event.issue.pull_request }}
```
crozzy, line 34: If we end up with a release branch and backport things to it in PRs that are merged with `/fast-forward` will that not temporarily override the docs until another PR is merged to `main`?

crozzy: edit: no the `update.yml` already checks I believe

crozzy (approved):

BradLugo:
Should we move the `deploy-docs` job to `.github/workflows/main.yml`?

hdonnay:
> Should we move the `deploy-docs` job to `.github/workflows/main.yml`?

Could do, but I'd need to check what the releases do, as well.

PR #1861, opened 2026-04-30
Hum distro

crozzy:
Hummingbird being a Red Hat product but having a non-typical RHEL
os-release file means it needs a special-case in the RHEL distribution
detector.

Also remove `redhat-release` support (and oracle hack)


crozzy:
"distributions": {
        "1": {
            "id": "1",
            "did": "hummingbird",
            "name": "Hummingbird OS",
            "version": "20251124",
            "version_code_name": "",
            "version_id": "20251124",
            "arch": "",
            "cpe": "cpe:2.3:a:redhat:hummingbird:1:*:*:*:*:*:*:*",
            "pretty_name": "Hummingbird OS 20251124"
        }
    },

rhel/distributionscanner_test.go
```diff
@@ -22,7 +23,7 @@ func TestDistributionScanner(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
-			d, err := findDistribution(sys)
+			d, err := findDistribution(context.Background(), sys)
```
BradLugo: nit (optional): could pass the `test.Logging()` context just in case we make use of slog in the future
```suggestion
			d, err := findDistribution(test.Logging(t), sys)
```

rhel/distributionscanner.go
```diff
@@ -45,64 +48,53 @@ func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]
...
+				DID:        m["ID"],
+				VersionID:  m["VERSION_ID"],
+				Version:    m["VERSION"],
+				PrettyName: m["PRETTY_NAME"],
+			}
+			if s := m["CPE_NAME"]; s != "" {
+				if wfn, err := cpe.Unbind(s); err == nil {
+					d.CPE = wfn
+				}
+			}
```
BradLugo, line 99: nit (optional): could write a log (maybe debug?) if we don't find the CPE name or encounter an error during unbinding.

rhel/distributionscanner.go
```diff
@@ -45,64 +48,53 @@ func (ds *DistributionScanner) Scan(ctx context.Context, l *claircore.Layer) ([]
...
+			return mkRelease(num), nil
 		}
-		num, err := strconv.ParseInt(string(ms[1]), 10, 64)
+		m, err := osrelease.Parse(ctx, bytes.NewReader(b))
 		if err != nil {
-			return nil, fmt.Errorf("rhel: unexpected error reading files: %w", err)
+			continue
+		}
+		if m["ID"] == "hummingbird" {
```
BradLugo: nit (optional): I saw this string comparison function in the ubuntu distro scanner:


```suggestion
		if strings.EqualFold(m["ID"], "hummingbird") {
```

BradLugo (dismissed):
LGTM, with some optional nits

rhel/distributionscanner.go
crozzy: Yeah, I suppose as there is no official guidance we should be flexible.

hdonnay (approved):
LGTM

I do like removing the funky Oracle bits 👍

PR #1863, opened 2026-05-04
test/vex: move exposed `testing` use to dedicated package

hdonnay:
As a general rule, code that's exported and uses the `testing` package in its API should go under the `test` hierarchy. Ultimately, we shouldn't have non-test builds pull in the `testing` package.

See-Also: CLAIRDEV-238


hdonnay:
There's some tracing changes that snuck in here, but they should be harmless to beneficial (eventually).

crozzy (approved):
SGTM

PR #1865, opened 2026-05-05
csaf: add product status enumerations

hdonnay:
This records the valid product statuses, per CSAF spec.


jvdm (approved):
Reviewed the statuses comparing to the referenced spec.

PR #1866, opened 2026-05-05
purl: allow the registry Parse to fallback

crozzy:
Registered parsing functions with dynamic namespaces were not being looked up correctly. This change adds a fallback to the registry's parse function for situations when a purl namespace is unknowable at time of registration.


BradLugo (approved):
LGTM. Thank you!

PR #1871, opened 2026-05-11
VEX: add rh vex beta feed testing and parsing mitigations

crozzy:
In order to ingest the new Red Hat VEX feed we need to update a couple of things:

- Reconstruct the "self" URL as it doesn't exist in the document anymore.
- Ignore source (src) entries in the `fixed` and `known_not_affected` arrays, the system is always assuming binaries here and we're always matching to binary RPMs from the layer filesystem

### DIFF
--- a/rhel/vex/updater.go
+++ b/rhel/vex/updater.go
@@ -175,6 +175,9 @@ func (u *Updater) parserOpts() []ParserOption {
 		prog := u.fixedInCEL
 		opts = append(opts, func(p *Parser) { p.fixedInCEL = prog })
 	}
+	if u.url != nil {
+		opts = append(opts, WithBaseURL(u.url))
+	}
 	return opts
 }
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -126,7 +126,7 @@ jobs:
             cd "$dir"
             go list -m
             go mod download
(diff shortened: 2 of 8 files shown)


crozzy:
I moved the VEX integration tests to `test/periodic` because of latency, I think that is technically a better home but I'm going to prioritize cleaning up the periodic failures so we actually get good signal

test/periodic/doc.go
hdonnay, line 1: Why drop the comment explaining the flag?

hdonnay (changes requested):

test/periodic/doc.go
crozzy: I had added an example command to the comment, then thought hard-coding the test path and timeout might be a bad idea so deleted but ended up throwing the baby out with the bath water. Restored

rhel/vex/updater.go
```diff
@@ -28,7 +26,7 @@ const (
 	// BaseURL is the base url for the Red Hat VEX security data.
 	//
 	//doc:url updater
-	BaseURL = "https://security.access.redhat.com/data/csaf/v2/vex/"
+	BaseURL = "https://security.access.redhat.com/data/csaf/v2/vex-feed/"
```
crozzy: I'm open to switching in a separate PR too

hdonnay: I think that would be for the best.

hdonnay (changes requested):

hdonnay (dismissed):

crozzy:
@hdonnay sorry, turns out the rebase was a little more involved due to me circling the same files for ~ a month so I think I need another review

hdonnay (approved):

PR #1872, opened 2026-05-12
matcher: fix send on closed channel

crozzy:
It was possible for the matching process to panic if the TryGo loop sneaks in an extra worker as the Sender or Collector finishes and frees up a slot mid-loop. This would mean that an attempted write could happen on a closed channel.

This change adds workerWg and immediately decrements if TryGo doesn't, errrm, go, then does the usual thing of waiting for the workerWg before closing.

### DIFF
--- a/internal/matcher/match.go
+++ b/internal/matcher/match.go
@@ -8,7 +8,6 @@ import (
 	"log/slog"
 	"runtime"
 	"sync"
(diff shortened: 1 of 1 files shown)


hdonnay (approved):
Yeah, much better than `atomic` tricks.

PR #1770, opened 2026-03-05
rpmver: move to internal rpm version package

hdonnay:
<details>
<summary>
Committer details
</summary>
Local-Branch: hack/rpm-version
</details>
<details>
<summary>
Related changes
</summary>
<details>
<summary>
Parent changes
</summary>
<details>
<summary>
rpm: add common version comparison function (#1769)
</summary>

</details>
</details>
</details>

PR #1873, opened 2026-05-13
rhcc: fix matching for timestamp-based versions

BradLugo:
rpmver.Parse requires a string in the EVR format (version-release, with
at least one dash). For newer (e.g., Konflux-built) Red Hat images, the
rhcc detector parses a Unix timestamp-based version (e.g., "1744596866")
from the labels.json file and the VEX parser extracts matching timestamp
tags from OCI PURLs, neither of which contain a dash. This caused
rpmver.Parse to reject them with "missing separators", preventing any
vulnerability matching for containers using the new labels.json
metadata.

These changes normalize version strings in the matcher by appending "-0"
when no dash (release) is present, consistent with what rhctag.toEVR()
already does.

### DIFF
--- a/rhel/rhcc/matcher.go
+++ b/rhel/rhcc/matcher.go
@@ -64,13 +64,16 @@ func (*matcher) Vulnerable(ctx context.Context, record *claircore.IndexRecord, v
 	if record.Package.Version == "" {
 		return false, errors.New("rhcc: unable to parse package version: empty version")
 	}
+	if vuln.FixedInVersion == "" {
+		return true, nil
+	}
 	pkgVer, err := rpmver.Parse(ensureEVR(record.Package.Version))
 	if err != nil {
 		return false, fmt.Errorf("rhcc: unable to parse version %q: %w", record.Package.Version, err)
 	}
 	fixedVer, err := rpmver.Parse(ensureEVR(vuln.FixedInVersion))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


BradLugo:
This seemed like the simplest fix. The alternative would be to adjust the versioning on both sides (i.e., the rhcc detector/IndexReport and the VEX updater/Vulnerability). This way, we can adjust the matcher without requiring any database changes for claircore instances that may already have these non-EVR string versions in their databases. I'm definitely open to alternatives, though.

BradLugo:
It just occurred to me that this wouldn't work for unfixed versions. Converting to draft and giving that a think.

BradLugo:
I went the easy route 🤷 I think we should consider throwing an error for `rpmver.Parse("-0")`.

There is an interesting situation with these changes: the rhcc matcher previously didn't match unfixed vulnerabilities. I had thought it was resolved after https://github.com/quay/claircore/commit/1c411ec75c6fd91653e0f194a1fc656d12abf09a (https://github.com/quay/claircore/pull/1770#issuecomment-4315116270), but that doesn't seem to be the case.

crozzy (changes requested):
Could you add a test case for the known affected legacy gold-repo case? I think everything else looks like it checks out

hdonnay (commented):
Yeah, LGTM with Crozzy's test

crozzy (approved):
LGTM

PR #1880, opened 2026-05-15
claircore: tweak Alias formatting

hdonnay:
This allows for URI namespaces that end in "/" to just have the name appended.


crozzy (approved):
LGTM

PR #1882, opened 2026-05-15
vex: "not affected" changes

hdonnay:
This add "not affected" support to the Red Hat VEX machinery.

The first few commits lay some ground work to unify the existing document processing, then the existing work is ported over, then the "not affected" support is added, then it's made faster.

Requires: #1880 #1881


rhel/vex/parser.go
```diff
@@ -227,223 +255,494 @@ func newRepoCache() *repoCache {
...
 				continue
 			}
-			if !checkPURL(purl) {
+			var pkgID, repoID string
+			for pkgID = rel.ProductRef; ; {
+				r := c.defaultComponent.Get(pkgID)
+				if r == nil {
+					break
+				}
+				pkgID = r.ProductRef
```
crozzy, line 392: I know this exists now so it might be a subsequent change thing but it might be worth some kind of check that this isn't in an endless cycle given some well crafted nefarious VEX file (or incompetence).

rhel/vex/parser.go
```diff
@@ -128,48 +82,94 @@ func (u *Updater) DeltaParse(ctx context.Context, contents io.ReadCloser) ([]*cl
...
+
+// ErrDeleted is used to signal that a document has been set to status
+// "deleted".
+var errDeleted = errors.New("deleted")
+
+// ParseDoc is the common code for [Parser.Parse] and [Updater.DeltaParse].
+//
+// This function always reports the document's tracking ID, if known.
+//
+// Is the documented is marked as "deleted", [errDeleted] is reported.
```
crozzy: Nit: `If`

rhel/vex/parser.go
```diff
@@ -128,48 +82,94 @@ func (u *Updater) DeltaParse(ctx context.Context, contents io.ReadCloser) ([]*cl
...
+		threatImpact:     newThreatImpactIndex(),
+		remediation:      newRemediationIndex(),
+		defaultComponent: newDefaultComponentIndex(),
 	}
 }
 
 // Parse parses a single RHEL CSAF/VEX document and returns claircore vulnerabilities.
 // The Parser's internal caches for claircore objects are reused, so parsing multiple
 // documents avoids redundant allocations for shared CPEs and repositories.
 func (p *Parser) Parse(ctx context.Context, doc []byte) ([]*claircore.Vulnerability, error) {
```
crozzy, line 110: Nit: I know this is not really used, but it is part of the public API, should we stick a "not for concurrent use" disclaimer in there as the `p.parseDoc` will reset the parser's indexes?

crozzy (commented):
Looks like a really good change, 2 general comments 1 nit

rhel/vex/parser.go
hdonnay: Yeah, I'll add a max depth check here.

rhel/vex/parser.go
```diff
@@ -494,154 +797,219 @@ func (r *ranger) resetLowest() {
...
+			if err != nil {
+				log.WarnContext(ctx, "bad purl", "reason", err, "purl", st.PURL, "missing", "PackageName")
+				continue
+			}
+			modName, err := st.Module()
+			if err != nil {
+				log.WarnContext(ctx, "bad purl", "reason", err, "purl", st.PURL, "missing", "ModuleName")
+				continue
+			}
+			sev, err := cvssVectorFromScore(st.Score)
```
jvdm, line 849: When parsing knownAffected and knownNotAffected, this is protected by a nil check on `sc.Score`.

hdonnay: Products in the "fixed" status should always have a score, I think. I'll add a check in the called function.

crozzy (approved):

PR #1883, opened 2026-05-16
osv: implement Aliases

hdonnay:
This change makes the OSV update fill in the new alias fields.

### DIFF
--- a/updater/osv/osv.go
+++ b/updater/osv/osv.go
@@ -19,6 +19,7 @@ import (
 	"slices"
 	"strings"
 	"time"
+	"unique"
 
 	"github.com/Masterminds/semver"
 
@@ -541,6 +542,10 @@ func (e *ecs) Insert(ctx context.Context, log *slog.Logger, skipped *stats, name
 	var b strings.Builder
 	var proto claircore.Vulnerability
 	proto.Name = a.ID
+	proto.Self, err = idToAlias(a.ID)
+	if err != nil {
+		return err
+	}
 	proto.Description = a.Summary
 	proto.Issued = a.Published
 	proto.Updater = e.Updater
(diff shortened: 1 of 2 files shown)


updater/osv/osv.go
```diff
@@ -787,6 +802,35 @@ func (e *ecs) Insert(ctx context.Context, log *slog.Logger, skipped *stats, name
 	return nil
 }
 
+// IdToAlias turns an OSV identifier into a [claircore.Alias].
+func idToAlias(id string) claircore.Alias {
+	space, name, ok := strings.Cut(id, "-")
+	if !ok {
+		panic(invalidIdentifier(id))
```
crozzy: Why the panic and not just return an `invalidIdentifier` error?

updater/osv/osv.go
```diff
@@ -538,9 +539,22 @@ func (e *ecs) Insert(ctx context.Context, log *slog.Logger, skipped *stats, name
 	if a.GitOnly() {
 		return nil
 	}
+	defer func() {
+		if r := recover(); r != nil {
+			if perr, ok := r.(error); ok {
+				switch {
```
crozzy: Nit: If this stays, thoughts on making this an `if`?

updater/osv/osv.go
hdonnay: Because:
1. This should be truly exceptional: OSV spec says the format is `${DB}-${ID}`
2. It's annoying to use in the `Insert` method.

It's probably better to just put the error handling in the "right" place, fair enough.

hdonnay: rewritten with normal error flow

crozzy (approved):
LGTM

PR #1886, opened 2026-05-19
bdb: support UNSORTED_HASH databases

hdonnay:
I saw this error being tripped somewhere that I now can't find.

I think this should be find to do, as the code doesn't rely on ordering in any way: it simply reads out all the populated entries.


hdonnay:
TODO:
- [ ] find a testcase

PR #1887, opened 2026-05-20
fix: make sure all whiteout files are processed

knrc:
This PR fixes the mapping for `IndexReport` files within a layer.  The previous mapping was for one file per layer, with one consequence being the inability to handle more than one whiteout file per layer.

### DIFF
--- a/whiteout/resolver.go
+++ b/whiteout/resolver.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log/slog"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/quay/claircore"
@@ -18,6 +19,16 @@ func (r *Resolver) Resolve(ctx context.Context, ir *claircore.IndexReport, layer
 	// Here we need to check if any of the packages
 	// found are moot due to whiteouts.
 	ls := newLayerSorter(layers)
+
+	whiteouts := make(map[string][]string)
+	for layer, files := range ir.Files {
+		for path, kind := range files {
+			if kind == claircore.FileKindWhiteout {
+				whiteouts[layer] = append(whiteouts[layer], path)
+			}
+		}
+	}
+
 	finalPackages := map[string]*claircore.Package{}
 	finalEnvironments := map[string][]*claircore.Environment{}
 	for pkgID, pkg := range ir.Packages {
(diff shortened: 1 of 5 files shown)


crozzy:
https://github.com/quay/claircore/pull/1320/changes/f45f1174c1d2af1311c60f23a23367aad2701c26 for reference

knrc:
> [f45f117](https://github.com/quay/claircore/commit/f45f1174c1d2af1311c60f23a23367aad2701c26) for reference

Yeah, using a slice was my original approach from a year or so back.  I changed the implementation when I looked at the way the tests were being written, assuming the intent was to coalesce the entries.

Any idea why that's not in main?

crozzy:
I believe it was a classic case of "fix a bug while working on a feature and then decide on a new approach for the feature and lose the bug-fix"

whiteout/resolver_test.go
```diff
@@ -80,18 +79,15 @@ func TestResolver(t *testing.T) {
 					"1": {{IntroducedIn: firstLayerHash}},
 					"2": {{IntroducedIn: firstLayerHash}},
 				},
-				Files: map[string]claircore.File{
-					secondLayerHash.String(): {
-						Path: "a/path/to/some/different_file/site-packages/.wh.a_package",
-						Kind: claircore.FileKindWhiteout,
+				Files: map[string]map[string]claircore.FileKind{
+					secondLayerHash.String(): map[string]claircore.FileKind{
```
crozzy, line 83: This should be inserting 3 different files into the same layer hash key rather than overwriting it. Existing bug but seems like a good time to address.

whiteout/resolver.go
```diff
@@ -29,18 +29,20 @@ func (r *Resolver) Resolve(ctx context.Context, ir *claircore.IndexReport, layer
...
+				// filepath and if the layer the whiteout file came from came after.
+				// The spec states: "Whiteout files MUST only apply to resources in
+				// lower/parent layers" hence why we don't check if they're in the same
+				// layer.
+				if kind == claircore.FileKindWhiteout && ls.isChildOf(fileLayer, packageLayer) && fileIsDeleted(pkg.Filepath, path) {
+					packageDeleted = true
+					slog.DebugContext(ctx, "package determined to be deleted",
+						"package name", pkg.Name,
+						"package file", pkg.Filepath,
+						"whiteout file", path)
```
crozzy: Can we bail out here rather than checking the rest of the files / layers?

crozzy (commented):
I left a couple of comments, both seem kind of pre-existing but it'd be good to fold in if possible

whiteout/resolver_test.go
knrc: Done

whiteout/resolver.go
knrc: I optimised this a bit more, collecting whiteout files first before going through the loop for each package.  Let me know what you think.

knrc:
Sorry, one more to rebase on the current main

hdonnay:
This LGTM. Going to rebase this myself and have the CI run.

hdonnay (approved):

PR #1888, opened 2026-05-20
alpine: add Alias support

hdonnay:
Does what it says on the tin.

### DIFF
--- a/alpine/parser.go
+++ b/alpine/parser.go
@@ -5,15 +5,16 @@ import (
 	"encoding/json"
 	"io"
 	"log/slog"
+	"unique"
 
 	"github.com/quay/claircore"
 	"github.com/quay/claircore/libvuln/driver"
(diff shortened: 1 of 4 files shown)


alpine/parser.go
```diff
@@ -5,15 +5,16 @@ import (
...
 	"github.com/quay/claircore/libvuln/driver"
 	"github.com/quay/claircore/toolkit/types"
 )
 
-const (
-	cveURLPrefix = "https://security.alpinelinux.org/vuln/"
-)
+const cveURLPrefix = "https://security.alpinelinux.org/vuln/"
+
+var space = unique.Make(cveURLPrefix)
```
crozzy, line 17: Thoughts on these being declared in secdb.go it seems like that's where they're used

crozzy (commented):
LGTM one nit

PR #1889, opened 2026-05-20
osv: ignore malformed identifiers

hdonnay:
Why have a spec if you can't violate it a little, as a treat?

### DIFF
--- a/updater/osv/osv.go
+++ b/updater/osv/osv.go
@@ -544,6 +544,7 @@ func (e *ecs) Insert(ctx context.Context, log *slog.Logger, skipped *stats, name
 	proto.Name = a.ID
 	proto.Self, err = idToAlias(a.ID)
 	if err != nil {
+		// If the main Identifier is wrong, hard error.
 		return err
 	}
 	proto.Description = a.Summary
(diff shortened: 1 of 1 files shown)


hdonnay:
This works around google/osv.dev#5426

jvdm (approved):
Nit: Does it worth adding a debug log line?

updater/osv/osv.go
```diff
@@ -620,11 +621,17 @@ func (e *ecs) Insert(ctx context.Context, log *slog.Logger, skipped *stats, name
 			b.WriteByte(' ')
 		}
 		b.WriteString(aliasBaseURL + alias)
-		a, err := idToAlias(alias)
-		if err != nil {
+		aka, err := idToAlias(alias)
+		switch {
+		case err == nil:
+		case errors.Is(err, errInvalidIdentifier):
+			// If an alias is malformed, bail.
```
crozzy, line 628: I feel like we should log something here, it seems useful for debugging why an advisory might not exist in the matcher DB

hdonnay: It's logged on line 419.

crozzy (approved):

PR #1890, opened 2026-05-20
matcher: initialize PackageNotVulnerable map

jvdm:
## Summary

Initialize the `PackageNotVulnerable` map in `EnrichedMatch()` to prevent a nil map panic when vulnerabilities with `Invert: true` are processed.

## Test

- Added unit test `TestEnrichedMatchInvert` that reproduces the panic with an `Invert: true` vulnerability

### DIFF
--- a/internal/matcher/match_test.go
+++ b/internal/matcher/match_test.go
@@ -1,0 +1,51 @@
+package matcher
+
+import (
+	"context"
+	"testing"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/datastore"
+	"github.com/quay/claircore/libvuln/driver"
+)
+
+type stubMatcher struct{}
+
+func (stubMatcher) Name() string                          { return "stub" }
+func (stubMatcher) Filter(*claircore.IndexRecord) bool    { return true }
+func (stubMatcher) Query() []driver.MatchConstraint       { return nil }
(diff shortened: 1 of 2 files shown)


internal/matcher/match_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (s stubStore) Get(context.Context, []*claircore.IndexRecord, datastore.GetOpts) (map[string][]*claircore.Vulnerability, error) {
+	return s.vulns, nil
+}
+func (stubStore) GetEnrichment(context.Context, string, []string) ([]driver.EnrichmentRecord, error) {
+	return nil, nil
+}
+
+func TestEnrichedMatchInvert(t *testing.T) {
+	ctx := t.Context()
```
BradLugo:
```suggestion
	ctx := test.Logging(t)
```

BradLugo (commented):
Should we add initialize `vr.PackageNotVulnerable` in the `Match()` function as well? Probably not since we never use `vr.PackageNotVulnerable` in that function, but figured I'd flag it just in case.

hdonnay:
This LGTM once the sign-off is added

internal/matcher/match_test.go
jvdm: Can't use `test.Logging(t)` — import cycle (`test` imports `internal/matcher`). I will leave as is.

crozzy (approved):

PR #1891, opened 2026-05-21
postgres: alias insertion deadlock

hdonnay:
This fixes a transaction conflict where concurrent updaters could attempt to insert the same alias namespaces and cause the entire update to abort.

### DIFF
--- a/datastore/postgres/updatevulnerabilities.go
+++ b/datastore/postgres/updatevulnerabilities.go
@@ -151,9 +151,8 @@ func (s *MatcherStore) updateVulnerabilities(ctx context.Context, updater string
 			$3,
 			(SELECT id FROM vuln WHERE hash_kind = $1 AND hash = $2))
 		ON CONFLICT DO NOTHING;`
-		refreshView           = `REFRESH MATERIALIZED VIEW CONCURRENTLY latest_update_operations;`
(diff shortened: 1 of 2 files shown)


crozzy:
This appears to stop the deadlocks but causes the time to insert to increase considerably. Since we are inserting the aliases and namespaces outside of the transaction (so IIUC deadlocking is not an issue), would it be possible to collect all the aliases to be inserted using the dedupe maps that already exist and fire two bulk statements for the alias and the namespace?

hdonnay:
I'll take a look and try structuring it that way.

hdonnay:
I've integrated the changes in quay/claircore#1934 and crozzy/claircore#89 here. I've kept the authorship correct as best I could.

jvdm (commented):
Can we pull `TestConcurrentUpdateVulnerabilities` from #1934 which seems to test the deadlock? 

Also, looks like this does not have the self alias fields on the test vulns in `TestGetLatestVulnerabilities` that #1934 added. Without the self fields, `bulkLinkSelf` is never exercised: no self set, `vsHashKinds` stays empty and the `bulkLinkSelf` statement never runs.

hdonnay:
I removed that test because it didn't work; I could never get it to trip. The other changes to the tests were unintentional, I'll pull those back.

crozzy (approved):
LGTM

PR #1894, opened 2026-05-26
rhcc: support VEX known_not_affected assertions for containers

jvdm:
## Summary

This PR adds support for VEX "known_not_affected" assertions at the container level. When Red Hat VEX data marks a container as not affected by a CVE, this should suppress false positives from other package-level vulnerability detections.

### Changes

1. **Scanner (commit 1)**: Adds `AncestryPackage` to RHCC scanner output alongside the existing `BinaryPackage`. This package type matches against VEX vulnerability entries that assert the container is not affected. Scanner version bumped to "2" to invalidate cached Index Reports.

2. **Matcher (commit 2)**: For inverted vulnerabilities (`Invert == true`), skips version comparison since the VEX assertion covers the entire container. Also adds a version Range to OCI PURL entries in the VEX parser.

### Example

For CVE-2024-24786, Red Hat's VEX data marks `mta/mta-rhel8-operator` as `known_not_affected`. Without this change, OSV would flag the Go protobuf module inside the container. With this change, the container-level VEX assertion suppresses that false positive.

## Test plan

- [x] Unit tests pass (`go test ./rhel/rhcc/...`)


rhel/vex/parser.go
```diff
@@ -980,6 +980,12 @@ func (c *creator) knownNotAffectedVulnerabilities(ctx context.Context, v *csaf.V
 			case packageurl.TypeOCI:
 				vuln.Repo = c.rc.Get(st.WFN, rhcc.RepositoryKey)
 				vuln.Package.Kind = types.AncestryPackage
+				vuln.Range = &claircore.Range{
```
crozzy, line 987: Maybe a comment here about how this is a "flood-gates" range.

rhel/vex/parser_test.go
```diff
@@ -666,3 +666,61 @@ func TestExtractPackageName(t *testing.T) {
 		})
 	}
 }
+
+func TestKnownNotAffectedOCI(t *testing.T) {
```
crozzy: Could this be a table (even if it's one case for now), or perhaps better, can it be a first class part of `TestParse`, where we include an expectedAncestryCount in the comparison?

rhel/rhcc/scanner.go
```diff
@@ -208,6 +208,17 @@ func (s *scanner) Scan(ctx context.Context, l *claircore.Layer) ([]*claircore.Pa
...
+		pkgs = append(pkgs, &claircore.Package{
+			Kind:              types.AncestryPackage,
+			Name:              name,
+			Version:           vr,
+			NormalizedVersion: minorRange.Version(true),
+			Source:            &src,
+			PackageDB:         p,
+			Arch:              arch,
+			RepositoryHint:    `rhcc`,
+		})
```
crozzy, line 221: This seems right, but the more important is that the `rhel/rhcc/detector.go` is updated, that's the detector that is reading the labels.json and should be how all modern RHEL containers are indexed.

crozzy (changes requested):

rhel/vex/parser.go
jvdm: Added a comment explaining the flood-gates range and why version comparison is skipped for `known_not_affected` assertions.

rhel/vex/parser_test.go
jvdm: Integrated into TestParse with an expectedAncestry field. The CVE-2024-24786 testdata produces 732 ancestry vulnerabilities.

rhel/rhcc/scanner.go
jvdm: `detector.go` is now updated as well. Its version bumped to "2". The `scanner.go` change maintains parity for legacy Dockerfile-based indexing.

crozzy (approved):

PR #1896, opened 2026-05-27
claircore: add JSON marshaling for Alias type

jvdm:
`unique.Handle[string]` does not implement `json.Marshaler`, causing `Alias.Space` to serialize as `{}` instead of the actual namespace value. This breaks vulnerability bundle serialization via jsonblob.

Add `MarshalJSON` and `UnmarshalJSON` to preserve Space values through JSON round-trips.

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -1,6 +1,7 @@
 package claircore
 
 import (
+	"encoding/json"
 	"strings"
 	"unique"
 )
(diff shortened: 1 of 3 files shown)


alias.go
```diff
@@ -49,3 +50,31 @@ func (a Alias) Equal(b Alias) bool {
 func (a Alias) Valid() bool {
 	return a.Space != unique.Handle[string]{} && a.Space.Value() != "" && a.Name != ""
 }
+
+type aliasJSON struct {
+	Space string `json:"Space"`
```
crozzy: Do we want the JSON keys to be capitalized? Seems like everything else is lower

alias_test.go
```diff
@@ -71,4 +72,36 @@ func TestAlias(t *testing.T) {
...
+
+	t.Run("JSONZero", func(t *testing.T) {
+		a := Alias{}
+		data, err := json.Marshal(a)
+		if err != nil {
+			t.Fatal(err)
+		}
+		want := `{"Space":"","Name":""}`
+		if got := string(data); got != want {
+			t.Errorf("marshal: got %s, want %s", got, want)
```
crozzy, line 104: It seems like this isn't going to be totally symmetrical, i.e unmarshaling `{"Space":"","Name":""}` will return something different from `Alias{}` should. Do you think it's worth calling unmarshal and checking `Valid()` is false?

crozzy:
blocked by https://github.com/quay/claircore/pull/1891

crozzy (approved):

hdonnay (approved):
If the outstanding deadlock isn't a concern, this should be fine.

I think we can live with the JSON marshaling continuing to be load bearing until a `v2` or new updater subsystem.

PR #1904, opened 2026-06-03
vex: only check vendor_fix remediations

crozzy:
RHSA links were being lost for vulnerabilities where the productID had multiple remediations and the last one (without a `url` was winning). This change modifies populateRemediations to only consider `vendor_fix` remediations.

### DIFF
--- a/test/acceptance/auditor_claircore.go
+++ b/test/acceptance/auditor_claircore.go
@@ -638,23 +638,24 @@ func convertVulnReport(vr *claircore.VulnerabilityReport) []Result {
 }
 
 // ExtractProductIDFromLinks extracts the VEX product ID from the URL fragment
-// appended to the last link in the Links string.
+// in the Links string. The product ID is encoded as a fragment on the VEX
+// self-link, which may be followed by a remediation URL without a fragment.
+// All links are scanned and all fragments are collected; only one product ID
+// fragment is expected, so the first one found is returned.
 func extractProductIDFromLinks(links string) string {
 	if links == "" {
 		return ""
 	}
-	parts := strings.Split(links, " ")
-	lastLink := parts[len(parts)-1]
-	u, err := url.Parse(lastLink)
-	if err != nil {
-		return ""
+	for _, link := range strings.Fields(links) {
+		u, err := url.Parse(link)
+		if err != nil || u.Fragment == "" {
+			continue
+		}
+		productID, err := url.PathUnescape(u.Fragment)
+		if err != nil {
+			continue
+		}
+		return productID
 	}
-	if u.Fragment == "" {
-		return ""
-	}
-	productID, err := url.PathUnescape(u.Fragment)
-	if err != nil {
-		return ""
-	}
-	return productID
+	return ""
 }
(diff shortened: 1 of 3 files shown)


test/acceptance/auditor_claircore.go
```diff
@@ -638,23 +638,24 @@ func convertVulnReport(vr *claircore.VulnerabilityReport) []Result {
...
+	for _, link := range strings.Fields(links) {
+		u, err := url.Parse(link)
+		if err != nil || u.Fragment == "" {
+			continue
+		}
+		productID, err := url.PathUnescape(u.Fragment)
+		if err != nil {
+			continue
+		}
+		return productID
```
dcaravel, line 658: Fragments can appear in links naturally, for example in [CVE-2024-13176](https://security.access.redhat.com/data/csaf/v2/vex/2024/cve-2024-13176.json)

```
https://www.oracle.com/security-alerts/cpuapr2025.html#AppendixMSQL
```

Given that this pulls the first fragment found and the self link with product ID fragment is always last - starting at the last link may be better, but still no guarantee that it represents a product ID in all scenarios, could just be a random fragment.

dcaravel (commented):
Adding the fragment to links breaks existing deduping logic in [stackrox](https://github.com/stackrox/stackrox/blob/dc3516d5a4a641081eac1c349402ce5159c530f3/pkg/scannerv4/mappers/mappers.go#L1585)

Keeping the fragment in `links` would require a stackrox re-work and a `v3` vuln bundle to not introduce 'duplicate' vulns into existing deployments.

Given that fragments can appear naturally in URLs, a client may not be able to reliably extract the product ID without making assumptions (ie: product ID is a fragment appended to links for the RH VEX domain+path, which will need to be updated anytime that changes)

Is the placement of the product ID worth a revisit? perhaps a dedicated field or general metadata? Making the product ID available to clients in a relaible way could be a step towards making results explainable.

test/acceptance/auditor_claircore.go
crozzy: Yes, we/I should think about locking this down to the self-link to avoid this corner-case

crozzy:
> Adding the fragment to links breaks existing deduping logic in [stackrox](https://github.com/stackrox/stackrox/blob/dc3516d5a4a641081eac1c349402ce5159c530f3/pkg/scannerv4/mappers/mappers.go#L1585)
> 
> Keeping the fragment in `links` would require a stackrox re-work and a `v3` vuln bundle to not introduce 'duplicate' vulns into existing deployments.
> 
> Given that fragments can appear naturally in URLs, a client may not be able to reliably extract the product ID without making assumptions (ie: product ID is a fragment appended to links for the RH VEX domain+path, which will need to be updated anytime that changes)
> 
> Is the placement of the product ID worth a revisit? perhaps a dedicated field or general metadata? Making the product ID available to clients in a relaible way could be a step towards making results explainable.

@dcaravel I believe this subsequent PR should address these concerns: https://github.com/quay/claircore/pull/1918

dcaravel (approved):

jvdm (approved):

BradLugo (approved):

vulerh (approved):
Approved! thanks

PR #1908, opened 2026-06-09
osv: skip ECOSYSTEM ranges for Go and npm

jvdm:
The GitHub Advisory Database published advisories for the npm package `nocodb` using `ECOSYSTEM` range type instead of `SEMVER` (e.g., [GHSA-4w6r-5c2j-qf5f](https://github.com/advisories/GHSA-4w6r-5c2j-qf5f)). This happens because nocodb switched to calendar-based versioning (`2026.04.1`), which is not valid SemVer 2.0 — the advisory database chose `ECOSYSTEM` accordingly.

The parser currently returns a hard error for `ECOSYSTEM` ranges on Go and npm ecosystems, added as a defensive check in a607a05d and 87be0246. This error propagates through Parse, aborting the entire ecosystem update.

This change skips the range with a warning instead, using the same pattern as `GIT` ranges. The advisory is recorded via `stats.Ignored`.

### DIFF
--- a/updater/osv/osv.go
+++ b/updater/osv/osv.go
@@ -641,6 +641,16 @@ func (e *ecs) Insert(ctx context.Context, log *slog.Logger, skipped *stats, name
 			switch r.Type {
 			case `SEMVER`:
 			case `ECOSYSTEM`:
+				switch af.Package.Ecosystem {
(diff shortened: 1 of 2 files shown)


dcaravel (approved):
LGTM

PR #1912, opened 2026-06-11
osv: backport ECOSYSTEM skip to stable-1.5.28

jvdm:
Backport of #1908 to `stable-1.5.28` for the ACS 4.5.x v1 vulnerability bundle stream.

Adapted for the v1.5.28 API: uses `zlog` instead of `slog`, `skipped.Ignored` as a slice append instead of a method call, and removes the orphan vulnerability with `RemoveVulnerability` since the v1.5.28 code creates the vulnerability eagerly before processing ranges.

Test cases from #1908 that depend on upstream multi-range behavior (not present in v1.5.28) were excluded.

Context: [ROX-34975](https://redhat.atlassian.net/browse/ROX-34975)


crozzy:
Going to go ahead and close this @jvdm

PR #1913, opened 2026-06-11
osv: backport ECOSYSTEM skip to stable-1.5.44

jvdm:
Backport of #1908 to `stable-1.5.44` for the ACS 4.9.x v2 vulnerability bundle stream.

Adapted for the v1.5.44 API: uses `zlog` instead of `slog`, `skipped.Ignored` as a slice append instead of a method call.

Context: [ROX-34975](https://redhat.atlassian.net/browse/ROX-34975)


dcaravel (dismissed):

dcaravel (approved):
LGTM - IMO focus on this one first as `v2` is what is most impacting current users.

PR #1920, opened 2026-06-18
Upgrade github.com/doug-martin/goqu from v8 to v9

chimanjain:
Upgrade `github.com/doug-martin/goqu` from `v8` to `v9`.


chimanjain:
Can the reviewers these verify these changes.

PR #1921, opened 2026-06-18
rhcc: account for incorrectly double-quoted labels.json values

crozzy:
The Red Hat build system was writing labels.json files that had double quoted values which would cause false negatives during matching.

### DIFF
--- a/rhel/rhcc/detector.go
+++ b/rhel/rhcc/detector.go
@@ -121,6 +121,14 @@ func findLabelsJSON(sys fs.FS) (*labels, string, error) {
 			if err := json.Unmarshal(l, &lb); err != nil {
 				return nil, "", err
 			}
+			// A build tooling bug caused some images to be published with label
+			// values wrapped in an extra layer of JSON string quoting (e.g.
+			// `"\"rhacs-main-container\""` instead of `"rhacs-main-container"`).
+			for _, p := range []*string{&lb.Name, &lb.Architecture, &lb.CPE} {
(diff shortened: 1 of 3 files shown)


rhel/rhcc/detector.go
```diff
@@ -121,6 +121,14 @@ func findLabelsJSON(sys fs.FS) (*labels, string, error) {
...
 				return nil, "", err
 			}
+			// A build tooling bug caused some images to be published with label
+			// values wrapped in an extra layer of JSON string quoting (e.g.
+			// `"\"rhacs-main-container\""` instead of `"rhacs-main-container"`).
+			for _, p := range []*string{&lb.Name, &lb.Architecture, &lb.CPE} {
+				if u, err := strconv.Unquote(*p); err == nil {
+					*p = u
+				}
+			}
```
BradLugo, line 131: If we're going to do this, would it be worth unquoting all fields? perhaps a custom unmarshaler?

BradLugo (approved):
LGTM. I noted a possible improvement, but it's not necessary to resolve this issue.

dcaravel (approved):
LGTM

rhel/rhcc/detector.go
crozzy: I think the conclusion was to not add more machinery to this hack fix, LMK if I interpreted that correctly.

BradLugo (approved):

hdonnay (approved):

PR #1922, opened 2026-06-18
acceptance: add VEX parser option to include the productID

crozzy:
The VEX productID is useful for testing but can cause issues when included in production. This adds the parser option to be able to turn on adding the productID to the VEX file link for use during testing.


dcaravel (approved):
LGTM

nit: consider adding a unit test to `parser_test.go` for when the option is enabled.

jvdm:
> nit: consider adding a unit test to `parser_test.go` for when the option is enabled.

+1

hdonnay:
@dcaravel @jvdm @crozzy somewhat idle question: what would break (if anything) if one of the space-separated links weren't an `http` URI?

hdonnay:
What if the product ID became an [RFC4151](https://datatracker.ietf.org/doc/html/rfc4151) `tag`, e.g. `tag:clairproject.org,2026:updater/rhel-vex#9Base-RHOSE-4.16:cri-tools-0:1.29.0-3.1.el9.aarch64`. Alternatively, a `urn` scheme could encode similar information.

hdonnay (approved):

crozzy:
> What if the product ID became an [RFC4151](https://datatracker.ietf.org/doc/html/rfc4151) `tag`, e.g. `tag:clairproject.org,2026:updater/rhel-vex#9Base-RHOSE-4.16:cri-tools-0:1.29.0-3.1.el9.aarch64`. Alternatively, a `urn` scheme could encode similar information.

This might be nice, and we could keep the ability to turn it on and off

crozzy:
https://github.com/quay/claircore/issues/1924

PR #1941, opened 2026-07-06
language: deduplicate packages across layers 

BradLugo:
Since all language ecosystem coalescers shared the same logic - skip layers without repos, associate packages with repos - these changes refactor those coalescers to use a new, shared function. The gobin coalescer had unnecessary repo name/PackageDB prefix filtering since scanners already ensure only relevant packages appear.

When the same package appears at the same path across multiple layers, keep only the latest version. Duplicate entries at the same version are collapsed to the topmost layer.

Reimplements https://github.com/quay/claircore/pull/1275, which was reverted in https://github.com/quay/claircore/pull/1562.


hdonnay (approved):
I didn't scrutinize this too hard, seeing as it's ported from the `nodejs` implementation.

Great change to have this consolidated.

PR #1942, opened 2026-07-06
vex: warn and skip unparseable entries in knownAffectedVulnerabilities

jvdm:
`knownAffectedVulnerabilities` hard-fails on `PackageName()`, `Module()`, and `cvssVectorFromScore()` errors, while its sibling functions `fixedVulnerabilities` and `knownNotAffectedVulnerabilities` warn-and-skip for the same operations. This was introduced in f9fc989e during the iterator migration.

This change moves validation before allocation so `continue` doesn't leave orphaned entries in the backing rope, and aligns the error handling with the other two functions.

### DIFF
--- a/rhel/vex/parser.go
+++ b/rhel/vex/parser.go
@@ -708,25 +708,34 @@ func (r *rope[E]) All() iter.Seq[*E] {
 // KnownAffectedVulnerabilities processes the "known_affected" array of products
 // in the VEX object.
 func (c *creator) knownAffectedVulnerabilities(ctx context.Context, v *csaf.Vulnerability, init vulnHook) ([]*claircore.Vulnerability, error) {
+	log := slog.With("link", c.docLink)
 	var backing rope[claircore.Vulnerability]
 	for st, err := range c.Status(ctx, v, csaf.ProductStatusKnownAffected) {
 		if err != nil {
 			return nil, err
 		}
 
-		// This loop never skips returned [status] values, so we can always just
-		// append a new [claircore.Vulnerability].
-		vuln := backing.New()
-
-		if err := init(ctx, vuln); err != nil {
-			return nil, err
-		}
 		pkgName, err := st.PackageName()
 		if err != nil {
-			return nil, err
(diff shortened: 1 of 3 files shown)


rhel/vex/parser.go
```diff
@@ -708,38 +708,42 @@ func (r *rope[E]) All() iter.Seq[*E] {
...
+			log.WarnContext(ctx, "bad purl", "reason", err, "purl", st.PURL, "missing", "PackageName")
+			continue
 		}
 		modName, err := st.Module()
 		if err != nil {
+			log.WarnContext(ctx, "bad purl", "reason", err, "purl", st.PURL, "missing", "ModuleName")
+			continue
+		}
+		var sev string
+		if sc := st.Score; sc != nil {
```
jvdm, line 729: It feels we should skip if we don't find scores for "known affected".

crozzy:
I think we need to perform admin on the `backing` rope ala: https://github.com/quay/claircore/pull/1943/changes

jvdm:
> I think we need to perform admin on the `backing` rope ala: https://github.com/quay/claircore/pull/1943/changes

The fixedVulnerabilities and knownNotAffectedVulnerabilities use `Drop()` because they dedup, so allocation has to happen before validation there since the dedup lookup itself triggers `backing.New()`. But, this function doesn't dedup, so I moved allocation after validation instead of speculatively allocating and cleaning up. Happy to switch the `Drop()` if you'd rather keep the same across all three.

crozzy:
> > I think we need to perform admin on the `backing` rope ala: https://github.com/quay/claircore/pull/1943/changes
> 
> The fixedVulnerabilities and knownNotAffectedVulnerabilities use `Drop()` because they dedup, so allocation has to happen before validation there since the dedup lookup itself triggers `backing.New()`. But, this function doesn't dedup, so I moved allocation after validation instead of speculatively allocating and cleaning up. Happy to switch the `Drop()` if you'd rather keep the same across all three.

Oh my bad, I didn't see the instantiation moved

crozzy (approved):

PR #1945, opened 2026-07-10
debug: add a script for investigating a jar

hdonnay:
This script automates the process of extracting a jar from an OCI image and running the `java/jar` "TestJAR" test against it.

### DIFF
--- a/etc/debug/bin/investigate-jar
+++ b/etc/debug/bin/investigate-jar
@@ -1,0 +1,111 @@
+#!/usr/bin/env zsh
+set -euo pipefail
+
+printhelp() {
+	if (( ${#*} )); then
+		print ${(F)*}
+		print
+	fi
+	cat <<-. >&2
+	usage: investigate-jar [<flags>] IMAGE-REF
+
+	To run from the root of the repository, the invocation would be:
+		etc/debug/bin/investigate-jar --pkgdir java/jar IMAGE-REF
+
+	Flags:
+	  --ocidir:  Output directory for OCI image.                                         (default: temporary directory)
+	  --tmpdir:  Directory for temporary OCI image output.                               (default: /var/tmp)
+	  --pkgdir:  Directory containing the 'github.com/quay/claircore/java/jar' package.  (default: .)         
+
+.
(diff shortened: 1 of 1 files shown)


etc/debug/bin/investigate-jar
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/zsh
```
BradLugo:
```suggestion
#!/usr/bin/env zsh
```

etc/debug/bin/investigate-jar
```diff
@@ -0,0 +1,106 @@
...
+local -a jarlist
+local -A jarlayer
+for l in $layers; do
+	tar taf ${opts[--ocidir]}/blobs/${l/://} |
+	grep -E '\.[jwe]ar$' |
+	while read -r n; do
+		jarlist+=($n)
+		jarlayer[$n]=$l
+	done
+done
```
BradLugo, line 85: What happens when a jar is "overwritten" in another layer?

BradLugo (commented):
Thanks for taking the time to publish this! I love debug tooling.

One more consideration I wanted to note: what happens in the case of whiteout files? Maybe not something worth worrying about, but I was curious.

I'd really like the `env` change so I can use this on my NixOS machine 🙂

etc/debug/bin/investigate-jar
hdonnay: I think this will only allow you to select the earliest file with a given path. We could fix this if it ever comes up, or I can hack at it some more.

hdonnay:
> One more consideration I wanted to note: what happens in the case of whiteout files? Maybe not something worth worrying about, but I was curious.

Yeah, this doesn't contend with whiteouts at all. The idea is that a user of this script knows what to investigate; this just bundles it all up.

etc/debug/bin/investigate-jar
BradLugo: Maybe a comment/todo would suffice? I'll mark this as resolved since we're at least aware of it.

BradLugo (approved):

PR #1949, opened 2026-07-15
apk: guard against short lines in installed database

arpitjain099:
I do supply-chain security research, mostly on scanners and indexers.

The apk record loop in `apk/scanner.go` slices every line as `line[2:]`. A bare newline is a 1 byte line, so that's `slice bounds out of range [2:1]` and the scan panics.

You get one from a stray blank line inside a record. Splitting on `"\n\n"` leaves the extra newline sitting at the front of the next record, and its first `ReadBytes` hands back just `"\n"`. Layer contents come from whatever image is being indexed, so a malformed or crafted `lib/apk/db/installed` is enough to crash the indexer.

Fix is a length check before the slice. dpkg doesn't have this problem because textproto tolerates the junk for it, which the comment above the loop already explains apk can't use.

Test is `TestBlankLine`, a tarred-up installed database with an extra blank line, run through `Scan` the normal way. Panics on main, passes with the fix, and no network needed. `go test ./apk/` and `go test -race ./apk/` are clean.

### DIFF
--- a/apk/scanner.go
+++ b/apk/scanner.go
@@ -90,6 +90,11 @@ func (*Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircore.
 		}
 		r := bytes.NewBuffer(entry)
 		for line, err := r.ReadBytes('\n'); err == nil; line, err = r.ReadBytes('\n') {
+			// Every well-formed line is a one byte key, a ":", then the value.
+			// Anything shorter can't be one, so skip it.
+			if len(line) < 2 {
+				continue
+			}
 			l := string(bytes.TrimSpace(line[2:]))
 			switch line[0] {
 			case 'P':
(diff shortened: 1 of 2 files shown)


apk/scanner.go
```diff
@@ -90,6 +90,11 @@ func (*Scanner) Scan(ctx context.Context, layer *claircore.Layer) ([]*claircore.
 		}
 		r := bytes.NewBuffer(entry)
 		for line, err := r.ReadBytes('\n'); err == nil; line, err = r.ReadBytes('\n') {
+			// Every well-formed line is a one byte key, a ":", then the value.
+			// Anything shorter can't be one, so skip it.
+			if len(line) < 2 {
```
crozzy, line 95: NIT: To align with the comment, should this be `if len(line) < 3` eg. `<key><:><value><\n>`?

crozzy (approved):
One comment, non-blocking. Thanks for the patch

PR #1953, opened 2026-07-17
httpreader: random access over an HTTP resource

hdonnay:
This adds a package (`internal/httpreader`) that implements `io.ReaderAt` over HTTP requests. It incorporates a lot of tricks I know from a past life dealing with [RFC7233] non- and selectively- compliant servers.

Then, the `fetcher` package gains the capability to use an `httpreader.Reader` when it notices that a layer is uncompressed.

Future work may involve the `tarfs` layer doing transparent caching and being able to handle compressed layers directly.

[RFC7233]: https://datatracker.ietf.org/doc/html/rfc7233


hdonnay:
Conflicts with #1948

internal/httpreader/reader.go
```diff
@@ -0,0 +1,442 @@
...
+	"time"
+)
+
+// Reader implements [io.ReaderAt] over an HTTP resource that implements range requests.
+//
+// The Reader is guaranteed to only makes GET requests.
+//
+// Callers can use an [io.SectionReader] if they require an [io.Reader] interface.
+type Reader struct {
+	ctx     context.Context // This is OK, I swear.
```
BradLugo, line 23: 👀

internal/httpreader/content_range.rl
BradLugo, line 1: I want to approve this, because it's cool (and novel for me), but my maintainer brain is weary. I think at a minimum, I'd like to see documentation and benchmarks for this approach. I understand that's asking a lot, but there isn't even community support for `ragel`/`rl` in GoLand.

BradLugo: We discussed this on Slack a bit more. Looks like we already ragel in the codebase anyway (for `toolkit/types/cvss`). Because of that, I think this is much less of a concern, but we should improve the tooling and documentation around our use of ragel.

internal/httpreader/reader.go
```diff
@@ -0,0 +1,442 @@
...
+	c       *http.Client
+	done    context.CancelFunc
+	res     string
+	headers http.Header
+
+	validator string
+	modTime   time.Time
+	size      int64 // Discovered size of the resource "res."
+	guessed   bool  // Did the rangefind method guess at the size?
```
BradLugo: Looks like this is unused in 87abab5. Not really a concern for me, but I know you like to keep commits pretty tidy.

internal/httpreader/metrics.go
BradLugo, line 1: Note for anyone coming across this in the future: this generally looks correct to me, but I haven't tested it.

internal/httpreader/ragel.sh
BradLugo, line 1: Is this wired up anywhere?

internal/httpreader/reader.go
```diff
@@ -0,0 +1,442 @@
...
+	switch res.StatusCode {
+	case http.StatusPartialContent: // OK
+	case http.StatusOK:
+		if off != 0 {
+			return 0, fmt.Errorf("httpreader: server botch: %d (range: %s)", res.StatusCode, req.Header.Get(`range`))
+		}
+	case http.StatusRequestedRangeNotSatisfiable:
+		return 0, fmt.Errorf("httpreader: server botch: %d (range: %s)", res.StatusCode, req.Header.Get(`range`))
+	default:
+		return 0, fmt.Errorf("httpreader: unexpected status: %d", res.StatusCode)
```
BradLugo: I'm seeing a common pattern with these errors. Should we create a type to, for example, include the status code as a field?

internal/httpreader/reader.go
```diff
@@ -0,0 +1,442 @@
...
+	u, err := url.Parse(s)
+	if err != nil {
+		panic(fmt.Sprintf("programmer error: bogus URL: %v", err))
+	}
+	return &errRangeUnsupported{u: u}
+}
+
+var _ error = (*errRangeUnsupported)(nil)
+
+type errRangeUnsupported struct{ u *url.URL }
```
BradLugo: Should we export this?

internal/httpreader/reader.go
```diff
@@ -0,0 +1,442 @@
...
+	// Alternatively, the window size could be shrunk, but that means more rangefinding requests in the loop above.
+	// Shrinking the window size to 100 MiB still results in a worst-case 5.3 requests here.
+	// Seeing as one additional rangefinding request negates any advantage, the current magic numbers seem OK.
+	const searchWin = 1024
+	lowerBucket, upperBucket := lower/searchWin, upper/searchWin
+	nBuckets := upperBucket - lowerBucket
+	if (upper-lower)%searchWin != 0 {
+		nBuckets++
+	}
+	ceil := int(math.Ceil(math.Log(float64(nBuckets))))
```
BradLugo: nit: the math in your comment uses log10, whereas Go's `math.Log` is ln.
```suggestion
	ceil := int(math.Ceil(math.Log10(float64(nBuckets))))
```
then again, I might use Log2 since it looks like we're trying to do a binary search:
```suggestion
	ceil := int(math.Ceil(math.Log2(float64(nBuckets))))
```
Either way, it's `O(log(n))`. I don't think it affects the search in a meaningful way (hence the nit), but might want double check that.

internal/httpreader/reader.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+	defer res.Body.Close()
+	switch res.StatusCode {
+	case http.StatusPartialContent: // OK
+	case http.StatusOK:
+		if off != 0 {
+			return 0, fmt.Errorf("httpreader: server botch: %d (range: %s)", res.StatusCode, req.Header.Get(`range`))
+		}
+	case http.StatusRequestedRangeNotSatisfiable:
+		return 0, fmt.Errorf("httpreader: server botch: %d (range: %s)", res.StatusCode, req.Header.Get(`range`))
```
BradLugo: What happens when `off > r.size`?

internal/httpreader/reader.go
```diff
@@ -0,0 +1,442 @@
...
+		}
+		// Check if we manged to request the actual end:
+		if (rg.Last+1)-rg.First < searchWin {
+			r.size = rg.Last + 1
+			break
+		}
+		lower = pivot
+	}
+	if reqs > lim {
+		return errors.New("loop")
```
BradLugo: nit: improve error (at least `httpreader: ...`)

pkg/tarfs/testdata/paxsize.tar
BradLugo, line 1: Does this need to be committed?

pkg/tarfs/tarfs_test.go
```diff
@@ -618,3 +618,27 @@ func TestInvalidName(t *testing.T) {
 		}
 	}
 }
+
+//go:generate sh testdata/create_paxsize.sh
+func TestPAXSize(t *testing.T) {
```
BradLugo, line 623: Looks like this test fails without 5d4f550. Again, not really a blocker for me, mostly fyi.

pkg/tarfs/testdata/paxsize.tar
BradLugo: I suppose there isn't much harm, and systems that don't have `sh` will be able to run the test.

pkg/tarfs/parse.go
```diff
@@ -112,20 +117,60 @@ Scan:
...
+			for s.Scan() {
+				fs := strings.FieldsFunc(s.Text(), func(r rune) bool { return r == ' ' || r == '=' })
+				if fs[1] == "size" {
+					// This is *not* the stringified octal madness of a real header.
+					*p, err = strconv.ParseInt(fs[2], 10, 64)
+					if err != nil {
+						return nil, parseErr("bad block at %d: weird PAX header: bad size: %v", off, err)
+					}
+				}
+			}
```
BradLugo, line 168: I've never really looked at the pax spec before, but after briefly reading about it, it seems like we'd at least need to break after finding `size` so we don't hit one of the extended headers, which might cause a an error: https://pubs.opengroup.org/onlinepubs/009695399/utilities/pax.html#tag_04_100_13_03

libindex/fetcher.go
```diff
@@ -173,12 +215,106 @@ func (f closeFunc) Close() error {
...
+
+	var req *http.Request
+	var res *http.Response
+	req, err = http.NewRequestWithContext(ctx, http.MethodGet, desc.URI, nil)
+	req.Header = http.Header(desc.Headers).Clone()
+	req.Header.Set(`claircore-reason`, `inspect`)
+	req.Header.Set(`range`, `bytes=0-15`)
+	if err != nil {
+		return d, fmt.Errorf("fetcher: failed to construct request: %w", err)
+	}
```
BradLugo:
```suggestion
	if err != nil {
		return d, fmt.Errorf("fetcher: failed to construct request: %w", err)
	}
	req.Header = http.Header(desc.Headers).Clone()
	req.Header.Set(`claircore-reason`, `inspect`)
	req.Header.Set(`range`, `bytes=0-15`)
```

BradLugo (changes requested):
I think there's enough to address for now. I'll do another deep dive in the next review

internal/httpreader/ragel.sh
BradLugo: It is (`//go:generate sh -e ragel.sh`)

internal/httpreader/reader.go
BradLugo: Yeah, after the full review, it seems that this field isn't used at all

internal/httpreader/reader.go
hdonnay: I was assuming a user would just `errors.Is(err, errors.ErrUnsupported)`. It's also an internal package, so exporting it should be fine.

internal/httpreader/reader.go
hdonnay: Yeah, it's been a long time since I wrote that comment, so thanks for working through it.

internal/httpreader/reader.go
hdonnay: server should return `416 Range Not Satisfiable`

pkg/tarfs/parse.go
hdonnay: I'm not sure what you mean. This is handling the extended headers.

internal/httpreader/reader.go
BradLugo: That seems like a valid case (as opposed to an error case), i.e., `return 0, io.EOF`

hdonnay: let me quickly check what the `os.File` behavior is

hdonnay: OK, updated behavior around reading past the limit of a resource.

internal/httpreader/content_range.rl
```diff
@@ -0,0 +1,82 @@
...
+	main := 'bytes ' ( byte_range_resp | unsatisfied_range );
+
+	write data;
+	write init;
+	write exec;
+}%%
+	if cs >= content_range_start && cs < content_range_first_final {
+		// Expect that err is populated
+		return err
+	}
```
BradLugo: If `cs < content_range_first_final` is true, what would `err` be? It looks to me like it'd be `nil`, but I'm still trying to wrap my head around rl.

internal/httpreader/reader.go
```diff
@@ -0,0 +1,481 @@
...
+			return nil, err
+		}
+	}
+	if err := r.rangefind(); err != nil {
+		return nil, err
+	}
+
+	_, file, line, _ := runtime.Caller(1)
+	runtime.SetFinalizer(&r, func(r *Reader) {
+		panic(fmt.Sprintf("%s:%d: httpreader.Reader not closed", file, line))
```
BradLugo, line 69: Why panic instead of closing?

internal/httpreader/content_range.rl
```diff
@@ -0,0 +1,82 @@
...
+// Reset sets all fields to a known value (-1).
+func (r *ContentRange) Reset() {
+	r.First = -1
+	r.Last = -1
+	r.Length = -1
+}
+
+// Parse populates the receiver with the "bytes" content range from the
+// supplied header value or reports an error.
+func (r *ContentRange) Parse(data string) ( error) {
```
BradLugo: nit:
```suggestion
func (r *ContentRange) Parse(data string) error {
```

pkg/tarfs/parse.go
```diff
@@ -112,20 +117,60 @@ Scan:
...
+			if err := s.Err(); err != nil {
+				return nil, parseErr("bad block at %d: weird PAX header: %v", off, err)
+			}
+			if typ == tar.TypeXGlobalHeader {
+				// Make sure to immediately "promote" the global size.
+				// Without this, the global size would only take effect after
+				// the next regular file.
+				xSize = gSize
+			}
```
BradLugo: I think there might be a really subtle issue here. Since we aren't resetting `cur` for `tar.TypeXGlobalHeader` anymore, I think we would skip the header we're actually looking for (after the global header): https://github.com/golang/go/blob/98e6631a5aa274aadc7f5d8c390057fca50368fb/src/archive/tar/reader.go#L106-L114

BradLugo: I think there might be a really subtle issue here. Since we aren't resetting `cur` for `tar.TypeXGlobalHeader` anymore, I think we would skip the header we're actually looking for (after the global header): https://github.com/golang/go/blob/98e6631a5aa274aadc7f5d8c390057fca50368fb/src/archive/tar/reader.go#L106-L114

pkg/tarfs/parse.go
```diff
@@ -112,20 +117,60 @@ Scan:
...
+			// This is just the "file" data, which is why this is different from
+			// the [segment] math below.
+			sr := io.NewSectionReader(r, (blk-nBlk)*blockSz, sz)
+			s := bufio.NewScanner(sr)
+			for s.Scan() {
+				fs := strings.FieldsFunc(s.Text(), func(r rune) bool { return r == ' ' || r == '=' })
+				if fs[1] == "size" {
+					// This is *not* the stringified octal madness of a real header.
+					*p, err = strconv.ParseInt(fs[2], 10, 64)
```
BradLugo: Is it worth checking for negative values?

libindex/fetcher.go
```diff
@@ -173,12 +215,106 @@ func (f closeFunc) Close() error {
...
+	}()
+	span.SetStatus(codes.Error, "")
+
+	var req *http.Request
+	var res *http.Response
+	req, err = http.NewRequestWithContext(ctx, http.MethodGet, desc.URI, nil)
+	if err != nil {
+		return d, fmt.Errorf("fetcher: failed to construct request: %w", err)
+	}
+	req.Header = http.Header(desc.Headers).Clone()
```
BradLugo, line 245: Could `desc.Headers` be `nil`?

libindex/fetcher.go
```diff
@@ -125,6 +129,43 @@ func (a *RemoteFetchArena) fetchInto(ctx context.Context, l *claircore.Layer, cl
...
+			}
+			if len(desc.Headers) != 0 {
+				opts = append(opts, httpreader.WithHeaders(desc.Headers))
+			}
+			var rd *httpreader.Reader
+			rd, err = httpreader.New(ctx, a.wc, desc.URI, opts...)
+			switch {
+			case err == nil:
+				if err = l.Init(ctx, desc, rd); err != nil {
+					return err
```
BradLugo: I think `rd` needs to be closed here.

libindex/fetcher.go
```diff
@@ -125,6 +129,43 @@ func (a *RemoteFetchArena) fetchInto(ctx context.Context, l *claircore.Layer, cl
 			span.End()
 		}()
 
+		var d details
+		d, err = a.inspect(ctx, desc)
+		if err != nil {
+			return err
+		}
```
BradLugo: Should we fall back to spooling to disk instead?

libindex/fetcher.go
```diff
@@ -173,12 +215,106 @@ func (f closeFunc) Close() error {
...
+	_ = res.Body.Close()
+	if err != nil {
+		return d, fmt.Errorf("fetcher: unexpected read: %w", err)
+	}
+
+	const (
+		ctKey = `content-type`
+		arKey = `accept-ranges`
+	)
+	ct, _, err := mime.ParseMediaType(res.Header.Get(ctKey))
```
BradLugo: What happens if the mim type is empty?

libindex/fetcher.go
```diff
@@ -125,6 +129,43 @@ func (a *RemoteFetchArena) fetchInto(ctx context.Context, l *claircore.Layer, cl
...
 
+		var d details
+		d, err = a.inspect(ctx, desc)
+		if err != nil {
+			return err
+		}
+
+	HTTPReader:
+		switch { // Switch for a single case to be able to break.
+		case d.Uncompressed:
```
BradLugo: Should this be `d.Uncompressed: && d.RangeOK`?

BradLugo (commented):
More questions

pkg/tarfs/parse.go
BradLugo: It seems that we should parse `size` first, then read `size` bytes, instead of using `Scan()`. Otherwise, we might read junk (I think), like too few bytes because we reached the limit of `Scan()` or because we hit a newline, etc

internal/httpreader/reader.go
BradLugo: ah, didn't see it was internal. I don't know what the implications are for exporting types that are internal. I'll defer to your judgment

internal/httpreader/reader.go
hdonnay: I like [sharp-edged finalizers](https://crawshaw.io/blog/sharp-edged-finalizers).

internal/httpreader/content_range.rl
hdonnay: This means that the machine is in some nonfinal state. We had to have gotten here by an `fbreak` ("Advance `p`, save the target state to `cs` and immediately break out of the execute loop."), which only occurs in the error paths of the machine's actions.

pkg/tarfs/parse.go
hdonnay: I don't see it. Assume a tar stream like:

- global (size = 0)
- extended (size = n)
- reg "A"
  - file data
- reg "B"
- EOF

Then we should emit two sections:

- [ global, extended, reg "A" (size = n) ]
- [ reg "B" (size = 0) ]

Will the the stdlib tar decoder freak out because of the global size? Maybe, but I don't see how we skip a "regular" header.

I think the alternative is to ignore the global headers and say they don't make sense for setting the size, which is all our code cares about.

pkg/tarfs/parse.go
hdonnay: I don't think so, because a negative length doesn't mean anything. The stdlib implementation also doesn't check that the `ParseInt` return is >= 0.

libindex/fetcher.go
hdonnay: Yes. The Clone method is documented to handle `nil` by returning `nil`.

internal/httpreader/content_range.rl
BradLugo: > We had to have gotten here by an `fbreak`

I think that's the case only with `write exec noend`, but with `write exec`, it seems we can land in a nonfinal state without having called `fbreak`, i.e., the machine ran out of input. If I understand the manual correctly, I think we can explicitly check for an error with `content_range_err`.

libindex/fetcher.go
BradLugo: Yeah, I think that's going to be a problem in the lines that follow (sry, I should've highlighted all of them)
```go
	req.Header.Set(`claircore-reason`, `inspect`)
	req.Header.Set(`range`, `bytes=0-15`)
```

internal/httpreader/content_range.rl
hdonnay: Okay. I don't recall where I saw this pattern, but I can just have the actions return out of the function directly and then we'll be sure of how it works.

libindex/fetcher.go
hdonnay: This should report an error: https://go.dev/play/p/uFtpEXlJrQO

hdonnay: Will add a fallback

libindex/fetcher.go
hdonnay: I think any failure of the `inspect` routine means the spooling would fail, also.

pkg/tarfs/parse.go
BradLugo: According to the `findSegments` function doc:
> Each returned segment describes a region that is not a complete tar file, but can have exactly one file read from it.

Which, after finding the usage of `findSegments`, in my own words, seems to be: "each returned segment will have exactly one header that describes the file": 
https://github.com/quay/claircore/blob/6b23c3f53e46ab997e974eb54cc8a1d863f45de2/pkg/tarfs/tarfs.go#L114

i.e., in the case of `[ global, extended, reg "A" (size = n) ]`, I think only `global` will be read.

hdonnay: No, the stdlib `tar.Reader` handles the extra headers transparently

BradLugo: I think there's an exception for `TypeXGlobalHeader` (see link above)

hdonnay: Yeah, okay, after carefully re-reading the go package documentation and spelunking with a debugger, you're right.

The go stdlib package does not handle global headers, so we don't have to either. That'll make this simpler.

libindex/fetcher.go
BradLugo: I'm not so sure, e.g., I can see some weirdness happening where a server/proxy/WAF returns a non-200 when it receives a request with the `Range` header or something. In any case, after thinking about it for a bit, should we move the `inspect` after we check the cache anyway?

hdonnay: Reorganized it so the `inspect` and `fetchFileForCache` calls both happen only on the "cache miss" path.

pkg/tarfs/parse.go
hdonnay: This is using a `SectionReader` to effectively "fence" the `Scanner` to just the `size` bytes of the extended header records. The Scanner is nice because it takes care of short reads as long as the line fits in the buffer, which is 4KiB by default.

internal/httpreader/reader.go
BradLugo: I'm beginning to wonder if this is, in fact, OK. Hear me out:
So we use the reader here: https://github.com/hdonnay/claircore/blob/ba37694ca0ba2fdd1541dc11b4542ddc0622c4a1/libindex/fetcher.go#L187-L192
Which contains this context: https://github.com/hdonnay/claircore/blob/ba37694ca0ba2fdd1541dc11b4542ddc0622c4a1/libindex/fetcher.go#L533
Which gets waited on: https://github.com/hdonnay/claircore/blob/ba37694ca0ba2fdd1541dc11b4542ddc0622c4a1/libindex/fetcher.go#L542-L556
I _think_ the kicker here is that the wait will cancel the context, which will cause the `claircore.Layer` read methods that the scanners/detectors use to return a context canceled error when they're first called.

Assuming my thinking is correct, I don't have any good suggestions for fixing this off the top of my head (without some major rewriting/refactoring). _Maybe_ `context.WithoutCancel` would be _fine_?

hdonnay: Yeah, that sounds right. The `httpreader`'s use of the context is different from the fetch flow. I think a `WithoutCancel` is the correct thing to do, here.

pkg/tarfs/parse.go
```diff
@@ -112,27 +116,79 @@ Scan:
...
+			// This code lets the [bufio.Scanner] take care of line length
+			// considerations and ignores the "length" record.
+			s := bufio.NewScanner(sr)
+		Read:
+			for s.Scan() {
+				fs := strings.FieldsFunc(s.Text(), func(r rune) bool { return r == ' ' || r == '=' })
+				switch {
+				case len(fs) != 3:
+					// Malformed record, skip.
+				case fs[1] == "size" && fs[2] == "":
```
BradLugo: I think given the split condition `r == ' ' || r == '='`, `fs[2] == ""` won't be possible

libindex/fetcher.go
BradLugo: I still think there may be edge cases where the inspect would fail, but the GET from the spooling would succeed (e.g., a weirdly configured proxy/WAF that blocks Range but would succeed on a regular GET). I think falling back to spooling would be ideal, but I won't block on it. Feel free to resolve this thread and merge.

BradLugo (approved):
Non-blocking recommendation: https://github.com/quay/claircore/pull/1953#discussion_r3857488968

PR #1958, opened 2026-07-20
osv: exclude advisories from non applicable ecosystems

crozzy:
OSV will include advisories in files that cover more ecosystems than the one requested. This could lead to incorrect parsing logic being used to parse advisories resulting in false-positives. This patch avoids parsing advisories that don't match the requested ecosystem.

### DIFF
--- a/datastore/postgres/migrations/matcher/20-force-osv-ecosystem-filter.sql
+++ b/datastore/postgres/migrations/matcher/20-force-osv-ecosystem-filter.sql
@@ -2,13 +2,13 @@
 -- affected entries to the dump's ecosystem. Multi-ecosystem advisories
 -- previously produced cross-ecosystem vulnerability rows (e.g. npm ranges
 -- under osv/pypi).
-DELETE FROM update_operation
+--
+-- Clear fingerprints rather than deleting update_operations so matching keeps
+-- serving the existing (possibly imperfect) set until the next successful
+-- update, instead of a vulns → empty → vulns gap. Stale cross-ecosystem rows
+-- drop off once the new update_operation becomes latest; GC removes them later.
+UPDATE update_operation
+SET
+  fingerprint = ''
 WHERE
   updater LIKE 'osv/%';
-
-DELETE FROM vuln v1 USING vuln v2
-LEFT JOIN uo_vuln uvl ON v2.id = uvl.vuln
-WHERE
-  uvl.vuln IS NULL
-  AND v2.updater LIKE 'osv/%'
-  AND v1.id = v2.id;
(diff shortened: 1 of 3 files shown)


crozzy:
Before: 717,975 vulns, 613MB
After: 707,034 vulns, 598MB

Better than nothing

hdonnay (dismissed):

datastore/postgres/migrations/matcher/20-force-osv-ecosystem-filter.sql
```diff
@@ -0,0 +1,14 @@
+-- Force per-ecosystem OSV updaters to re-fetch and re-parse after filtering
+-- affected entries to the dump's ecosystem. Multi-ecosystem advisories
+-- previously produced cross-ecosystem vulnerability rows (e.g. npm ranges
+-- under osv/pypi).
+DELETE FROM update_operation
+WHERE
+  updater LIKE 'osv/%';
```
dcaravel: Thinking about how this would impact users - after this migration before OSV data reloaded vuln reports will not contain OSV records. 

Users will see (IIUC): lots of vulns, upgrade, on re-scan see zero vulns, OSV updater finishes, next re-scan see lots of vulns again - that may be disruptive.

Is there a way to change this so that matching will still succeed with the existing data, and only after the next update finishes results impacted?

Perhaps:
```sql
 UPDATE update_operation
  SET fingerprint = ''
  WHERE updater LIKE 'osv/%';
```

crozzy: What do you mean by re-scan? With this there probably will be a period in minutes where the OSV updater is repopulating the data but it shouldn't take a rescan/reindex.

We could fiddle with the fingerprint like we previously did with the pypi migration though, the implication being that we have more data hanging around until a later update event (given claircore keeps a minimum of 2 versions).

dcaravel: > What do you mean by re-scan?

Index + Match (or match only if index still valid)

> With this there probably will be a period in minutes where the OSV updater is repopulating the data but it shouldn't take a rescan/reindex.

That's a good point, the loading is short and window is small, which will limit impact. 

Images are re-scanned on Central startup (StackRox), on upgrade the subset of images that are re-scanned immediately will be impacted, also many users will have the steady stream of scan requests coming from CI.

> We could fiddle with the fingerprint like we previously did with the pypi migration though, the implication being that we have more data hanging around until a later update event (given claircore keeps a minimum of 2 versions).

Will the amount of data or how long it hangs around be different than normal?

crozzy: Good point, probably not, the only difference is all the osv updaters will create those rows at the same time as opposed to on a natural cycle. Anyway, I updated with your suggestion, I think it's better.

dcaravel (approved):

PR #1975, opened 2026-07-31
vex: allow FixedInVersion rewrite via CEL

crozzy:
This patch allows callers to rewrite FixedInVersions with a CEL expression. Expressions have access to all the PURL data with everything in a StringType except for the qualifiers which are MapType.


rhel/vex/fixed_in_cel.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// CelExprFingerprintDigest returns a short hex digest of expr for fingerprinting.
+// Empty expr returns an empty string.
+func celExprFingerprintDigest(expr string) string {
+	if expr == "" {
+		return ""
+	}
+	sum := sha256.Sum256([]byte(expr))
+	return hex.EncodeToString(sum[:8])
```
hdonnay: Why truncate the shasum?

crozzy: 64 bits / 8 bytes seemed enough to signal the expression had changed and for something that is non-security related but full sha-256 is also fine, I can change that

hdonnay: I think I'd rather use [xxhash](https://pkg.go.dev/github.com/cespare/xxhash/v2), which is purpose built for this sort of thing.

hdonnay (changes requested):

rhel/vex/fixed_in_cel.go
crozzy: Good idea

hdonnay (approved):
LGTM

PR #1979, opened 2026-08-10
cpe: unbind fs optimizations

hdonnay:
Some memory profiling revealed that in large imports, this function is doing a lot of allocation just to do the unbinding (meaning the underlying `cpe.WFN` value is not in that bookkeeping).

These changes include:
- some cleanup
- benchmarking
- a new, generated parser
- switching to the new parser

The results locally:
```
goos: linux
goarch: amd64
pkg: github.com/quay/claircore/toolkit/types/cpe
BenchmarkUnbindFS/Simple-16         	 1978447	       655.0 ns/op	      32 B/op	       3 allocs/op
BenchmarkUnbindFS/Escape-16         	 1889960	       641.0 ns/op	      24 B/op	       2 allocs/op
BenchmarkUnmarshalFS/Simple-16      	 3437407	       361.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkUnmarshalFS/Escape-16      	 2744743	       455.2 ns/op	      16 B/op	       1 allocs/op
PASS
ok  	github.com/quay/claircore/toolkit/types/cpe	6.532s
```


toolkit/types/cpe/unbind.go
```diff
@@ -153,21 +155,41 @@ var valueURI = strings.NewReplacer(
...
+			esc = true
+			continue
+		case r == ':':
+			if esc {
+				break
+			}
+			wfn.Attr[a].unbindFS(&b, s[prev:i])
+			a++
+			if a == NumAttr {
+				return wfn, fmt.Errorf("cpe: malformed CPE formatted string: bad components: >13")
```
alkmim: I might be missing something but I feel that we will not catch the <13 case.

hdonnay: That's done in the `wfn.Valid()` method.

alkmim: Can you show me where? I can see Valid() protecting against all fields being unset but not only some of them.

alkmim: Maybe a test case calling UnbindFS directly with an input that is too short would be a good idea. It will also protect against future changes.

hdonnay: After walking through the code with @alkmim, I was right for the wrong reason: some code motion lost the "too short" check that used to exist, but the ragel machine does do an equivalent check.

I updated the PR to have the `UnbindFS` implementation removed entirely in the last commit.

alkmim (dismissed):

toolkit/types/cpe/unbind.go
alkmim: Nice! Thanks for the walkthrough.

toolkit/types/cpe/unbind_fs.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+
+		}
+	_out:
+		{
+
+		}
+
+	}
+	return nil
```
crozzy, line 190: I'm not a great ragel-ite but if `UnbindFS()` is being changed to utilize this, should this also return `w.Valid()`, like the old `UnbindFS()` did (or I guess in `UnbindFS()` directly) so callers get the same behoviour?

crozzy (commented):
LGTM, one comment

toolkit/types/cpe/unbind_fs.go
hdonnay: All these validations happen in the parser, now. For example, the position of the wildcards is encoded in the generated machine rather than checked after the fact.

crozzy: I think that's true for the wildcards but maybe not for the quoted hyphen condition
```go
if s == "\\-" {
		return fmt.Errorf("cpe: quoted hyphen MUST NOT be used by itself")
	}
```

toolkit/types/cpe/wfn.go
crozzy: This comment should also probably be updated

crozzy (approved):
LGTM

PR #1983, opened 2026-08-13
cpe: bindfs optimisations

hdonnay:
This implements `encoding.AppendText` and then implements binding a WFN to a string on top of it.

Benchmarks:
```
% go test -run none -bench 'Bind|Append' 
goos: linux
goarch: amd64
pkg: github.com/quay/claircore/toolkit/types/cpe
BenchmarkBindFS/Simple-16 	 2466998	       495.0 ns/op	     280 B/op	       6 allocs/op
BenchmarkBindFS/Escape-16 	 2515683	       491.7 ns/op	     280 B/op	       6 allocs/op
BenchmarkAppendText/Simple-16         	 3454317	       348.1 ns/op	       0 B/op	       0 allocs/op
BenchmarkAppendText/Escape-16         	 2717823	       453.5 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/quay/claircore/toolkit/types/cpe	4.898s
```

### DIFF
--- a/toolkit/types/cpe/bind.go
+++ b/toolkit/types/cpe/bind.go
@@ -3,41 +3,22 @@ package cpe
 import (
 	"encoding"
 	"errors"
-	"strings"
 	"unsafe"
 )
 
 // BindFS returns the WFN bound as CPE 2.3 formatted string.
+//
+// Deprecated: use [WFN.AppendText].
+//
+//go:fix inline
 func (w WFN) BindFS() string {
-	b := strings.Builder{}
-	b.WriteString(`cpe:2.3`)
-	for i := range NumAttr {
-		b.WriteByte(':')
-		w.Attr[i].bind(&b)
+	b := make([]byte, 0, 64)
+	b, err := w.AppendText(b)
+	if err != nil {
+		panic(err)
 	}
-	return b.String()
+	return string(b)
 }
-
-// Bind binds the value to a formatted string, writing it into the provided
-// strings.Builder.
-func (v *Value) bind(b *strings.Builder) (err error) {
-	switch v.Kind {
-	case ValueUnset, ValueAny:
-		_, err = b.WriteRune('*')
-	case ValueNA:
-		_, err = b.WriteRune('-')
-	case ValueSet:
-		_, err = valueString.WriteString(b, v.V)
-	}
-	return err
-}
-
-// ValueString does FS character replacing.
-var valueString = strings.NewReplacer(
-	`\.`, `.`,
-	`\-`, `-`,
-	`\_`, `_`,
-)
 
 var _ encoding.TextAppender = (*WFN)(nil)
 
--- a/toolkit/types/cpe/bind.go
+++ b/toolkit/types/cpe/bind.go
@@ -1,6 +1,11 @@
 package cpe
 
-import "strings"
+import (
+	"encoding"
+	"errors"
+	"strings"
+	"unsafe"
+)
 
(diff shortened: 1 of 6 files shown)


toolkit/types/cpe/bind.go
```diff
@@ -1,35 +1,75 @@
...
+	_ encoding.TextAppender = (*Value)(nil)
+	_ encoding.TextAppender = (*WFN)(nil)
+)
+
+// AppendText implements [encoding.TextAppender].
+func (w *WFN) AppendText(b []byte) ([]byte, error) {
+	switch err := w.Valid(); {
+	case err == nil:
+	case errors.Is(err, ErrUnset):
+		return []byte{}, nil
```
crozzy: Nit: Should this return `b, nil`?

toolkit/types/cpe/bind.go
```diff
@@ -1,35 +1,75 @@
...
 	}
-	return b.String()
+	return b, nil
 }
 
-// Bind binds the value to a formatted string, writing it into the provided
-// strings.Builder.
-func (v *Value) bind(b *strings.Builder) (err error) {
+// AppendText implements [encoding.TextAppender].
+func (v *Value) AppendText(b []byte) ([]byte, error) {
```
crozzy: What is the advantage of this method implementing `encoding.TextAppender`, or being public?

toolkit/types/cpe/bind.go
```diff
@@ -1,35 +1,75 @@
...
+// Deprecated: use [WFN.AppendText].
+//
+//go:fix inline
 func (w WFN) BindFS() string {
-	b := strings.Builder{}
-	b.WriteString(`cpe:2.3`)
+	b := make([]byte, 0, 64)
+	b, err := w.AppendText(b)
+	if err != nil {
+		panic(err)
```
crozzy, line 18: What is the point in panicking here? Is the idea that we won't call it anymore? If so should we change the usage within this package `/toolkit/types/cpe/marshaling.go`? I'm guessing there will be a subsequent claircore PP once the PR lands and the toolkit tag is cut to update non-toolkit `BindFS()` usage?

hdonnay: There's no error return, so I'm not sure what else to do with a returned error. Previously, this method would just construct an invalid bound WFN and the `MarshalText` method checked validity and then called this.

With the `//go:fix inline` annotation, a `go fix` call will replace uses for us; the `panic` call will be obviously wrong and easy to update as needed.

The use of this method is removed in change `Icc05f1b5ebaae2f6ac6bc218dfd8a76b6a6a6964`.

toolkit/types/cpe/bind.go
hdonnay: The value of this implementing `encoding.TextAppender` is that it all works on the "appender" pattern, then.

I have it exported because this type also implements `fmt.Stringer`, but I suppose these methods do subtly different things, then. I think it should probably not be exported.

toolkit/types/cpe/bind.go
crozzy: So, just so I'm understanding:
* //go:fix inline is going to modify the code in the calling sites
* Another PR will address all those panics at the calling sites to see what is appropriate
* That will need to be done after the toolkit tag
?

hdonnay: Yeah: `go fix` can be used to update callers. Those changes will have to deal with an `error` being returned. This needs to be done after the toolkit module is released with this function.

crozzy (approved):

PR #1984, opened 2026-08-13
postgres: some perf work

hdonnay:
This is a stab at some memory optimizations that jumped out. This now uses less than half the memory for calculating a hash:
```
% go test -run none -bench Hash\|MD5        
goos: linux
goarch: amd64
pkg: github.com/quay/claircore/datastore/postgres
BenchmarkMD5Vuln/Baseline-16     	  886275	      1622 ns/op	     648 B/op	      10 allocs/op
BenchmarkMD5Vuln/Current-16      	 1000000	      1195 ns/op	     232 B/op	       7 allocs/op
BenchmarkVulnerabilityHash/MD5-16         	 1000000	      1259 ns/op	     232 B/op	       7 allocs/op
BenchmarkVulnerabilityHash/XXH64-16       	 1379953	       863.0 ns/op	     224 B/op	       7 allocs/op
PASS
ok  	github.com/quay/claircore/datastore/postgres	5.093s
```

### DIFF
--- a/datastore/postgres/bench_test.go
+++ b/datastore/postgres/bench_test.go
@@ -1,0 +1,86 @@
+package postgres
+
+import (
+	"bytes"
+	"crypto/md5"
+	"math/rand"
+	"reflect"
+	"testing"
+	"testing/quick"
+
+	"github.com/quay/claircore"
+	"github.com/quay/claircore/internal/wart"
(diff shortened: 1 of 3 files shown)


crozzy (approved):
SGTM

PR #1986, opened 2026-08-14
libvuln: offline import performance improvements

hdonnay:
This uses iterator interfaces to avoid buffering entire UpdateOperations in memory.

### DIFF
--- a/libvuln/jsonblob/jsonblob.go
+++ b/libvuln/jsonblob/jsonblob.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"iter"
 	"os"
 	"sort"
 	"sync"
@@ -45,7 +46,17 @@ type Store struct {
 }
 
 // Load reads in all the records serialized in the provided [io.Reader].
+//
+// Deprecated: This just calls [NewLoader].
+//
+//go:fix inline
 func Load(ctx context.Context, r io.Reader) (*Loader, error) {
+	return NewLoader(ctx, r)
+}
+
+// NewLoader returns a loader configured to read the records serialized in the
+// provided [io.Reader].
+func NewLoader(ctx context.Context, r io.Reader) (*Loader, error) {
 	l := Loader{
 		dec: json.NewDecoder(r),
 		cur: uuid.Nil,
@@ -53,21 +64,25 @@ func Load(ctx context.Context, r io.Reader) (*Loader, error) {
 	return &l, nil
 }
 
-// Loader is an iterator that returns a series of [Entry].
+// Loader is an iterator over serialized records.
 //
-// Users should call [*Loader.Next] until it reports false, then check for
-// errors via [*Loader.Err].
+// Users should consume the [Loader.All] iterator, then check for errors via
+// [Loader.Err].
 type Loader struct {
-	err error
-	e   *Entry
+	err     error
+	dec     *json.Decoder
+	started bool
 
-	dec  *json.Decoder
+	// Used in the deprecated [Loader.Next]+[Loader.Entry] flow.
+	e    *Entry
 	next *Entry
 	de   diskEntry
 	cur  uuid.UUID
 }
 
 // Next reports whether there's an [Entry] to be processed.
+//
+// Deprecated: Use the [Loader.All] iterator.
 func (l *Loader) Next() bool {
 	if l.err != nil {
 		return false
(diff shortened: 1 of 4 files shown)


libvuln/updates.go
```diff
@@ -2,67 +2,84 @@ package libvuln
...
+			break
 		}
-		log.InfoContext(ctx, "update imported",
-			"ref", ref,
-			"vuln_count", len(e.Vuln),
-			"enrichment_count", len(e.Enrichment))
 	}
-	if err := l.Err(); err != nil {
+
+	if err := errors.Join(err, l.Err()); err != nil {
```
crozzy: I think this `err` might be the original `l, err := jsonblob.NewLoader(ctx, in)` error so IIUC the `UpdateVulnerabilitiesIter` call might fail, log, break, but the function might return nil?

libvuln/jsonblob/jsonblob.go
```diff
@@ -113,10 +128,75 @@ func (l *Loader) Next() bool {
...
+				l.err = json.Unmarshal(de.Enrichment.buf, e)
+			}
+			if l.err != nil || !yield(v, e) {
+				return
+			}
+		}
+	}
+	// Outer reads the first entry and handles when the Ref changes.
+	outer := func(yield func(*Entry, iter.Seq2[*claircore.Vulnerability, *driver.EnrichmentRecord]) bool) {
+		l.started = true
```
jvdm, line 179: Fail the iteration if `l.started == true`?

libvuln/jsonblob/jsonblob.go
```diff
@@ -113,10 +128,75 @@ func (l *Loader) Next() bool {
...
+	// The inner iterator decodes every entry after the first one, stopping
+	// iteration and passing control to the outer iterator when the Ref
+	// changes.
+	inner := func(yield func(*claircore.Vulnerability, *driver.EnrichmentRecord) bool) {
+		for ; l.err == nil && cur == de.Ref; l.err = l.dec.Decode(&de) {
+			var v *claircore.Vulnerability
+			var e *driver.EnrichmentRecord
+			switch de.Kind {
+			case driver.VulnerabilityKind:
+				v = getVulnerability()
```
crozzy, line 155: Is there a chance of stale info in here from the previous vuln? Should we "reset"?

libvuln/updates.go
```diff
@@ -2,67 +2,84 @@ package libvuln
...
+	l, err := jsonblob.NewLoader(ctx, in)
 	if err != nil {
 		return err
 	}
 
-	ops, err := s.GetUpdateOperations(ctx, driver.VulnerabilityKind)
-	if err != nil {
-		return err
-	}
+	// Don't bother checking the fingerprint, just gogogo.
```
crozzy: What is the intent behind skipping the fingerprint checking, wouldn't it stop duplicate work if the exports/imports are the same?

libvuln/updates.go
```diff
@@ -2,67 +2,84 @@ package libvuln
...
 			}
+			ref, err = s.UpdateEnrichmentsIter(ctx, e.Updater, e.Fingerprint, wrap)
+		default:
+			panic("unreachable")
+		}
+		if err == nil {
+			l.InfoContext(
+				ctx, "update imported",
+				"ref", ref,
+				"vuln_count", len(e.Vuln),
```
crozzy: I think this should use `vulnCt` and `enrichCt`

libvuln/offlineimport_test.go
```diff
@@ -0,0 +1,120 @@
...
+	defer func() { os.Exit(c) }()
+	defer integration.DBSetup()()
+	c = m.Run()
+}
+
+// TestLiveOfflineImport is meant to be used for profiling and testing the
+// system on a "real" export as produced by `clairctl`.
+func TestOfflineImport(t *testing.T) {
+	if importFile == nil {
+		t.Skip(`needed flag "-load-export" not provided`)
```
crozzy: nit: I think this is `-load-file`

libvuln/jsonblob/jsonblob.go
```diff
@@ -113,10 +128,75 @@ func (l *Loader) Next() bool {
...
+//
+// Deprecated: Use the [Loader.All] iterator.
 func (l *Loader) Entry() *Entry {
 	return l.e
 }
 
+// All returns an iterator-of-iterators yielding all entries in the Loader.
+//
+// The inner iterator returns exactly one of the two values.
+func (l *Loader) All() iter.Seq2[*Entry, iter.Seq2[*claircore.Vulnerability, *driver.EnrichmentRecord]] {
```
jvdm, line 141: If a decode error occurs mid-operation, the inner iterator silently stops producing records and caller will only detect the error via `Err()` after the outer iteration completes. This means a caller that's streaming records into a transactional store (like `UpdateVulnerabilitiesIter`) can't distinguish "all records consumed successfully" from "decode failed after record N" and transaction should stop. The store commits a partial result without knowing something went wrong.

libvuln/jsonblob/jsonblob.go
```diff
@@ -45,29 +46,43 @@ type Store struct {
...
 	return &l, nil
 }
 
-// Loader is an iterator that returns a series of [Entry].
+// Loader is an iterator over serialized records.
 //
-// Users should call [*Loader.Next] until it reports false, then check for
-// errors via [*Loader.Err].
+// Users should consume one of the [Loader.Vulnerabilities] or
+// [Loader.Enrichments] iterators, then check for errors via [Loader.Err].
```
crozzy: Should this just reference `Loader.All()` now?

crozzy (changes requested):
Seems like this will be a nice upgrade

libvuln/jsonblob/jsonblob.go
hdonnay: Yeah, stale comment.

libvuln/offlineimport_test.go
hdonnay: Yeah, stale name 👍

libvuln/jsonblob/jsonblob.go
hdonnay: Yeah, I suppose that could happen. Will change.

libvuln/jsonblob/jsonblob.go
hdonnay: Yeah I guess we should check here also, as a caller could conceivably call `All()` twice before using the first iterator.

libvuln/jsonblob/jsonblob.go
hdonnay: Would just returning a pair of `nil` work as a signal? Or do you think it should be `iter.Seq2[struct{ *claircore.Vulnerability; *dirver.EnrichmentRecord}, error]`?

libvuln/updates.go
hdonnay: It would, but how often is that happening? I can add it back if you think it's worth it.

libvuln/updates.go
hdonnay: Yeah good catch there was a shadowing `err`

libvuln/jsonblob/jsonblob.go
jvdm: I think returning `nil`  would be OK. Or, you could make `l.Err()` safe to call in between iterations, so that we can:

```golang
for v := range seq {
  if !yield(v, nil) {
    break
  }
}
if err := l.Err(); err != nil {
  if !yield(nil, err) {
       return
  }
}

jvdm: > Or, you could make `l.Err()` safe to call in between iterations,

I mean, make calling `l.Err()` in between iterators part of the contract.

hdonnay: I implemented this as returning a `nil` pointer as a signal to go check `l.Err()`.

jvdm (approved):

libvuln/updates.go
hdonnay: Update: I did this. Also made it work for enrichments.

crozzy (approved):
SGTM

PR #1997, opened 2026-08-26
fetcher: inspect gzip trailer for size

hdonnay:
As an optimization, pre-allocate the destination on disk when possible.


hdonnay:
Congrats @BradLugo, you're now the fellow domain expert on how this works.

PR #2013, opened 2026-09-04
osv: backport non-applicable ecosystem advisory filter to stable-1.5.44

guzalv:
Backports #1958 ("osv: exclude advisories from non applicable ecosystems") to `stable-1.5.44`.

## Problem

OSV publishes per-ecosystem dumps that still embed advisories covering *other* ecosystems. The OSV updater ingested every `affected` entry regardless of `package.ecosystem`, so fix data from a foreign ecosystem leaked into the wrong repo. Concretely, PyPI `langsmith` 0.10.16 was flagged for CVE-2026-45134 (GHSA-3644-q5cj-c5c7) using npm's fix version `0.6.0` (lower than the installed version) instead of PyPI's `0.8.0`, producing a false positive.

## Fix

Cherry-pick of upstream `3f07716d`:
- `updater/osv/osv.go`: skip `affected` entries whose normalized ecosystem doesn't match the dump's ecosystem (`normalizeEcosystem(af.Package.Ecosystem) != dumpEcosystem`).
- matcher migration to force the OSV updaters to re-fetch/re-parse and drop the contaminated rows.
- regression test `TestInsert/skip_other_ecosystems` (PyPI `0.8.0` vs npm `0.6.0`).

## Backport adaptations for `stable-1.5.44`

- Migration renumbered `20-` → `17-`: this branch's matcher migrations top out at `16`, and migrations are applied by readdir position rather than the filename number (`migrations.go`: `id := i + 1`), so this is naming-only — the runtime behavior is unchanged.
- Removed the `Vulnerability.Self` / `claircore.Alias` expectation from the new test case: typed aliases (`unique.Make`, `Self`) don't exist on this branch, and `cmpIgnore` doesn't compare that field. The test still asserts `FixedInVersion="fixed=0.8.0"`, i.e. npm's `0.6.0` no longer leaks into the PyPI record.

## Testing

`go test ./updater/osv/...` passes, including the new `skip_other_ecosystems` case.

### DIFF
--- a/datastore/postgres/migrations/matcher/17-force-osv-ecosystem-filter.sql
+++ b/datastore/postgres/migrations/matcher/17-force-osv-ecosystem-filter.sql
@@ -1,0 +1,14 @@
+-- Force per-ecosystem OSV updaters to re-fetch and re-parse after filtering
+-- affected entries to the dump's ecosystem. Multi-ecosystem advisories
+-- previously produced cross-ecosystem vulnerability rows (e.g. npm ranges
+-- under osv/pypi).
+DELETE FROM update_operation
+WHERE
+  updater LIKE 'osv/%';
+
+DELETE FROM vuln v1 USING vuln v2
+LEFT JOIN uo_vuln uvl ON v2.id = uvl.vuln
+WHERE
+  uvl.vuln IS NULL
+  AND v2.updater LIKE 'osv/%'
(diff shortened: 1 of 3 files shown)


hdonnay:
This will prevent database upgrades because it creates a divergent migration numbering.

PR #2014, opened 2026-09-08
all: update toolkit

hdonnay:
This updates the dependency on `toolkit` and callsites for the `(cpe.WFN).BindFS` method.


crozzy:
Looks good, should `(w *WFN) Value()` also be update to not use BindFS()?

hdonnay:
> Looks good, should `(w *WFN) Value()` also be update to not use BindFS()?

We can do that as another PR & toolkit release.

crozzy (approved):

PR #2022, opened 2026-09-10
internal/rpm: accept every type in a tag's class when loading package info

arpitjain099:
`Info.Load` asserts one concrete Go type per tag, but the header verifier does not guarantee one. `verifyInfo` accepts an entry whose `Type` merely shares a *class* with the type named in the tag table, and `checkTagType` says why:

```go
// Check the type. Some versions of string are typed incorrectly in a
// compatible way.
return t == typ || t.class() == typ.class()
```

`TypeString`, `TypeStringArray` and `TypeI18nString` are all `classString`. `ReadData` switches on the type the entry declared rather than on the tag, so a `Name` entry declared as an i18n string or a string array comes back as a `[]string` while `Load` does `v.(string)`. The result is a panic on a header the verifier has just accepted:

```
panic: interface conversion: interface {} is []string, not string
```

The same gap covers every numeric tag, since `TypeChar`, `TypeInt8`, `TypeInt16`, `TypeInt32` and `TypeInt64` share `classNumeric` and yield five different slice types against a single `v.([]int32)`. It is wider still for BDB headers, where there is no region, `typecheck` is false and `checkTagType` is skipped entirely.

Worth noting that this is not contained to the one bad package. Nothing on the path recovers, whereas `Load` returning an error is already handled: the iterator in `database.go` yields it and continues to the next package in the database. So a header like this currently takes down more than it needs to.

This adds small helpers that accept any member of the tag's class, widening the numeric variants to `int32`, and report anything genuinely outside the class.

Verification: a new test in `internal/rpm` rewrites the type field of an entry in `rpmdb/testdata/package.header` and runs the result through the real `ParseHeader`, so the header goes through `verifyInfo` on the way in. It panics on main and passes here, recovering `name="crypto-policies" version="20210213"` in all three cases. `go test ./internal/... ./rpm/...` is unchanged against main, 17 packages, no failures either side.

### DIFF
--- a/internal/rpm/info.go
+++ b/internal/rpm/info.go
@@ -50,31 +50,36 @@ func (i *Info) Load(ctx context.Context, h *rpmdb.Header) error {
 		}
 		switch e.Tag {
 		case rpmdb.TagName:
-			i.Name = v.(string)
+			i.Name, err = tagString(e, v)
 		case rpmdb.TagEpoch:
-			i.Epoch = int(v.([]int32)[0])
+			i.Epoch, err = tagInt(e, v)
 		case rpmdb.TagVersion:
-			i.Version = v.(string)
+			i.Version, err = tagString(e, v)
 		case rpmdb.TagRelease:
-			i.Release = v.(string)
+			i.Release, err = tagString(e, v)
 		case rpmdb.TagSourceRPM:
-			i.SourceRPM = v.(string)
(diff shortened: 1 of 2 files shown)


hdonnay:
I'm disinclined to accept this without evidence of this happening in the wild.

Rpm tags have documented types, which are the ones used in the code.

PR #2023, opened 2026-09-10
java: update DefaultSearchAPI to central.sonatype.com

kovayur:
The legacy search.maven.org Solr index is no longer being updated by Sonatype — it appears frozen around June 2025. SHA1 lookups for artifacts published after that date return no results or 403 errors, breaking Maven search for recent Java packages.

Switch to central.sonatype.com/solrsearch/select which uses the same Solr query format but is backed by a current index.

Ref: https://central.sonatype.org/faq/what-happened-to-search-maven-org/
Ref: https://github.com/dependency-check/DependencyCheck/issues/7889


.gitignore
hdonnay, line 1: This change is unneeded.

hdonnay (changes requested):
Changes needed:

- [x] Fix commit message: the prefix should be the package that was changed, in this case `java`.
- [x] Fix the DCO sign-off.
- [x] Remove the `.gitignore` change.

.gitignore
kovayur: removed

hdonnay (changes requested):
Still needs the commit message fixed.

kovayur:
Commit message fixed.

hdonnay (approved):

PR #2027, opened 2026-09-16
rhel-vex: only ingest OCI known_not_affected advisory data

crozzy:
RPM known_not_affected data accounts for roughly half of the rhel-vex rows in the vuln table and offers little value to consumers. In the interests of resources and space the VEX parser no longer consumes them. OCI known_not_affected advisories are consumed as normal.


hdonnay (approved):
SGTM

PR #2034, opened 2026-09-18
migrations: complexify the migrations

hdonnay:
This change allows for two things:

1. Nonconsecutive application of migrations.
2. Application-driven migrations.

The beefed-up documentation covers the why and how.


hdonnay:
Addressed some out-of-band feedback; this now avoids calling the migration detection "feature flags".

Note on the structure of the change: this is one commit to allow backporting the migration scheme itself, if/when needed.

